PR #163, opened 2024-12-05
Display the block even if `column.default` is JS falsy value (for example, 0.)

hoshinotsuyoshi:
## Summary

Display the block even if `column.default` is JS falsy value (for example, 0.)


MH4GF (approved):
Oh!! nice catch 👍🏻

PR #167, opened 2024-12-05
Add foreign key parsing for schemarb parser

sasamuku:
## Summary

Add foreign key parsing for schemarb parser.

## Related Issue

For postgres paser:

* https://github.com/liam-hq/liam/pull/131

## Other Information

Two patterns of notation for foreign key constraints in schema.rb

1️⃣: Foreign key constraints at the same time as table creation

```rb
create_table :products do |t|
  t.references :category, foreign_key: true
end
```

2️⃣: Foreign key constraints after table creation

```rb
add_foreign_key :articles, :authors
```

I looked through several Rails projects and they are all written in  2️⃣.

https://github.com/mastodon/mastodon/blob/main/db/schema.rb
https://github.com/autolab/Autolab/blob/master/db/schema.rb
https://github.com/lobsters/lobsters/blob/master/db/schema.rb
https://github.com/astuto/astuto/blob/main/db/schema.rb
https://github.com/chatwoot/chatwoot/blob/develop/db/schema.rb

It would be good to support  2️⃣ first.
If we find a project that uses  1️⃣, we can support that as well.

### DIFF
--- a/frontend/packages/db-structure/src/parser/schemarb/parser.ts
+++ b/frontend/packages/db-structure/src/parser/schemarb/parser.ts
@@ -242,9 +242,9 @@ function normalizeConstraintName(constraint: string): ForeignKeyConstraint {
     case 'restrict':
       return 'RESTRICT'
     case 'nullify':
-      return 'SET NULL'
+      return 'SET_NULL'
     default:
-      return 'NO ACTION'
+      return 'NO_ACTION'
   }
 }
 
--- a/frontend/packages/db-structure/src/parser/schemarb/parser.ts
+++ b/frontend/packages/db-structure/src/parser/schemarb/parser.ts
@@ -20,10 +20,11 @@ import type {
   DBStructure,
   Index,
   Indices,
+  Relationship,
   Table,
   Tables,
 } from '../../schema/index.js'
-import { aColumn, aTable, anIndex } from '../../schema/index.js'
+import { aColumn, aRelationship, aTable, anIndex } from '../../schema/index.js'
 import type { Processor } from '../types.js'
 
 function extractTableName(argNodes: Node[]): string {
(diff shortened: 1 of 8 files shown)


frontend/packages/db-structure/src/parser/schemarb/parser.ts
```diff
@@ -216,54 +218,159 @@ function extractDefaultValue(
...
+  handleAddForeignKey(node: CallNode): void {
+    const argNodes = node.arguments_?.compactChildNodes() || []
+
+    const [primaryTableName, foreignTableName] =
+      extractRelationshipTableNames(argNodes)
+
+    const relationship = aRelationship({
+      primaryTableName: primaryTableName,
+      // TODO: This is a guess, we should add a way to specify the primary column name
+      primaryColumnName: 'id',
```
MH4GF, line 362: Looks good now that we only support it if it's on Rails' Convention 👍🏻

frontend/packages/db-structure/src/schema/dbStructure.ts
```diff
@@ -43,15 +43,31 @@ const tableSchema = v.object({
...
+const cardinalitySchema = v.union([
+  v.literal('ONE_TO_MANY'),
+  v.literal('ONE_TO_ONE'),
+])
+const foreignKeyConstraintSchema = v.union([
+  v.literal('CASCADE'),
+  v.literal('RESTRICT'),
+  v.literal('SET NULL'),
+  v.literal('SET DEFAULT'),
+  v.literal('NO ACTION'),
```
MH4GF, line 55: I thought it would be good to unify with UPPER_CAMEL_CASE. What do you think?

```suggestion
  v.literal('SET_NULL'),
  v.literal('SET_DEFAULT'),
  v.literal('NO_ACTION'),
```

MH4GF (approved):
👍🏻

frontend/packages/db-structure/src/schema/dbStructure.ts
sasamuku: Yeah, I was torn between `SET NULL` or `SET_NULL`.
I decided `SET NULL` because values can be used in frontend view as is.
I thought it would be better to display it with spaces for user.

Is it necessary to use it for the frontend view as is?

MH4GF: The format as data and the label as UI should be considered separately.
This is because there are many possible ways of expression, such as trial and error in UX writing or the use of icons.

hoshinotsuyoshi:
> 1️⃣: Foreign key constraints at the same time as table creation
> 
> ```rb
> create_table :products do |t|
>   t.references :category
> end
> ```

`t.references :category, foreign_key: true` is needed? 

------

> It would be good to support 2️⃣ first.
> If we find a project that uses 1️⃣, we can support that as well.

Yes, I also think Rails' schema dumper always uses 2️⃣. Since `db/schema.rb` isn’t manually written, I believe `t.references :category, foreign_key: true` won’t appear in the schema file (`db/schema.rb` ) as a result. 👍

hoshinotsuyoshi (approved):
LGTM!

frontend/packages/db-structure/src/schema/dbStructure.ts
sasamuku: Fixed 👍:
9bbf00668b87d236ab12764d26c3e6c0e8671c10

MH4GF: thanks 👍🏻

sasamuku:
@hoshinotsuyoshi 
>t.references :category, foreign_key: true is needed?

Yes, thank you for letting me know.
When we need foreign key constraint on database, `foreign_key: true` is needed.

>When using references in a migration, you are creating a new column in the table and you'll have the option to add a foreign key using foreign_key: true to that column.
https://guides.rubyonrails.org/active_record_migrations.html#foreign-keys

PR #169, opened 2024-12-05
maintenance: enhance test case for parser and chore schema

MH4GF:
Details are described in the in-line comments.


frontend/packages/db-structure/src/parser/sql/postgresql/converter.ts
```diff
@@ -118,10 +122,6 @@ export const convertToDBStructure = (ast: RawStmtWrapper[]): DBStructure => {
               ?.filter(isConstraintNode)
               .some((c) => c.Constraint.contype === 'CONSTR_PRIMARY') ||
             false,
-          increment:
-            colDef.typeName?.names
-              ?.filter(isStringNode)
-              .some((n) => n.String.sval === 'serial') || false,
```
MH4GF, line 124: I've remove unnecessary increment field.

frontend/packages/db-structure/src/parser/__tests__/testcase.ts
```diff
@@ -0,0 +1,35 @@
...
+        comment: override?.comment ?? null,
+      }),
+    },
+  })
+
+export const parserTestCases = {
+  'table comment': userTable({
+    comment: 'store our users.',
+  }),
+}
```
MH4GF, line 35: To unify the behaviour across multiple parsers, an object defining common test cases has been added.

frontend/packages/db-structure/src/parser/sql/postgresql/converter.ts
```diff
@@ -108,7 +108,11 @@ export const convertToDBStructure = (ast: RawStmtWrapper[]): DBStructure => {
           unique:
             colDef.constraints
               ?.filter(isConstraintNode)
-              .some((c) => c.Constraint.contype === 'CONSTR_UNIQUE') || false,
+              .some((c) =>
+                ['CONSTR_UNIQUE', 'CONSTR_PRIMARY'].includes(
+                  c.Constraint.contype ?? '',
+                ),
+              ) || false,
```
MH4GF, line 115: Behaviour aligned with schemarb-parser; 
`unique: true` for primary key.

frontend/packages/db-structure/src/parser/sql/postgresql/index.test.ts
```diff
@@ -36,23 +37,18 @@ describe(processor, () => {
...
-
-      expect(result).toEqual(expected)
+      expect(result).toEqual(parserTestCases['table comment'])
     })
 
     it('not null', async () => {
       const result = await processor(/* sql */ `
         CREATE TABLE users (
-          id SERIAL PRIMARY KEY,
+          id BIGSERIAL PRIMARY KEY,
```
MH4GF, line 51: Behaviour aligned with schemarb-parser: test data of type id is `BIGSERIAL`.

frontend/packages/db-structure/src/parser/__tests__/testcase.ts
hoshinotsuyoshi: > If this policy is acceptable, other test cases will be changed.

ask:
`userTable` will be defined only here(`src/parser/__tests__/testcase.ts`) ?

frontend/packages/db-structure/src/parser/__tests__/testcase.ts
```diff
@@ -0,0 +1,35 @@
...
+        indices: {
+          ...override?.indices,
+        },
+        comment: override?.comment ?? null,
+      }),
+    },
+  })
+
+export const parserTestCases = {
+  'table comment': userTable({
```
hoshinotsuyoshi, line 32: Nice 👍 
It seems great that type inference works!

frontend/packages/db-structure/src/parser/__tests__/testcase.ts
MH4GF: @hoshinotsuyoshi 

> userTable will be defined only here(src/parser/__tests__/testcase.ts) ?

Yes, I think it's better that way.

hoshinotsuyoshi (approved):
LGTM, but I’d like to hear others' opinions as well!

frontend/packages/db-structure/src/parser/__tests__/testcase.ts
sasamuku: @MH4GF 
Looks nice 👍
If this policy is accepted, we have to unify type name as well, right?
I would like to know how to change type name for Oracle, SQL Server and so on.

MH4GF: @sasamuku Yes, there will inevitably be subtle differences.
I will consider how to override this as we move forward with implementation. 👍🏻

sasamuku (approved):
Thank you for enhancement!

PR #173, opened 2024-12-06
fix: display selected table name

junkisai:
I fixed it so that the table name of the selected TableNode is displayed, as previously all TableNodes showed only 'Table Name' regardless of which one was selected.


hoshinotsuyoshi (approved):
LGTM!

PR #178, opened 2024-12-06
chore: remove lucide-react from erd-core package

junkisai:
I modified icon imports to use the ui package exclusively and removed lucide-react from the erd-core package.


MH4GF (approved):
Nice!! That's better.

PR #180, opened 2024-12-06
feat: Adjust animation-duration for TableDetail toggle to 165ms

junkisai:
I adjusted the animation-duration for showing and hiding the TableDetail to 165ms.


MH4GF (approved):
thanks 👍🏻

PR #166, opened 2024-12-05
feat: parse one to one relationships

MH4GF:
## What changed

- **feat: Implement handling of one-to-one relationships based on unique indexes in PostgreSQL converter**
- **feat: Add cardinality validation to relationship schema**

PR #183, opened 2024-12-06
feat: parse one to one relationships for schema.rb

sasamuku:
## Summary

Support `ONE_TO_ONE` cardinality for schema.rb parser.

If foreign key is unique, it must be one to one relationship.

```ruby
create_table "posts" do |t|
  t.bigint "user_id", unique: true
end

add_foreign_key "posts", "users", column: "user_id", name: "fk_posts_user_id"
```

## Related Issue

* https://github.com/liam-hq/liam/pull/166

### DIFF
--- a/frontend/packages/db-structure/src/parser/schemarb/index.test.ts
+++ b/frontend/packages/db-structure/src/parser/schemarb/index.test.ts
@@ -220,5 +220,28 @@ describe(processor, () => {
 
       expect(result.relationships).toEqual(expected)
     })
+
+    it('unique foreign key', async () => {
+      const result = await processor(/* Ruby */ `
+        create_table "posts" do |t|
+          t.bigint "user_id", unique: true
+        end
+
+        add_foreign_key "posts", "users", column: "user_id", name: "fk_posts_user_id"
+      `)
+
+      const rel = aRelationship({
+        name: 'fk_posts_user_id',
+        primaryTableName: 'users',
+        primaryColumnName: 'id',
+        foreignTableName: 'posts',
+        foreignColumnName: 'user_id',
+        cardinality: 'ONE_TO_ONE',
+      })
+
+      const expected = { fk_posts_user_id: rel }
+
+      expect(result.relationships).toEqual(expected)
+    })
   })
 })
--- a/frontend/packages/db-structure/src/parser/schemarb/parser.ts
+++ b/frontend/packages/db-structure/src/parser/schemarb/parser.ts
@@ -22,27 +22,12 @@ import type {
   Index,
   Indices,
   Relationship,
-  Relationships,
   Table,
   Tables,
 } from '../../schema/index.js'
 import { aColumn, aRelationship, aTable, anIndex } from '../../schema/index.js'
 import type { Processor } from '../types.js'
-
-// If there is a unique index for a column in relationships, make it `ONE_TO_ONE` cardinality.
-const handleOneToOneRelationships = (
-  tables: Tables,
-  relationships: Relationships,
-) => {
-  for (const relationship of Object.values(relationships)) {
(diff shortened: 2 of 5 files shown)


frontend/packages/db-structure/src/parser/schemarb/parser.ts
```diff
@@ -22,12 +22,28 @@ import type {
...
+) => {
+  for (const relationship of Object.values(relationships)) {
+    const foreignTable = tables[relationship.foreignTableName]
+    const foreignColumn = foreignTable?.columns[relationship.foreignColumnName]
+
+    if (foreignColumn?.unique) {
+      relationship.cardinality = 'ONE_TO_ONE'
+    }
+  }
+}
```
MH4GF, line 45: As this is exactly the same as the postgres parser implementation, it seems better to make this common.
What do you think about putting it in `db-structure/src/parser/utils/handleOneToOneRelationships.ts`?

MH4GF (commented):
I've wrote a comment!

frontend/packages/db-structure/src/parser/schemarb/index.test.ts
```diff
@@ -220,5 +220,28 @@ describe(processor, () => {
 
       expect(result.relationships).toEqual(expected)
     })
+
+    it('unique foreign key', async () => {
+      const result = await processor(/* Ruby */ `
+        create_table "posts" do |t|
+          t.bigint "user_id", unique: true
```
hoshinotsuyoshi, line 227: I believe this test case is necessary:

```suggestion
          t.bigint "user_id"
          t.index ["user_id"], name: "index_posts_on_user_id", unique: true
```

This is a real-world example of a `has_one` association. Rails' schema dumper will be expected to place `unique: true` in the `t.index` declaration.



Rails schema dumper output `db/schema.rb`:

https://github.com/mastodon/mastodon/blob/fbcd14ff7e18c8689426433aa5ad673c9ebd6409/db/schema.rb#L272  
https://github.com/mastodon/mastodon/blob/fbcd14ff7e18c8689426433aa5ad673c9ebd6409/db/schema.rb#L281


Model (for reference):

https://github.com/mastodon/mastodon/blob/fbcd14ff7e18c8689426433aa5ad673c9ebd6409/app/models/account_warning.rb#L38

frontend/packages/db-structure/src/parser/schemarb/parser.ts
sasamuku: You are right!
I've added `handleOneToOneRelationships.ts` as utils.
9fb06b3bb413bf363757520ae1ed44e40903abfd

frontend/packages/db-structure/src/parser/schemarb/index.test.ts
sasamuku: @hoshinotsuyoshi  Thank you for information!

I thought Rails schema dumper don't care `has_one` or `has_many`.
Is this a standard Rails specification?

hoshinotsuyoshi: @sasamuku 

> Rails schema dumper don't care has_one or has_many.

You're absolutely right. Rails schema dumper doesn't care about `has_one` or `has_many`. My description in the model was just a reference. 🙏

sasamuku: @hoshinotsuyoshi 
I see! Thanks.

`t.index` can not support this implementation for now.
And we may need to care about postgres parser as well.

So let me make another pr for supporting index.

hoshinotsuyoshi: @sasamuku thanks!

MH4GF (approved):
LGTM 🚀

PR #186, opened 2024-12-09
Enable text cursor in TableDetail

sasamuku:
## Summary

* `user-select: auto;` to allow selection of TableDetail text.
* `snapPoints` to be an empty array to disable snapping function

### before
https://github.com/user-attachments/assets/81edc3ca-4f2f-4773-9896-865fea6d0ce5

### after
https://github.com/user-attachments/assets/639b0a56-bde6-426b-8d5c-1b9745705f80

## Related Issue

* https://github.com/liam-hq/liam/pull/160

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableNode.tsx
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableNode.tsx
@@ -27,7 +27,12 @@ export const TableNode: FC<Props> = ({ data: { table } }) => {
   const { relationships } = useDBStructureStore()
   return (
     <>
-      <DrawerRoot direction="right">
+      {/*
+      Set snapPoints to an empty array to disable the drawer snapping functionality.
+      This behavior is an undocumented, unofficial usage and might change in the future.
+      ref: https://github.com/emilkowalski/vaul/blob/main/src/use-snap-points.ts
+      */}
+      <DrawerRoot direction="right" snapPoints={[]}>
         <DrawerTrigger>
           <div className={styles.wrapper}>
             <TableHeader name={table.name} />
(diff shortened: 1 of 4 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableNode.tsx
```diff
@@ -25,9 +25,16 @@ type Props = NodeProps<TableNodeType>
...
 export const TableNode: FC<Props> = ({ data: { table } }) => {
   const { relationships } = useDBStructureStore()
+  const [isOpen, setIsOpen] = useState(false)
+
   return (
     <>
-      <DrawerRoot direction="right">
+      <DrawerRoot
+        direction="right"
+        dismissible={false}
```
sasamuku, line 34: see: https://vaul.emilkowal.ski/other#non-dismissible

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableNode.tsx
```diff
@@ -25,9 +25,16 @@ type Props = NodeProps<TableNodeType>
...
+        open={isOpen}
+        onOpenChange={setIsOpen}
```
sasamuku, line 36: I'm not sure, but when dismissible turn to false, set handler will be needed.
I feel this behavior is not correct 🤔 and would appreciate your review.

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableNode.tsx
MH4GF: The close button is far away, so it looked good on dismissible, should we set it to false?

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableNode.tsx
MH4GF: It seems to be necessary if dismissible=false!

MH4GF (commented):
I've comments!!

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableNode.tsx
MH4GF: nice comments 👍🏻

MH4GF (approved):
thanks your work!!

PR #189, opened 2024-12-09
feat: implement related tables

MH4GF:
I've implemented related tables viewer in table detail.

https://github.com/user-attachments/assets/d1b02e61-8172-491d-bef9-7ae197b0414c

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableDetail/RelatedTables/RelatedTables.module.css
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableDetail/RelatedTables/RelatedTables.module.css
@@ -24,4 +24,6 @@
   border-radius: var(--border-radius-smbase);
   border: 1px solid var(--pane-border);
   background: var(--global-background);
+  width: 100%;
+  height: 333px;
 }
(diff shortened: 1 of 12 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableDetail/RelatedTables/RelatedTables.module.css
```diff
@@ -24,4 +24,6 @@
   border-radius: var(--border-radius-smbase);
   border: 1px solid var(--pane-border);
   background: var(--global-background);
+  width: 100%;
+  height: 333px;
```
MH4GF, line 28: 📝 The width of the tableDetail is 500px, so the aspect ratio is set to 3:2.

frontend/turbo.json
```diff
@@ -22,10 +22,8 @@
     },
     "fmt": {},
     "test": {
+      "dependsOn": ["^build"],
       "outputs": []
-    },
-    "@liam-hq/cli#test": {
-      "dependsOn": ["^build"]
```
MH4GF, line 28: 📝 Applied across the all, as build was needed first in `erd-core` as well as `cli`.

sasamuku (approved):
LGTM 🚀

Repeatedly opening `TableDetail` from `Related tables` is something we will need to discuss.

PR #190, opened 2024-12-09
Support FalseNode for id args

sasamuku:
## Summary

Support `FalseNode` for `create_table` `id` args.

## Why we only support `FalseNode`:

`id: true` is not supported in Rails 8.0.0, so there's no need to support it.

schema.rb:

```ruby
create_table "users", id: true do |t|
  t.string "name"
end
```

Error occur:

```ruby
[ERROR] undefined method `to_sym' for true
```

Reason:

```ruby
def column(name, type, index: nil, **options)
  name = name.to_s
  type = type.to_sym if type # <-- This is the cause
end
```

https://github.com/rails/rails/blob/v8.0.0/activerecord/lib/active_record/connection_adapters/abstract/schema_definitions.rb#L490

### DIFF
--- a/frontend/packages/db-structure/src/parser/schemarb/index.test.ts
+++ b/frontend/packages/db-structure/src/parser/schemarb/index.test.ts
@@ -168,6 +168,30 @@ describe(processor, () => {
       expect(result).toEqual(expected)
     })
 
+    it('no primary key', async () => {
+      const result = await processor(/* Ruby */ `
+        create_table "users", id: false do |t|
+          t.string "name"
+        end
+      `)
+
+      const expected = aDBStructure({
+        tables: {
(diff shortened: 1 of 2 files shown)


frontend/packages/db-structure/src/parser/schemarb/index.test.ts
```diff
@@ -168,6 +168,30 @@ describe(processor, () => {
...
+                name: 'name',
+                type: 'varchar',
+              }),
+            },
+          }),
+        },
+      })
+
+      expect(result).toEqual(expected)
+    })
```
MH4GF, line 193: nice test!!

MH4GF (approved):
nice work 🚀

PR #191, opened 2024-12-09
feat: show warning errors when table name is not string

MH4GF:
## Description

Even if there are unsupported syntaxes, only warnings are displayed and the system exits normally.

## Background

Previously, if a name was passed to `create_table` as a variable, an error would occur because it was not a string.

```rb
variable = "posts"
create_table variable do |t|
end
```

This does not happen very often because schema.rb is generally generated code, but it often happens when users write their own table definitions while using ActiveRecord's DSL, such as [Ridgepole](https://github.com/ridgepole/ridgepole).

## Solution

Error information is logged but exits normally.
The output is still dirty, but the formatting of the logs will be tackled in a separate issue.

```
node packages/cli/dist-cli/bin/cli.js erd build --input ../../../test/db/Schemafile --format schemarb
UnsupportedTokenError: Expected a string for the table name, but received different data
    at extractTableName (file:///Users/mh4gf/ghq/github.com/liam-hq/liam/frontend/packages/cli/dist-cli/bin/cli.js:21231:20)
    at DBStructureFinder.handleCreateTable (file:///Users/mh4gf/ghq/github.com/liam-hq/liam/frontend/packages/cli/dist-cli/bin/cli.js:21489:28)
    at DBStructureFinder.visitCallNode (file:///Users/mh4gf/ghq/github.com/liam-hq/liam/frontend/packages/cli/dist-cli/bin/cli.js:21531:18)
    at CallNode.accept (file:///Users/mh4gf/ghq/github.com/liam-hq/liam/frontend/packages/cli/dist-cli/bin/cli.js:4330:13)
    at file:///Users/mh4gf/ghq/github.com/liam-hq/liam/frontend/packages/cli/dist-cli/bin/cli.js:612:17
    at Array.forEach (<anonymous>)
    at DBStructureFinder.visitChildNodes (file:///Users/mh4gf/ghq/github.com/liam-hq/liam/frontend/packages/cli/dist-cli/bin/cli.js:611:30)
    at DBStructureFinder.visitStatementsNode (file:///Users/mh4gf/ghq/github.com/liam-hq/liam/frontend/packages/cli/dist-cli/bin/cli.js:1897:10)
    at StatementsNode.accept (file:///Users/mh4gf/ghq/github.com/liam-hq/liam/frontend/packages/cli/dist-cli/bin/cli.js:17967:13)
    at file:///Users/mh4gf/ghq/github.com/liam-hq/liam/frontend/packages/cli/dist-cli/bin/cli.js:612:17
```

### DIFF
--- a/frontend/packages/cli/src/cli/runPreprocess.ts
+++ b/frontend/packages/cli/src/cli/runPreprocess.ts
@@ -26,7 +26,13 @@ export async function runPreprocess(
 
   let json = null
   try {
-    json = await parse(input, format)
+    const { value, errors } = await parse(input, format)
+    json = value
+    if (errors.length > 0) {
+      for (const error of errors) {
+        console.error(error)
+      }
+    }
   } catch (error) {
     throw new Error(
       `Failed to parse ${format} file: ${error instanceof Error ? error.message : String(error)}`,
--- a/frontend/packages/db-structure/src/parser/schemarb/parser.ts
+++ b/frontend/packages/db-structure/src/parser/schemarb/parser.ts
@@ -32,17 +32,23 @@ import type { ProcessResult, Processor } from '../types.js'
 import { handleOneToOneRelationships } from '../utils/index.js'
 import { convertColumnType } from './convertColumnType.js'
 
-export class TableNameNotFound extends WarningError {
+export class UnsupportedTokenError extends WarningError {
   constructor(message: string) {
     super(message)
-    this.name = 'WarningError'
+    this.name = 'UnsupportedTokenError'
   }
 }
 
-function extractTableName(argNodes: Node[]): Result<string, TableNameNotFound> {
+function extractTableName(
+  argNodes: Node[],
+): Result<string, UnsupportedTokenError> {
   const nameNode = argNodes.find((node) => node instanceof StringNode)
   if (!nameNode) {
-    return err(new TableNameNotFound('Table name not found'))
+    return err(
+      new UnsupportedTokenError(
+        'Expected a string for the table name, but received different data',
+      ),
+    )
   }
 
   // @ts-expect-error: unescaped is defined as string but it is actually object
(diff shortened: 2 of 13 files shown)


frontend/packages/cli/src/cli/runPreprocess.ts
```diff
@@ -26,7 +26,13 @@ export async function runPreprocess(
...
   let json = null
   try {
-    json = await parse(input, format)
+    const { value, errors } = await parse(input, format)
+    json = value
+    if (errors.length > 0) {
+      for (const error of errors) {
+        console.error(error)
+      }
+    }
```
MH4GF, line 35: Eventually, we would like to remove try-catch altogether, but this will be addressed in a separate PR.

frontend/packages/db-structure/src/parser/schemarb/parser.ts
```diff
@@ -26,17 +27,34 @@ import type {
...
+
+function extractTableName(
+  argNodes: Node[],
+): Result<string, UnsupportedTokenError> {
   const nameNode = argNodes.find((node) => node instanceof StringNode)
   if (!nameNode) {
-    throw new Error('Table name not found')
+    return err(
+      new UnsupportedTokenError(
+        'Expected a string for the table name, but received different data',
```
sasamuku, line 49: Table name could be symbol (and we need to support it)

```suggestion
        'Expected string or symbol for the table name, but received different data',
```

sasamuku (approved):
LGTM 🚀🚀

frontend/packages/db-structure/src/parser/schemarb/parser.ts
MH4GF: I see, but it just seems that the timing of the message change would be better to support.

sasamuku: Absolutely!

PR #193, opened 2024-12-10
fix: Add support for foreign key without column name

MH4GF:
## Description

`add_column_key()` to retrieve relationships even if the column name is omitted.
I have ported the method of `ActiveSupport::Inflector#singularize`.


frontend/packages/db-structure/src/parser/schemarb/singularize.ts
```diff
@@ -0,0 +1,47 @@
...
+    if (rule) {
+      const [regex, replacement] = rule
+      if (regex.test(word)) {
+        return word.replace(regex, replacement)
+      }
+    }
+  }
+
+  return word
+}
```
MH4GF, line 47: 📝 ActiveSupport::Inflector allows users to add dictionaries in the configuration, but does not support it once.

sasamuku (approved):
Great job!

PR #195, opened 2024-12-10
fix: set default relationship name when key name is not found

MH4GF:
- ref: https://github.com/liam-hq/liam/pull/193


sasamuku (approved):
Thanks! LGTM🚀

PR #196, opened 2024-12-10
feat: store active table name

MH4GF:
Keep the active table in the global state, so that the TableDetail can also be opened from the left pane.

https://github.com/user-attachments/assets/dd0052aa-04bd-46b0-bd3d-e6a5ab1188b0

## TODO(will be separated other PRs)

- [ ] Change style TableNode when active
- [ ] Fit view when changed active table
- [ ] Move files of TableDetail

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableNode.tsx
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableNode.tsx
@@ -1,17 +1,8 @@
 import { useDBStructureStore } from '@/stores'
 import type { Table } from '@liam-hq/db-structure'
-import {
-  DiamondFillIcon,
-  DiamondIcon,
-  DrawerContent,
-  DrawerPortal,
-  DrawerRoot,
-  DrawerTrigger,
-  KeyRound,
-} from '@liam-hq/ui'
+import { DiamondFillIcon, DiamondIcon, KeyRound } from '@liam-hq/ui'
 import { Handle, type Node, type NodeProps, Position } from '@xyflow/react'
 import type { FC } from 'react'
-import { TableDetail } from './TableDetail'
 import { TableHeader } from './TableHeader'
 import styles from './TableNode.module.css'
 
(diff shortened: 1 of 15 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableNode.tsx
MH4GF, line 1: 📝 Remove the Drawer part and just update the active tableName.

frontend/packages/erd-core/src/components/ERDRenderer/ERDRenderer.tsx
```diff
@@ -29,9 +30,12 @@ export const ERDRenderer: FC = () => {
...
             </div>
-            <ReactFlowProvider>
-              <ERDContent nodes={nodes} edges={edges} />
-            </ReactFlowProvider>
+            <TableDetailDrawerRoot>
+              <ReactFlowProvider>
+                <ERDContent nodes={nodes} edges={edges} />
+              </ReactFlowProvider>
+              <TableDetailDrawer />
+            </TableDetailDrawerRoot>
```
MH4GF, line 38: 📝 There is now only one TableDetail in the react render tree.

frontend/packages/erd-core/src/components/ERDRenderer/TableDetailDrawer/TableDetailDrawer.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { DrawerContent, DrawerPortal, DrawerRoot } from '@liam-hq/ui'
+import {
+  type FC,
+  type PropsWithChildren,
+  useCallback,
+  useEffect,
+  useState,
+} from 'react'
+// biome-ignore lint/nursery/useImportRestrictions: Fixed in the next PR.
+import { TableDetail } from '../ERDContent/TableNode/TableDetail'
```
MH4GF, line 15: As a large number of files will be changed, this will be done in a separate PR.

frontend/packages/erd-core/src/components/ERDRenderer/TableDetailDrawer/TableDetailDrawer.tsx
```diff
@@ -0,0 +1,68 @@
...
+import styles from './TableDetailDrawer.module.css'
+
+const ANIMATION_DURATION = 165
```
MH4GF, line 18: 📝 Defined here because we want to use duration on the React side as well.

junkisai (approved):
LGTM!!🎉

PR #197, opened 2024-12-10
Add edge styling (animation and coloring)

sasamuku:
## Summary

Added RelationshipEdge styling:

* When hovering over a node, only the primary-to-external edges are colored and animated.
* When hovering over a edge, edges are also colored and animated.

https://github.com/user-attachments/assets/b5634ddc-a787-4bae-8f1c-221a8cd202b7

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/RelationshipEdge/RelationshipEdge.tsx
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/RelationshipEdge/RelationshipEdge.tsx
@@ -5,7 +5,9 @@ import {
   getSmoothStepPath,
 } from '@xyflow/react'
 
+import clsx from 'clsx'
 import type { FC } from 'react'
+import styles from './RelationshipEdge.module.css'
 
 type Data = {
   isHovered: boolean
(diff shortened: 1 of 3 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/RelationshipEdge/RelationshipEdge.tsx
```diff
@@ -29,7 +36,17 @@ export const RelationshipEdge: FC<Props> = ({
...
+      <BaseEdge
+        id={id}
+        path={edgePath}
+        style={{
+          stroke: data?.isHovered
+            ? 'var(--node-layout)'
+            : 'var(--global-border)',
+          strokeDasharray: '2',
+          transition: 'stroke 0.3s var(--default-timing-function)',
+        }}
```
MH4GF, line 48: Styles should be implemented in the CSS Module as far as possible.
Perhaps it could be written as follows:

```suggestion
        className={clsx(styles.edge, data?.isHovered && styles.hovered)}
```

MH4GF (commented):
thanks 👍🏻 I've wrote comments!!

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/RelationshipEdge/RelationshipEdge.tsx
sasamuku: That's right 👍
6cc25f891f6053f70824cd05207ad857364da34f

MH4GF (approved):
Thanks 🚀

PR #199, opened 2024-12-10
chore: Make sample schema more complicated

sasamuku:
## Summary

Make sample schema more complicated for local development:

image


MH4GF (approved):
Great 👍🏻

PR #200, opened 2024-12-10
refactor(postgres parser): refactor constraints node handling

MH4GF:
As there was code to extract the relationship in each of ALTER TABLE and CREATE TABLE, it has been made common.

### DIFF
--- a/frontend/packages/db-structure/src/parser/sql/postgresql/converter.ts
+++ b/frontend/packages/db-structure/src/parser/sql/postgresql/converter.ts
@@ -69,12 +69,14 @@ const constraintToRelationship = (
     throw new Error('Invalid foreign key constraint')
   }
 
-  const name = defaultRelationshipName(
-    primaryTableName,
-    primaryColumnName,
-    foreignTableName,
-    foreignColumnName,
-  )
+  const name =
+    constraint.conname ??
+    defaultRelationshipName(
+      primaryTableName,
+      primaryColumnName,
+      foreignTableName,
+      foreignColumnName,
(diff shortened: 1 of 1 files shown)


frontend/packages/db-structure/src/parser/sql/postgresql/converter.ts
```diff
@@ -69,12 +69,14 @@ const constraintToRelationship = (
...
-    foreignColumnName,
-  )
+  const name =
+    constraint.conname ??
+    defaultRelationshipName(
+      primaryTableName,
+      primaryColumnName,
+      foreignTableName,
+      foreignColumnName,
+    )
```
MH4GF, line 79: 📝 `ALTER TABLE` uses conname and `CREATE TABLE` uses defaultRelationshipName, so we have now standardised them both.

frontend/packages/db-structure/src/parser/sql/postgresql/converter.ts
```diff
@@ -21,21 +21,83 @@ import {
...
+    case 'n':
+      return 'SET_NULL'
+    case 'd':
+      return 'SET_DEFAULT'
+    case 'a':
+      return 'NO_ACTION'
+    default:
+      return 'NO_ACTION' // Default to 'NO_ACTION' for unknown or missing values
+  }
+}
```
FunamaYukina, line 52: Thank you for moving 🙏.

FunamaYukina (approved):
LGTM! Thank you for refactoring!✨

PR #203, opened 2024-12-11
Support input url option for cli

sasamuku:
## Summary

Support input url option for cli.

```sh
./dist-cli/bin/cli.js erd build --input https://github.com/someapp/someapp/blob/main/db/schema.rb  --format schemarb
```

In most cases, we will receive a GitHub URL.
In that case, convert it to a RawContent URL such as:

```
https://raw.githubusercontent.com/someapp/someapp/main/db/schema.rb
```

### DIFF
--- a/frontend/packages/cli/src/cli/getInputContent.ts
+++ b/frontend/packages/cli/src/cli/getInputContent.ts
@@ -1,0 +1,49 @@
+import fs from 'node:fs'
+import { URL } from 'node:url'
+
+function isValidUrl(url: string): boolean {
+  try {
+    new URL(url)
+    return true
+  } catch {
+    return false
+  }
+}
+
+function isGitHubFileUrl(url: string): boolean {
+  const parsedUrl = new URL(url)
+  return parsedUrl.hostname === 'github.com' && url.includes('/blob/')
+}
+
+function readLocalFile(filePath: string): string {
+  if (!fs.existsSync(filePath)) {
+    throw new Error('Invalid input path. Please provide a valid file.')
+  }
+  return fs.readFileSync(filePath, 'utf8')
+}
+
+async function downloadGitHubRawContent(githubUrl: string): Promise<string> {
+  const rawFileUrl = githubUrl
+    .replace('github.com', 'raw.githubusercontent.com')
+    .replace('/blob', '')
+  return await downloadFile(rawFileUrl)
+}
+
+async function downloadFile(url: string): Promise<string> {
+  const response = await fetch(url)
+  if (!response.ok) {
+    throw new Error(`Failed to download file: ${response.statusText}`)
+  }
+  const data = await response.text()
+  return data
+}
+
+export async function getInputContent(inputPath: string): Promise<string> {
+  if (!isValidUrl(inputPath)) {
+    return readLocalFile(inputPath)
+  }
+
+  return isGitHubFileUrl(inputPath)
+    ? await downloadGitHubRawContent(inputPath)
+    : await downloadFile(inputPath)
+}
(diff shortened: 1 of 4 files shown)


frontend/packages/cli/src/cli/getInputContent.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export async function getInputContent(inputPath: string): Promise<string> {
+  if (!isValidUrl(inputPath)) {
+    return readLocalFile(inputPath)
+  }
+
+  return isGitHubFileUrl(inputPath)
+    ? await downloadGitHubRawContent(inputPath)
+    : await downloadFile(inputPath)
+}
```
hoshinotsuyoshi, line 49: It seems that the CLI is handling quite a lot under the `--input` argument, which might be somewhat unexpected for users.

Perhaps generating the raw URL could be left as the user's responsibility. I’d like to hear others’ opinions on this as well.

Example:

```
@liam-hq/cli erd build --input https://raw.githubusercontent.com/mastodon/mastodon/refs/heads/main/db/schema.rb --format schemarb
```

frontend/packages/cli/src/cli/getInputContent.ts
```diff
@@ -0,0 +1,49 @@
...
+import { URL } from 'node:url'
+
+function isValidUrl(url: string): boolean {
+  try {
+    new URL(url)
+    return true
+  } catch {
+    return false
+  }
+}
```
MH4GF, line 11: 💭 I thought URL.canParse() could be used, but it looks like this was only supported in Node.js in v19.
We have not considered the minimum version of Node.js to be supported by the CLI, but since the LTS is v18, it would be better to support that.
So it looks good as it is 👍🏻 

ref: https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static

frontend/packages/cli/src/cli/getInputContent.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export async function getInputContent(inputPath: string): Promise<string> {
+  if (!isValidUrl(inputPath)) {
+    return readLocalFile(inputPath)
+  }
+
+  return isGitHubFileUrl(inputPath)
+    ? await downloadGitHubRawContent(inputPath)
+    : await downloadFile(inputPath)
```
MH4GF, line 48: Only GitHub gets special handling, I see.

MH4GF (approved):
LGTM 👍🏻

frontend/packages/cli/src/cli/getInputContent.ts
hoshinotsuyoshi: > Perhaps generating the raw URL should be left to the user.  

For private repositories, GitHub's `raw` URL appears to include a pre-signed token as a `?token=` parameter.  

Therefore, this URL conversion feature would only support public repositories.   But it is ok, I think now.

hoshinotsuyoshi (approved):
LGTM. This is fine. While the functionality feels slightly extensive, it’s still perfectly reasonable.

frontend/packages/cli/src/cli/getInputContent.ts
sasamuku: Exactly, I was wondering if i should convert url.
If it is not useful and understandable for user, let's change implementation👍

Yeah, only public repo is available for now.

MH4GF: IMO: I thought slightly extensive too, but it was a minor point of contention, so once merged, it looked good.

PR #205, opened 2024-12-11
feat: Replace smooth step path with bezier path in RelationshipEdge component

MH4GF:
I switched to this one because it was potentially more visible.

Image


junkisai (approved):
Good! 👍

PR #207, opened 2024-12-11
feat: adjust view when active table changed

MH4GF:
https://github.com/user-attachments/assets/189b58c4-4fdb-49bd-8a31-e47970855e27


FunamaYukina (approved):
When I tried to check tables continuously, the Right Pane would close on the second click.
However, I think it's better to handle this in a separate issue👍.

https://github.com/user-attachments/assets/8733d43c-3bda-4b7f-b7f1-cdfc4d34bd75

MH4GF:
Oh!! Thanks for your information, I will check after the merge!

PR #208, opened 2024-12-11
feat: replace Comment div with DrawerDescription component and add aria-describedby for accessibility

MH4GF:
I've fixed following error:

Image


junkisai (approved):
I see that using DrawerDescription can solve the issue!

PR #209, opened 2024-12-11
fix: change stroke of RelationshipEdge

MH4GF:
Increased brightness of edge.

| Before | After |
|--------|--------|
| Image | Image |


junkisai (approved):
The contrast ratio has been improved, making the lines more visible than before!

PR #213, opened 2024-12-11
fix: simplify TableDetailDrawer state management

MH4GF:
## Description

Simplified state management in TableDetailDrawer.
This ensures that the TableNode is displayed even if it is clicked on in succession.

https://github.com/user-attachments/assets/68926485-9755-4fdc-b9f2-e66173f9f7a8

However, this has made it impossible to display the closing animation.
As it is difficult to reconcile the two, priority was given to opening them in succession this time.

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/TableDetailDrawer/TableDetailDrawer.tsx
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/TableDetailDrawer/TableDetailDrawer.tsx
@@ -4,34 +4,17 @@ import {
   useUserEditingStore,
 } from '@/stores'
 import { DrawerContent, DrawerPortal, DrawerRoot } from '@liam-hq/ui'
-import {
-  type FC,
-  type PropsWithChildren,
-  useCallback,
-  useEffect,
-  useState,
-} from 'react'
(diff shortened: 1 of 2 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/TableDetailDrawer/TableDetailDrawer.tsx
```diff
@@ -4,34 +4,17 @@ import {
...
-  const [open, setOpen] = useState(tableId !== undefined)
-
-  useEffect(() => {
-    setOpen(tableId !== undefined)
-  }, [tableId])
-
-  const handleClose = useCallback(() => {
-    setOpen(false)
-    // NOTE: Wait for the drawer to close before updating the active table name.
-    setTimeout(() => updateActiveTableName(undefined), ANIMATION_DURATION)
```
MH4GF, line 33: This delayed function was executing after switching the active table by clicking on another TableNode, which was then closed.

FunamaYukina (approved):
> However, this has made it impossible to display the closing animation.
> As it is difficult to reconcile the two, priority was given to opening them in succession this time.

I understand. That seems like a very sound judgment. I agree👍.

PR #220, opened 2024-12-12
Add cardinality icons next to handle

sasamuku:
## Summary

Add cardinality icons next to handle.

image

Highlighting will be next pr.


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableNode.module.css
```diff
@@ -74,3 +74,22 @@
 .handle[data-handlepos='left'] {
   transform: translate(-50%, 0%);
 }
+
+.handleDisabled {
+  opacity: 0;
+}
```
MH4GF, line 80: Q: Right now it seems that the handle always has an `opacity: 0`, but will the opacity change in the future depending on certain conditions?

If it is always `opacity: 0` in the future, it should be implemented in the `handle` class instead of the `handleDisabled` class.

MH4GF (approved):
LGTM, but I've wrote one comment!

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableNode.module.css
sasamuku: Maybe it will not be always `opacity: 0`.
`show Table Name` view mode might need default handle.
How do you think?

sasamuku: But it would be good to consider when implementing `show Table Name`.

MH4GF: By design, the cardinality is still shown for `show Table Name`, is this different?

スクリーンショット 2024-12-12 14 02 42

sasamuku: >By design, the cardinality is still shown for show Table Name, is this different?

Oh, really?
I was misunderstanding, sorry.

But in this case, how do we consider the node has more three `one_to_one` and `one_to_many` 🤔

sasamuku: Anyway, I've deleted handleDisabled class👍
74ace0cf70b3b1900df9747983a8589160473f00

MH4GF: > But in this case, how do we consider the node has more three one_to_one and one_to_many 🤔

That's right. We need to consider 💭

PR #221, opened 2024-12-12
Apply hover effect to related tables when hovered or active

FunamaYukina:
## Summary

Related tables are now highlighted when you hover or click on a TableNode.

https://github.com/user-attachments/assets/a19cfc7f-cdbf-43d5-99f7-30a8c3af9533

~preview: https://liam-erd-sample-92peiehb1-route-06-core.vercel.app/~

preview: https://liam-erd-sample-b2ey32y54-route-06-core.vercel.app/


FunamaYukina:
Due to some concerns I've found in the related tables processing logic, I'll be refactoring that part of the code.⚒️

FunamaYukina:
Hmm...I feel like the hover speed has slowed down a bit when there are many tables. We might need to restructure the relationships🤔.

https://github.com/user-attachments/assets/77cd2acf-7668-4b54-ae77-36331f917793

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
```diff
@@ -35,13 +37,27 @@ type Props = {
     | undefined
 }
 
+export const isRelatedToTable = (
+  relationships: Relationships,
+  tableName: string,
+  targetTableName: string | undefined,
+) =>
+  Object.values(relationships).some(
```
hoshinotsuyoshi, line 45: suggestion:

```suggestion
  targetTableName: string | undefined,
) =>
  if (!targetTableName) return
  Object.values(relationships).some(
```

might be faster if targetTableName is undefined often .

hoshinotsuyoshi:
> Hmm...I feel like the hover speed has slowed down a bit when there are many tables. We might need to restructure the relationships🤔.


Or should we use `useMemo` or something? I created a PoC branch and pushed it.

- 705a5db: PoC using `useMemo`
- 043a395: For testing;

PoC:  https://liam-erd-sample-ln46w292h-route-06-core.vercel.app/

That said, it doesn’t feel significantly faster. 😇

hoshinotsuyoshi (approved):
From what I can see, the logic looks correct, and the functionality seems to be working well. LGTM. I haven’t edited this package recently, so I’d appreciate it if someone else could also take a look! :)

MH4GF (approved):
hoshino-san's points seem to be acceptable to incorporate, but otherwise LGTM!

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
FunamaYukina: That's right! Thank you!🙏
Fixed: https://github.com/liam-hq/liam/pull/221/commits/db4dbc262b36b9acbca9f0e656a8028a8fb55d84

FunamaYukina:
Thank you very much for your detailed review🙏. I've addressed all the comments, and I'm ready to merge this pull request.

PR #223, opened 2024-12-12
feat: add zIndex property to node style in convertDBStructureToNodes

junkisai:
## Summary

I ensured that Nodes are always placed on top of Edges.

| Before | After |
|--------|--------|
|スクリーンショット 2024-12-12 14 25 55 |スクリーンショット 2024-12-12 14 25 20 |


MH4GF (approved):
Looks quite good!

PR #225, opened 2024-12-12
feat: move toolbar to ERDRenderer and disable toolbar in Related Tables

MH4GF:
https://github.com/user-attachments/assets/a8cd03ca-81c2-4921-b0c1-19402475e75d


hoshinotsuyoshi (approved):
LGTM!

PR #231, opened 2024-12-12
feat: show error when parse postgres schema

MH4GF:
- **refactor: Refactor SQL parser types to use ParseResult**
- **feat: Add UnexpectedTokenWarningError and enhance error handling in SQL processor**
- **refactor: Remove unnecessary null check for statements in SQL converter**

### DIFF
--- a/frontend/packages/db-structure/src/parser/sql/postgresql/index.ts
+++ b/frontend/packages/db-structure/src/parser/sql/postgresql/index.ts
@@ -1,4 +1,5 @@
 import type { DBStructure } from '../../../schema/index.js'
+import { UnexpectedTokenWarningError } from '../../errors.js'
 import type { Processor } from '../../types.js'
 import { convertToDBStructure } from './converter.js'
 import { mergeDBStructures } from './mergeDBStructures.js'
@@ -8,12 +9,17 @@ import { processSQLInChunks } from './processSQLInChunks.js'
 export const processor: Processor = async (str: string) => {
   const dbStructure: DBStructure = { tables: {}, relationships: {} }
   const CHUNK_SIZE = 1000
+  const errors: Error[] = []
 
   await processSQLInChunks(str, CHUNK_SIZE, async (chunk) => {
-    const parsed = await parse(chunk)
-    const partial = convertToDBStructure(parsed)
+    const { parse_tree, error } = await parse(chunk)
+    const partial = convertToDBStructure(parse_tree.stmts)
     mergeDBStructures(dbStructure, partial)
+
+    if (error !== null) {
+      errors.push(new UnexpectedTokenWarningError(error.message))
+    }
   })
 
-  return { value: dbStructure, errors: [] }
+  return { value: dbStructure, errors }
 }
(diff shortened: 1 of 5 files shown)


frontend/packages/db-structure/src/parser/sql/postgresql/parser.ts
```diff
@@ -3,14 +3,25 @@ import type { RawStmt } from '@pgsql/types'
...
   const pgQuery = await new Module()
   const result = pgQuery.parse(str)
   return result
 }
 
-// It was expected that postgresParse would return a ParseResult object,
-// but it was found that an array of RawStmtWrapper objects was returned.
-export interface RawStmtWrapper {
-  RawStmt: RawStmt
+// NOTE: pg-query-emscripten does not have types, so we need to define them ourselves
```
hoshinotsuyoshi, line 12: 👍 👍 👍

frontend/packages/db-structure/src/parser/sql/postgresql/parser.ts
```diff
@@ -3,14 +3,25 @@ import type { RawStmt } from '@pgsql/types'
...
+  }
+  stderr_buffer: string
+  error: {
+    message: string
+    funcname: string
+    filename: string
+    lineno: number
+    cursorpos: number
+    context: string
+  } | null
```
sasamuku, line 26: Is this correct type from `pg-query-emscripten`?
This means `RawStmtWrapper` was wrong type?

MH4GF: This is defined from the output of the pg-query-emscripten run.

> This means RawStmtWrapper was wrong type?

It was probably broken. All the use areas were `@ts-expect-error` .

frontend/packages/db-structure/src/parser/sql/postgresql/index.ts
```diff
@@ -8,12 +9,17 @@ import { processSQLInChunks } from './processSQLInChunks.js'
...
 
   await processSQLInChunks(str, CHUNK_SIZE, async (chunk) => {
-    const parsed = await parse(chunk)
-    const partial = convertToDBStructure(parsed)
+    const { parse_tree, error } = await parse(chunk)
+    const partial = convertToDBStructure(parse_tree.stmts)
     mergeDBStructures(dbStructure, partial)
+
+    if (error !== null) {
+      errors.push(new UnexpectedTokenWarningError(error.message))
```
hoshinotsuyoshi, line 20: ask: In this case, are the errors always limited to "UnexpectedToken"? I was wondering if there might be other possibilities.

frontend/packages/db-structure/src/parser/sql/postgresql/parser.ts
sasamuku: I understood, thank you for checking!

sasamuku (approved):
LGTM🚀

frontend/packages/db-structure/src/parser/sql/postgresql/index.ts
MH4GF: Perhaps that is true, and there are other possibilities.
However, we do not know what errors may occur at present, so we use this approach.
It is assumed that an error class is added together with a test case when another error is known.

hoshinotsuyoshi: Thank you, I thought the policy was good. When parsing fails, it can be said that an unexpected token has arrived.

hoshinotsuyoshi (approved):
LGTM!

PR #232, opened 2024-12-12
debug(docs): Add page with various components and examples

MH4GF:
To adjust the design, a page is provided where all components can be viewed once.
This will be removed before release.

スクリーンショット 2024-12-12 20 02 11

ref: https://liam-docs-k3x9juyjw-route-06-core.vercel.app/docs/debug


junkisai (approved):
thx 👍

PR #233, opened 2024-12-12
feat: enable hover effect to show full table name in TableHeader

hoshinotsuyoshi:
## Summary

https://github.com/user-attachments/assets/d59c824e-1caf-427c-88e8-01102e8de5d7


hoshinotsuyoshi:
I’ve reconsidered. I’d like to choose the approach from https://github.com/liam-hq/liam/pull/237.

PR #234, opened 2024-12-12
refactor: reduce throwing errors

MH4GF:
Stop throwing errors and return the error class to the caller.
This makes it easier to take error handling into account on the CLI side. For example, adjusting the log level.

### DIFF
--- a/frontend/packages/db-structure/src/parser/schemarb/parser.ts
+++ b/frontend/packages/db-structure/src/parser/schemarb/parser.ts
@@ -27,7 +27,11 @@ import type {
   Tables,
 } from '../../schema/index.js'
 import { aColumn, aRelationship, aTable, anIndex } from '../../schema/index.js'
-import { type ProcessError, WarningError } from '../errors.js'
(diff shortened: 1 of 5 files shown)


frontend/packages/db-structure/src/parser/schemarb/parser.ts
```diff
@@ -248,19 +252,27 @@ function extractDefaultValue(
...
-    if (node instanceof StringNode) return node.unescaped.value
-    return null
-  })
+  const [foreignTableName, primaryTableName] = stringNodes.map(
+    (node): string => {
+      // @ts-expect-error: unescaped is defined as string but it is actually object
+      if (node instanceof StringNode) return node.unescaped.value
+      return ''
+    },
+  ) as [string, string]
```
MH4GF, line 273: An as cast was required here, because TypeScript cannot infer that it is a tuple of strings.

sasamuku (approved):
Thank you for refactoring🙏

PR #235, opened 2024-12-13
feat: update ELK layout options for improved node placement and spacing

junkisai:
## Summary

I adjusted the layout so that the nodes are arranged from left to right.

スクリーンショット 2024-12-13 18 29 33


MH4GF (approved):
I haven't been able to check the settings properly, but the results of running it look good!

PR #236, opened 2024-12-13
feat: Enable panning on scroll

sasamuku:
## Summary

Enable Figma-like Viewport Controls:

* pan: Space + scroll, middle or right mouse
* zoom: pitch or cmd + scroll
* create selection: drag mouse

ref: https://reactflow.dev/learn/concepts/the-viewport#figma-like-viewport-controls

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
@@ -181,6 +181,8 @@ export const ERDContent: FC<Props> = ({
     [setEdges],
   )
 
+  const panOnDrag = [1, 2]
+
   return (
     <div className={styles.wrapper}>
       <ReactFlow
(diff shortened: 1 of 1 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
```diff
@@ -181,6 +181,8 @@ export const ERDContent: FC<Props> = ({
     [setEdges],
   )
 
+  const panOnDrag = [1, 2]
```
sasamuku, line 184: [1, 2] allows panning with the middle and right mouse buttons.

https://github.com/xyflow/xyflow/blob/0f21ec21ebf01567ec1de4fad1d0c55f40db3463/packages/react/src/types/component-props.ts#L350-L356

MH4GF (approved):
After touching the preview environment, I thought it would be quite useful once I got used to it!
LGTM 🚀

PR #248, opened 2024-12-13
feat: Add url query params for quick access

sasamuku:
## Summary

Add url query params for quick access.

https://github.com/user-attachments/assets/fca71107-bc93-4466-a720-74c41dd37935

### DIFF
--- a/frontend/.changeset/stupid-insects-sneeze.md
+++ b/frontend/.changeset/stupid-insects-sneeze.md
@@ -1,0 +1,6 @@
+---
+"@liam-hq/erd-core": patch
+"@liam-hq/cli": patch
+---
+
+Add url query params for quick access
--- a/frontend/.changeset/swift-worms-retire.md
+++ b/frontend/.changeset/swift-worms-retire.md
@@ -1,0 +1,6 @@
+---
+"@liam-hq/erd-core": patch
+"@liam-hq/cli": patch
+---
+
+Enable setting active table from url query parameter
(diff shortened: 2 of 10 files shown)


.changeset/swift-worms-retire.md
MH4GF, line 1: I thought one PR and one changeset would be fine, what do you think? cc @hoshinotsuyoshi

MH4GF:
@sasamuku Looks good, but I'm concerned about the lack of focus when it opens!

https://github.com/user-attachments/assets/1e78b5b5-03bb-4d85-9e23-66fbcba5fd58

Please consider this.

MH4GF:
It was safe to put in the appropriate table name. 👍🏻 

https://liam-erd-sample-6swvq5ps4-route-06-core.vercel.app/?active=not-found-table

MH4GF (approved):
LGTM 🚀 
nice work!

.changeset/swift-worms-retire.md
sasamuku: This is first time to write changeset, and it was quite hard work...

I'm not sure the best practice to write changeset, but one pr one changeset would be good for us and our contributors.

MH4GF: I agree with Hoshino-san, that changeset is unnecessary for every commit because it is better to write for the user rather than for the contributor!

hoshinotsuyoshi:
@sasamuku 
Apologies for the inconvenience! I previously mentioned that the `.changeset` directory would be located at `./.changeset`, but it has now been moved back to `./frontend/.changeset`. For more details, please refer to https://github.com/liam-hq/liam/pull/257. I’d greatly appreciate it if you could move the files to `./frontend/.changeset`. Thank you so much!

sasamuku:
>I'm concerned about the lack of focus when it opens!

I'll fix this another PR🚀

PR #249, opened 2024-12-13
refactor: optimize edge highlighting using useReactFlow hooks

junkisai:
## Summary

When there is only one target for updating, it is simpler to use the updateEdge function instead of the setEdges function, so I replaced it with the updateEdge function.

ref. https://github.com/xyflow/xyflow/discussions/4496#discussioncomment-10189904

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
@@ -10,6 +10,7 @@ import {
   ReactFlow,
   useEdgesState,
   useNodesState,
+  useReactFlow,
 } from '@xyflow/react'
 import { type FC, useCallback, useEffect } from 'react'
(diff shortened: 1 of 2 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
```diff
@@ -153,32 +155,18 @@ export const ERDContent: FC<Props> = ({
...
     (_, { id }) => {
-      setEdges((edges) =>
-        edges.map((e) =>
-          e.id === id
-            ? { ...e, animated: true, data: { ...e.data, isHighlighted: true } }
-            : e,
-        ),
-      )
+      updateEdge(id, { animated: true })
+      updateEdgeData(id, { isHighlighted: true })
```
MH4GF, line 159: I didn't know, this is amazing!!!

MH4GF (approved):
LGTM 😄

PR #250, opened 2024-12-13
fix: disable delete key functionality in ERDContent component

MH4GF:
I was reading the react flow documentation and noticed that the Node disappeared when I pressed the delete key during selection 😂

https://github.com/user-attachments/assets/b56c3fd3-809e-44ed-b696-5440b79df7e1

ref: https://reactflow.dev/api-reference/react-flow#keyboard-props

That functionality is not needed and has been turned off.


junkisai (approved):
wow!
There are likely quite a few React Flow users who are unaware of this specification.😂

PR #258, opened 2024-12-13
feat: Added the functionality to display only the selected table from TableDetail in the Main Pane

MH4GF:
https://github.com/user-attachments/assets/0e8950f8-c038-4712-bdd9-30bf9efd5595


junkisai (approved):
Good!! 🎉

PR #259, opened 2024-12-13
fix: reduce button-in-button

MH4GF:


junkisai (approved):
thx!!
It was a console error I was concerned about during development!

PR #261, opened 2024-12-16
feat: add script to update dist content for Vercel ERD-sample deploy

hoshinotsuyoshi:
## Summary

Added script to update dist content for Vercel ERD-sample deploy.
Prepared for future integration of internal analytics tool tags by adding the capability to inject custom content into the build process


MH4GF (approved):
This is excellent!
thanks 🚀

PR #263, opened 2024-12-16
refactor: remove unused convertElkEdgesToEdges function

MH4GF:
From the looks of it, the edges returned by elk.layout() do not change any particular information, so setEdges was removed.
The motivation is to reduce re-rendering as much as possible due to performance issues when it comes to large ER diagrams.

The only information that changed was the opacity, but in the case of edge, it seemed to have been displayed without changing even if `opacity: 0`.


MH4GF:
@junkisai I would like you to review!

junkisai (approved):
Indeed, it seems that the setEdges in this part is unnecessary!

PR #267, opened 2024-12-16
Fixed excessive highlighting of cardinality elements.

FunamaYukina:
## Summary

Fixed excessive highlighting of cardinality elements.

Before
ss 2499

After

ss 2500


MH4GF (approved):
Thanks 👍🏻

PR #268, opened 2024-12-16
Use version from `package.json` in Liam CLI

hoshinotsuyoshi:
## Summary

Refactored the CLI initialization to dynamically fetch the version number from `package.json` using `createRequire` from Node.js. This ensures the CLI version is automatically synchronized with the package version.

### DIFF
--- a/frontend/packages/cli/src/cli/index.test.ts
+++ b/frontend/packages/cli/src/cli/index.test.ts
@@ -1,3 +1,4 @@
+import { createRequire } from 'node:module'
 import type { Command } from 'commander'
 import { describe, expect, it, vi } from 'vitest'
 import { buildCommand } from './commands/index.js'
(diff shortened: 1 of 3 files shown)


frontend/packages/cli/src/cli/index.test.ts
```diff
@@ -21,6 +22,9 @@ describe('program', () => {
   it('should have the correct name and description', () => {
     expect(program.name()).toBe('liam')
     expect(program.description()).toBe('CLI tool for Liam')
+    const require = createRequire(import.meta.url)
+    const { version: packageVersion } = require('../../package.json')
+    expect(program.version()).toBe(packageVersion)
```
hoshinotsuyoshi, line 27: Since this is a test, it would be preferable to use a literal value for `packageVersion`. However, considering that it is a value that gets updated frequently, I chose not to do so.

sasamuku (approved):
LGTM🚀

PR #271, opened 2024-12-16
fix: Fixed an issue where edges were displayed during the initial loading

MH4GF:
Use React Context to define the state during loading.
As a side effect, the problem of the edges being displayed during initial loading is resolved.

## before

https://github.com/user-attachments/assets/742fdfc9-0c1c-48d7-bc23-ff59fd53fac6

## after

https://github.com/user-attachments/assets/52dc0fb6-c187-4f27-8330-67a0137a6831

## Note

- Nodes that are momentarily in the wrong position when displayed are displayed, but this is currently being investigated as a fix.
- Loading animations will be added later.


junkisai (approved):
I was able to confirm that the issue has been resolved!
It's also good that we were able to move the style code into the component. 👍

PR #274, opened 2024-12-16
refactor: reduce useEffect usage in ERDContent component

MH4GF:
Although it does not contribute to performance, useEffect has been removed because less useEffect is desirable.


sasamuku (approved):
Looks great!🚀

PR #276, opened 2024-12-16
fix: Fixed an issue where URL sharing in TableDetail sometimes did not work correctly

MH4GF:
The feature to open TableDetail when opened with a URL with a query parameter did not work in some cases, which will be corrected.

https://liam-erd-sample-pzarkcxaz-route-06-core.vercel.app/?active=featured_tags

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/TableDetailDrawer/TableDetailDrawer.tsx
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/TableDetailDrawer/TableDetailDrawer.tsx
@@ -1,7 +1,7 @@
 import {
   updateActiveTableName,
   useDBStructureStore,
-  useUserEditingStore,
+  useUserEditingActiveStore,
 } from '@/stores'
 import { DrawerContent, DrawerPortal, DrawerRoot } from '@liam-hq/ui'
 import type { FC, PropsWithChildren } from 'react'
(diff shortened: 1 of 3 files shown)


frontend/packages/erd-core/src/stores/userEditing/hooks.ts
```diff
@@ -2,3 +2,5 @@ import { useSnapshot } from 'valtio'
 import { userEditingStore } from './store'
 
 export const useUserEditingStore = () => useSnapshot(userEditingStore)
+export const useUserEditingActiveStore = () =>
+  useSnapshot(userEditingStore.active)
```
MH4GF, line 6: If this was not done, there was a problem with it not re-rendering. 🤔

frontend/packages/erd-core/src/components/ERDRenderer/TableDetailDrawer/TableDetailDrawer.tsx
```diff
@@ -12,9 +12,9 @@ import styles from './TableDetailDrawer.module.css'
 const handleClose = () => updateActiveTableName(undefined)
 
 export const TableDetailDrawerRoot: FC<PropsWithChildren> = ({ children }) => {
-  const {
-    active: { tableName },
-  } = useUserEditingStore()
+  const { tableName } = useUserEditingActiveStore()
+  const { tables } = useDBStructureStore()
+  const open = Object.keys(tables).length > 0 && tableName !== undefined
```
MH4GF, line 17: Had to wait to see if loading was complete from schema.json.

sasamuku (approved):
Thanks for fixing🙌

PR #279, opened 2024-12-16
Refactor hidden node cardinalities

sasamuku:
## Summary

When hiding node, cardinalities also should be hidden.
The case of using hidden tables function from Related Tables will be another PR.

### before

https://github.com/user-attachments/assets/90d3424f-c1a5-4d8f-b42e-f419043a6ac4

### after

https://github.com/user-attachments/assets/e3fa2a85-bb55-47e7-bfee-354b73986cb5


sasamuku:
Close this PR due to a lot of differences in other PRs.

PR #278, opened 2024-12-16
Optimize TableColumn component by caching relationship checks

hoshinotsuyoshi:
## Summary

Optimize TableColumn component by caching relationship checks.
Introduced useMemo to cache `isSource` and `targetCardinality` computations.

## Experiment

Triggered `handleMouseLeaveNode` manually on the accounts table node to measure performance impact.

<details><summary>Demo</summary>

https://github.com/user-attachments/assets/9cde8ec3-49ca-455a-afe5-a03f751fba1d

</details>

#### 1.  `ERDContentInner`

before: ~660 ms
after: ~170 ms

|  | before 90fae17723f7 | after f843a99746a2c37  |
|--------|--------|--------|
|  | main1-sk  | hoshino1-sk |

#### 2. `TableColumn`: `account_aliases` table > `id` column

before: ~0.6 ms
after: ~0.1 ms

|  | before 90fae17723f7 | after  f843a99746a2c37 |
|--------|--------|--------|
|  | main2-sk |  hoshino2-sk | 

## Testing

N/A

## Other Information

N/A

PR #280, opened 2024-12-16
refactor: Reduced performance degradation caused by calculations for source and target

MH4GF:
- Continued from https://github.com/liam-hq/liam/pull/278

The source and handle are now calculated only once at the beginning.
In addition to TableColumn, calculated values are now used in TableHeader.

## Experiment

Triggered `handleMouseLeaveNode` manually on the accounts table node to measure performance impact. (It is same as #278 )
I then checked the Ranked tab in the React Dev tool.

### Before

TableHeader calculations are at the top of the list.

スクリーンショット 2024-12-16 20 51 55

### After

TableHeader calculations have disappeared from the top!

スクリーンショット 2024-12-16 20 52 27

### DIFF
--- a/frontend/packages/db-structure/src/schema/dbStructure.ts
+++ b/frontend/packages/db-structure/src/schema/dbStructure.ts
@@ -43,6 +43,8 @@ const tableSchema = v.object({
 export type Table = v.InferOutput<typeof tableSchema>
 
 const cardinalitySchema = v.picklist(['ONE_TO_ONE', 'ONE_TO_MANY'])
+export type Cardinality = v.InferOutput<typeof cardinalitySchema>
+
 const foreignKeyConstraintSchema = v.picklist([
   'CASCADE',
   'RESTRICT',
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableColumnList/TableColumn/Cardinality/Cardinality.tsx
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableColumnList/TableColumn/Cardinality/Cardinality.tsx
@@ -1,3 +1,4 @@
+import type { Cardinality as CardinalityType } from '@liam-hq/db-structure'
 import {
   CardinalityZeroOrManyLeftIcon,
   CardinalityZeroOrOneLeftIcon,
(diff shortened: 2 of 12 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableColumnList/TableColumn/Cardinality/Cardinality.tsx
```diff
@@ -9,7 +10,7 @@ import styles from './Cardinality.module.css'
 
 type CardinalityProps = {
   direction: 'left' | 'right'
-  cardinality: 'ONE_TO_ONE' | 'ONE_TO_MANY'
+  cardinality: CardinalityType
```
MH4GF, line 13: Modified to import and use types from `@liam-hq/db-structure`.

frontend/packages/db-structure/src/schema/dbStructure.ts
```diff
@@ -43,6 +43,8 @@ const tableSchema = v.object({
 export type Table = v.InferOutput<typeof tableSchema>
 
 const cardinalitySchema = v.picklist(['ONE_TO_ONE', 'ONE_TO_MANY'])
+export type Cardinality = v.InferOutput<typeof cardinalitySchema>
```
hoshinotsuyoshi, line 46: 👍

frontend/packages/erd-core/src/components/ERDRenderer/convertDBStructureToNodes.ts
```diff
@@ -16,13 +16,30 @@ export const convertDBStructureToNodes = ({
...
+  for (const relationship of relationships) {
+    sourceColumns.set(
+      relationship.primaryTableName,
+      relationship.primaryColumnName,
+    )
+    tableColumnCardinalities.set(relationship.foreignTableName, {
+      ...tableColumnCardinalities.get(relationship.foreignTableName),
+      [relationship.foreignColumnName]: relationship.cardinality,
+    })
+  }
```
hoshinotsuyoshi, line 33: I do believe this is a good change (movement). If my memory serves me right, `convertDBStructureToNodes` is always called, though it happens infrequently.

hoshinotsuyoshi (approved):
LGTM!

frontend/packages/erd-core/src/components/ERDRenderer/convertDBStructureToNodes.ts
MH4GF: Yes, it is called once before rendering the React Flow!

PR #282, opened 2024-12-17
✨: Changed the default show mode to 'Table name only'

MH4GF:
The default mode has been changed to reduce performance issues and affect the user's first-time experience.


hoshinotsuyoshi (approved):
LGTM!

PR #284, opened 2024-12-17
Highlight related edges and cardinalities when a TableNode is active.

FunamaYukina:
## Summary

Highlight related edges and cardinalities when a TableNode is active.
Preview: https://liam-erd-sample-gx3t0p2yn-route-06-core.vercel.app/

### Before
ss 2506

### After
ss 2507

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
@@ -65,7 +65,7 @@ export const ERDContentInner: FC<Props> = ({
   const {
     state: { loading },
   } = useERDContentContext()
-  const [clickedNodeId, setClickedNodeId] = useState<string | null>(null)
+  const [activeNodeId, setActiveNodeId] = useState<string | null>(null)
 
   useInitialAutoLayout()
   useActiveTableNameFromUrl()
@@ -75,7 +75,7 @@ export const ERDContentInner: FC<Props> = ({
 
   const handleNodeClick = useCallback(
     (nodeId: string) => {
-      setClickedNodeId(nodeId)
+      setActiveNodeId(nodeId)
 
       const relatedEdges = edges.filter(
         (e) => e.source === nodeId || e.target === nodeId,
@@ -136,7 +136,7 @@ export const ERDContentInner: FC<Props> = ({
   )
 
   const handlePaneClick = useCallback(() => {
-    setClickedNodeId(null)
+    setActiveNodeId(null)
 
     const updatedEdges = edges.map((e) => ({
       ...e,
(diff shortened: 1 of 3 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/type.ts
```diff
@@ -8,6 +8,7 @@ export type Data = {
   highlightedHandles: string[]
   sourceColumnName: string | undefined
   targetColumnCardinalities?: Record<string, Cardinality> | undefined
+  onClick?: (tableName: string) => void
```
MH4GF, line 11: This does not appear to have been used from TableNode, was it necessary?

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
```diff
@@ -65,47 +65,66 @@ export const ERDContentInner: FC<Props> = ({
   const {
     state: { loading },
   } = useERDContentContext()
+  const [clickedNodeId, setClickedNodeId] = useState<string | null>(null)
```
MH4GF, line 68: It is painful to have to create a separate state from the `active` state managed by `useUserEditingStore()`, but it seems that the logic has to be unravelled to solve the problem. I'd like to try another PR.

MH4GF (approved):
The highlighting process is getting more complicated, but it looks good!

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
FunamaYukina: You're absolutely right. I wanted to write the code more simply, but I couldn't. In the end, the code became more complex🥲.Thank you for thinking about it🙏.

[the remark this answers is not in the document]
hoshinotsuyoshi: I agree that it is painful. I was thinking about checking whether the changes we made to how we handle data around PR #280 could be utilized. The functionality seems good, so I think it's fine to merge it now and integrate later!

hoshinotsuyoshi (approved):

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/type.ts
FunamaYukina: Oh,sorry, you are right.
I've removed the unnecessary props🙏. https://github.com/liam-hq/liam/pull/284/commits/a518be68e426f6ad81da6cd613249c18448c2738

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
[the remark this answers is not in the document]
FunamaYukina: Thank you for your suggestion! I agree!
I've fixed it. https://github.com/liam-hq/liam/pull/284/commits/9c8767ee97a49d4aae09eb5fbc4eacfe95605b7a

FunamaYukina:
Thank you all for your reviews.🙏

PR #286, opened 2024-12-17
feat: loading animation

MH4GF:
Added loading animation.
However, it is now almost impossible to see animation, because due to the fast loading of the ER diagram and the slow loading of the animation.
The following video shows the display when the system is in a constant loading state as debugging.

https://github.com/user-attachments/assets/0bfc6fde-1e2f-4413-a793-a337f5e05bc7

The issue of slow loading is checked at in another PR.


frontend/packages/cli/index.html
```diff
@@ -5,6 +5,12 @@
     <link rel="icon" href="/public/favicon.ico" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Liam ERD</title>
+    <link
+      rel="preload"
+      href="/loading.riv"
+      as="fetch"
+      crossorigin="anonymous"
+    />
```
MH4GF, line 13: 📝 Essentially unnecessary, but preloaded to speed up loading.
ref: [Optimization techniques for Rive animations in React apps - DEV Community](https://dev.to/alex_barashkov/optimization-techniques-for-rive-animations-in-react-apps-1a8p)

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/Loading/Loading.tsx
```diff
@@ -0,0 +1,14 @@
+import { useRive } from '@rive-app/react-canvas'
+import type { ComponentProps, FC } from 'react'
+
+type Props = ComponentProps<'canvas'>
+
+export const Loading: FC<Props> = (props) => {
+  const { RiveComponent } = useRive({
+    src: '/loading.riv',
```
MH4GF, line 8: Currently it works because it is bundled with @liam-hq/cli vite, but if you want to call @liam-hq/erd-core from another part, it will stop working.
Essentially, it should be better to load via a CDN.

MH4GF:
Give up on Rive as it is slow to load and does not seem to work for initial loading.

PR #290, opened 2024-12-17
Restored columnType visibility

FunamaYukina:
## Summary

Resolved the problem that prevented the **column type** from being shown on TableNode hover.

### Before
ss 2510

### After
ss 2511

## Other Information

In contrast to the design, where column types are shown for all relevant TableNodes on hover, this implementation displays the column type only for the hovered TableNode.（In order to fix the bug quickly.）

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableColumnList/TableColumn/TableColumn.module.css
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableColumnList/TableColumn/TableColumn.module.css
@@ -48,7 +48,7 @@
   transition: opacity 300ms var(--default-timing-function);
 }
 
-.wrapper:hover .columnType {
+:global([data-erd='table-node']):hover .columnType {
   opacity: 1;
 }
 
--- a/frontend/.changeset/twenty-eggs-beam.md
+++ b/frontend/.changeset/twenty-eggs-beam.md
@@ -1,0 +1,6 @@
+---
+"@liam-hq/erd-core": patch
+"@liam-hq/cli": patch
+---
+
+Restored columnType visibility.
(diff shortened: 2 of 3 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableColumnList/TableColumn/TableColumn.module.css
```diff
@@ -48,7 +48,7 @@
   transition: opacity 300ms var(--default-timing-function);
 }
 
-.wrapper:hover .columnType {
```
sasamuku, line 51: Let me know why this is not working ?👀

sasamuku (approved):
LGTM🚀

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableColumnList/TableColumn/TableColumn.module.css
FunamaYukina: I didn't know about that either, but I've just been informed(Mr. Miyagi told me). 
You're attempting to access a class named "columnType" from TableNode, 

```
TableNode.module.css

.wrapper:hover .columnType {
  opacity: 1;
}
```

but the real class name is obfuscated. **(it is not ``columnType``)**

ss 2513




As a result, you couldn't reference the class from another file. I've resolved this by utilizing the data attribute and ``global:.``

FunamaYukina:
@MH4GF 
Thank you for your previous guidance🙏. I've made the corrections, so I would be grateful if you could review them.

commit: https://github.com/liam-hq/liam/pull/290/commits/296fdaae2e084fadcae375b72e4cc7b09a479b16

preview: https://liam-erd-sample-6h1szphmk-route-06-core.vercel.app/

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableColumnList/TableColumn/TableColumn.module.css
sasamuku: Thank you for explaining.
`data-*` attributes can be used in that situation, I see 👍

FunamaYukina: Yes😆 This time, we were able to fix the issue by only modifying the CSS.

PR #293, opened 2024-12-17
Resolving the issue of remaining highlights

FunamaYukina:
## Summary

I've resolved the problem where irrelevant TableNodes, Edges, and Cardinalities were being unexpectedly highlighted.
It often happens when moving the mouse pointer from node to node in an ERD with many tables.↖️

### Before

ss 2535

### After

ss 2536

## Testing

Please confirm if this issue still persists👀.
Preiew: https://liam-erd-sample-k8i0nsnmb-route-06-core.vercel.app/


FunamaYukina:
I'm going to fix the conflict🙏.

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
```diff
@@ -164,44 +177,82 @@ export const ERDContentInner: FC<Props> = ({
...
+        (e) => e.source === activeNodeId || e.target === activeNodeId,
       )
 
+      const updatedEdges = edges.map((e) => {
+        if (relatedEdges.includes(e)) {
+          return highlightEdge(e)
+        }
+        if (relatedToActiveNodeEdges.includes(e)) {
+          return highlightEdge(e)
+        }
```
MH4GF, line 190: Is this the solution to the PR title cause?

MH4GF (approved):
The content looked good, please just resolve the conflicts!

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
FunamaYukina: I suspect that ``return unhighlightEdge(e)`` is referring to the code in the next line.
It seems that the ``unhighlightEdge`` function was not working correctly when the hovered TableNode changed.

FunamaYukina:
I've checked the preview and it looks good👀, so I'm going to merge it🚀.
 Please let me know if you have any concerns🙏.

PR #294, opened 2024-12-17
Enable hiding cardinalities on source node

sasamuku:
## Summary

Enable hiding cardinalities on source node if target node is hidden.

Uploading 画面収録 2024-12-17 19.40.28.mov…


MH4GF (approved):
Thanks, LGTM 👍🏻

PR #295, opened 2024-12-17
refactor: Move the calculation to TableColumnList and TableColumn only displays the props

MH4GF:
- **refactor: Refactor handle ID generation for table columns to use a dedicated function**
- **refactor: Simplify TableColumn component by passing handleId as a prop and removing unnecessary highlightedHandles logic**
- **maintenance: add changeset**

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableColumnList/TableColumn/TableColumn.tsx
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableColumnList/TableColumn/TableColumn.tsx
@@ -8,6 +8,7 @@ import { Handle, Position } from '@xyflow/react'
 import clsx from 'clsx'
 import type { FC } from 'react'
 import { match } from 'ts-pattern'
+import { columnHandleId } from '../../../../columnHandleId'
 import { Cardinality } from './Cardinality'
 import { CardinalityNotation } from './CardinalityNotation'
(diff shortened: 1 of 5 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/columnHandleId.ts
```diff
@@ -0,0 +1,2 @@
+export const columnHandleId = (tableName: string, columnName: string) =>
+  `${tableName}-${columnName}`
```
MH4GF, line 2: 📝 As it was defined in several places, it was cut out into a function to ensure that it was the same logic.

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableColumnList/TableColumn/TableColumn.tsx
```diff
@@ -77,16 +72,12 @@ export const TableColumn: FC<TableColumnProps> = ({
           <Cardinality
             direction="right"
             cardinality="ONE_TO_ONE"
-            isHighlighted={
-              isHighlighted || highlightedHandles.includes(handleId)
-            }
+            isHighlighted={isHighlighted}
```
MH4GF, line 75: 📝 The loop in `highlightedHandles.includes(handleId)` was being executed multiple times, so it was summarised.

sasamuku (approved):
LGMT👍👍

PR #299, opened 2024-12-18
refactor: The behavior of TableNode when clicked is unified to be handled by ERDContent

MH4GF:
It is quite minimal, I have been modified to match the way React Flow is implemented.


FunamaYukina (approved):
LGTM!
I noticed the cursor has changed, but it doesn't seem to be affecting anything major.👀

ss 2526

MH4GF:
@FunamaYukina Yeah, but it is a minor issue!

PR #302, opened 2024-12-18
:mag: erd-sample: Add meta tag `robots` to prevent search engines from indexing

hoshinotsuyoshi:
## Summary

erd-sample: Add meta tag `robots` to prevent search engines from indexing

## Testing

see preview: https://liam-erd-sample-fke6r5p72-route-06-core.vercel.app

html source:

スクリーンショット 2024-12-18 13 10 05


MH4GF (approved):
Thanks you!!!

PR #303, opened 2024-12-18
💄 feat: add nuance shadow for TableNode

kumanoayumi:
## Summary

Added shadow to Table Node to add a floating nuance to give the impression that it is not too attached to an unrelated Edge.

https://github.com/user-attachments/assets/78e1f4ef-dfa7-4269-90fe-3341f2c87d65

## Changes

- Add nuance shadow for TableNode.


MH4GF (approved):
good, It is considerably easier to see!

PR #305, opened 2024-12-18
📈 : add GTM utility functions and types for click logging

junkisai:


frontend/packages/erd-core/src/features/gtm/types.ts
```diff
@@ -0,0 +1,3 @@
+export type CommonLogEvent = {
+  appEnv: string
+}
```
MH4GF, line 3: This doesn't look like it's being used. 👀

frontend/packages/erd-core/src/features/gtm/utils/pushToDataLayer.ts
```diff
@@ -0,0 +1,10 @@
+type WindowWithDataLayer = Window & {
+  dataLayer: Record<string, unknown>[]
+}
+
+declare const window: WindowWithDataLayer
+
+export const pushToDataLayer = (obj: Record<string, unknown>) => {
+  window.dataLayer = window.dataLayer || []
+  window.dataLayer.push(obj)
+}
```
MH4GF, line 10: It's easy to clean up because we just need to delete the features/gtm directory and fix the compile errors! 👍🏻

MH4GF (approved):
LGTM, but please check a comments!

frontend/packages/erd-core/src/features/gtm/types.ts
junkisai: The implementation of common logs is planned to be addressed in a separate issue, so I have removed it. 🙏
[🔥 remove unused CommonLogEvent type from GTM types](https://github.com/liam-hq/liam/pull/305/commits/7bef593b384e1b244e76b7789c62f7901ba7d395)

MH4GF: Oh! OK, Please continue after this!

PR #307, opened 2024-12-18
refactor: Integrated `isRelated` into `isHighlighted`

MH4GF:
The uses were the same, so I will unify them.


FunamaYukina (approved):
Oh! Thank you for the refactoring🙏.
Looks good to me✨. I'll update the following PR accordingly!
- https://github.com/liam-hq/liam/pull/293

PR #309, opened 2024-12-18
refactor: Refactoring and testing of highlights on active tables

MH4GF:
Cutting out handling of active tables into functions and adding tests.

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/type.ts
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/type.ts
@@ -3,6 +3,7 @@ import type { Node } from '@xyflow/react'
 
 export type Data = {
   table: Table
+  isActiveHighlighted: boolean
   isHighlighted: boolean
   highlightedHandles: string[]
   sourceColumnName: string | undefined
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/highlightNodesAndEdges.test.ts
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/highlightNodesAndEdges.test.ts
@@ -27,11 +27,15 @@ const aTableNode = (
 const anEdge = (
   source: string,
   target: string,
+  sourceHandle: string | null,
+  targetHandle: string | null,
   override?: Partial<Edge>,
 ): Edge => ({
   id: `${source}-${target}`,
   source,
+  sourceHandle,
   target,
+  targetHandle,
   animated: false,
   data: { isHighlighted: false, ...override?.data },
   ...override,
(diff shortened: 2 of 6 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/type.ts
```diff
@@ -3,6 +3,7 @@ import type { Node } from '@xyflow/react'
 
 export type Data = {
   table: Table
+  isActiveHighlighted: boolean
```
MH4GF, line 6: This property is not yet used.
I want to stop calling `useUserEditingStoreActive()` directly in TableNode, and refer to it from here.

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/highlightNodesAndEdges.test.ts
```diff
@@ -0,0 +1,116 @@
...
+      aTableNode('comments'),
+      aTableNode('comment_users', {
+        data: aTableData('comment_users', {
+          isHighlighted: false,
+          highlightedHandles: [],
+        }),
+      }),
+    ])
+  })
+})
```
MH4GF, line 116: I would like you to review the test cases more intensively than the code!

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/highlightNodesAndEdges.ts
```diff
@@ -0,0 +1,127 @@
...
+        edges,
+        node,
+      )
+      return highlightNode(node, highlightedHandles)
+    }
+
+    return unhighlightNode(node)
+  })
+
+  return { nodes: updatedNodes, edges: [] }
```
MH4GF, line 126: As yet it can only support active table situations and does not support edges highlighting.
I want to be able to call this `highlightNodesAndEdges()` on all events to get the correct state of the nodes and edges.

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/highlightNodesAndEdges.ts
```diff
@@ -0,0 +1,127 @@
...
+const isActivelyRelatedNode = (
+  activeTableName: string | undefined,
+  edgeMap: EdgeMap,
+  node: TableNodeType,
+): boolean => {
+  if (!activeTableName) {
+    return false
+  }
+
+  return edgeMap.get(activeTableName)?.includes(node.data.table.name) ?? false
```
MH4GF, line 24: By making a memo of `Map<SourceTableName, TargetTableName[]>` once, you can then determine whether they are related with a small amount of looping.
This should also work for the logic during hover. This will be handled later.

junkisai (approved):
I thought it was good test code!

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/highlightNodesAndEdges.test.ts
```diff
@@ -0,0 +1,116 @@
...
+  source,
+  sourceHandle,
+  target,
+  targetHandle,
+  animated: false,
+  data: { isHighlighted: false, ...override?.data },
+  ...override,
+})
+
+describe(highlightNodesAndEdges, () => {
```
hoshinotsuyoshi, line 44: 👍

PR #311, opened 2024-12-18
PoC: Refactor edge cardinality by using custom mark

sasamuku:
## Summary

This PR will not be merged.
There are so many changes, so a part of the unnecessary implementation has been removed.
Check this PR and if the way is good, re-create the PR.

https://github.com/user-attachments/assets/0ae0d279-2896-47ca-afb2-dd582707a3ce

### Edge Markers

This PR's key concept is using `Edge Markers`.
Displaying / Hiding of cardinalities are always linked to edge.

https://reactflow.dev/examples/edges/markers


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
```diff
@@ -294,6 +297,30 @@ export const ERDContentInner: FC<Props> = ({
...
+        color="var(--node-layout)"
+      />
+      <CardinalityZeroOrManyLeftMarker
+        id="zeroOrManyLeft"
+        color="var(--pane-border-hover)"
+      />
+      <CardinalityZeroOrManyLeftMarker
+        id="zeroOrManyLeftHighlight"
+        color="var(--node-layout)"
+      />
```
sasamuku, line 323: Markers needs to be defined here.
ref: https://reactflow.dev/examples/edges/markers

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/RelationshipEdge/RelationshipEdge.tsx
```diff
@@ -41,6 +31,24 @@ export const RelationshipEdge: FC<Props> = ({
       <BaseEdge
         id={id}
         path={edgePath}
+        markerStart={
+          data?.cardinality
+            ? data?.isHighlighted
+              ? 'url(#zeroOrOneRightHighlight)'
+              : 'url(#zeroOrOneRight)'
```
sasamuku, line 38: className is not applied to markers because these are defined with outer component, I think.
That's why I define markers for non-highlighted or highlighted, but it is redundant 🥲

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
MH4GF: The top area seems to be hidden. Please check this!

スクリーンショット 2024-12-19 9 55 34

MH4GF (commented):
Very great!!

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
sasamuku: I did not realize that, thank you.
I've moved the markers into ReactFlow component.
73ddc62dca00cfbb9879079c8b3114e423a099b7

image

MH4GF:
@sasamuku The direction looked good! 🚀 
I haven't noted any minor adjustments, but are you going to rebuild the PR? I'll point it out there, is that OK?

sasamuku:
@MH4GF 
Thank you for review and I understood that the direction is not bad 👍

I'll rebuild PR so that I can delete and adjust some implementation on cardinalities.

PR #312, opened 2024-12-19
refactor: Refactoring of highlight edges on active table

MH4GF:
- continued: https://github.com/liam-hq/liam/pull/309

Support for edges with `highlightNodesAndEdges()`!

PR #313, opened 2024-12-19
fix: Fixed an issue where opening an active table from a query parameter would not highlight it

MH4GF:
Fixed an issue where opening an active table from a query parameter would not highlight it 👍🏻 

## Blocking

~Merge after the following PR merged.~

- https://github.com/liam-hq/liam/pull/312

## Testing

- Click on table node → Target node becomes active
- Click on table name in LeftPane → target node becomes active
- URL contains `?active=${tableName}` → target node becomes active

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
@@ -16,9 +16,9 @@ import { ERDContentProvider, useERDContentContext } from './ERDContentContext'
 import { RelationshipEdge } from './RelationshipEdge'
 import { Spinner } from './Spinner'
 import { TableNode } from './TableNode'
-import { highlightNodesAndEdges } from './highlightNodesAndEdges'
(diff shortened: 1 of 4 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
```diff
@@ -95,41 +95,25 @@ export const ERDContentInner: FC<Props> = ({
...
-      updateActiveTableName(nodeId)
-
-      const { nodes: updatedNodes, edges: updatedEdges } =
-        highlightNodesAndEdges(nodes, edges, nodeId)
-
-      setEdges(updatedEdges)
-      setNodes(updatedNodes)
-    },
-    [edges, nodes, setNodes, setEdges],
-  )
```
MH4GF, line 118: The reason was that Active highlighting was only done on click.
Monitoring active state and implementing it with useEffect, so that it is highlighted in all patterns.

FunamaYukina (approved):
The code is much cleaner and easier to read now✨. Thank you!

PR #292, opened 2024-12-17
Enable clicking while Table Detail opened

sasamuku:
## Summary

Enable clicking AppBar (mainly `Copy Link`) while Table Detail opened.

https://github.com/user-attachments/assets/2cb0a089-1645-4f43-b42c-9d8468168716

## Other Information

There is an Issue that option `modal={false} leaves Drawer open with `pointer-events: none`.
This PR includes the workaround.

* https://github.com/emilkowalski/vaul/issues/492

PR #316, opened 2024-12-19
Ensure `related tables` reload when any table node becomes active

hoshinotsuyoshi:
## Summary

For now, it appears that setting `modal={false}` in https://github.com/liam-hq/liam/pull/292 suppresses the re-rendering of the `RelatedTables` component. Therefore, I explicitly specified the `key` property.

https://github.com/liam-hq/liam/pull/316/commits/690b1594bb2e7a45150dca7eb4a1747c9b9975c2

https://github.com/user-attachments/assets/888f476b-9db8-490f-af65-d3d4bf4ed7e5


MH4GF (approved):
I see. LGTM 👍🏻

PR #317, opened 2024-12-19
refactor: integrate highlightNodesAndEdges function for improved node and edge highlighting on hover

MH4GF:
- ref: https://github.com/liam-hq/liam/pull/313
- ref: https://github.com/liam-hq/liam/pull/312
- ref: https://github.com/liam-hq/liam/pull/309

Highlighting on hover is now also supported with `highlightNodesAndEdges()`.

## Blocking
~Merge after the following PR merged.~

- https://github.com/liam-hq/liam/pull/313

## Testing

- Click on table node → Target node becomes active
- Click on table name in LeftPane → target node becomes active
- URL contains ?active=${tableName} → target node becomes active
- Hover on table node → Target node becomes hovered state

PR #321, opened 2024-12-19
fix: Fixed failure to highlight parent table

MH4GF:
- https://github.com/liam-hq/liam/pull/317

I noticed a problem during QA and have fixed it. 🙏🏻

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/highlightNodesAndEdges.ts
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/highlightNodesAndEdges.ts
@@ -1,9 +1,9 @@
 import type { Edge, Node } from '@xyflow/react'
 import { type TableNodeType, isTableNode } from './TableNode'
 
-type SourceTableName = string
 type TargetTableName = string
-type EdgeMap = Map<SourceTableName, TargetTableName[]>
+type RelatedTableName = string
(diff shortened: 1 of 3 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/highlightNodesAndEdges.ts
```diff
@@ -127,9 +127,13 @@ export const highlightNodesAndEdges = (
...
     if (!edgeMap.has(sourceTableName)) {
-      edgeMap.set(sourceTableName, [])
+      edgeMap.set(sourceTableName, new Set())
     }
-    edgeMap.get(sourceTableName)?.push(targetTableName)
+    if (!edgeMap.has(targetTableName)) {
+      edgeMap.set(targetTableName, new Set())
+    }
+    edgeMap.get(sourceTableName)?.add(targetTableName)
+    edgeMap.get(targetTableName)?.add(sourceTableName)
```
MH4GF, line 136: EdgeMap has so far only been able to cache information in the direction of the child from the parent.
Therefore, it was possible to highlight a child node when hovering a parent node, but not the reverse.

Added reverse cache and also set to Set to remove duplicates due to possible duplication.

junkisai (approved):
Using Sets seemed like a very good idea to me!

PR #322, opened 2024-12-19
✨ Grouping TableNodes without Relationships into a GroupNode and Applying a New Layout Definition

junkisai:
## Summary

I have grouped TableNodes without Relationships into a GroupNode, allowing a new layout definition to be applied. 
The new layout definition arranges the Nodes in a single vertical column.

| Before | After |
|--------|--------|
| スクリーンショット 2024-12-19 17 36 17 | スクリーンショット 2024-12-19 17 35 54 |

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/useInitialAutoLayout.ts
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/useInitialAutoLayout.ts
@@ -1,8 +1,8 @@
 import type { QueryParam } from '@/schemas/queryParam'
 import { addHiddenNodeIds, updateActiveTableName } from '@/stores'
 import { decompressFromEncodedURIComponent } from '@/utils'
-import { useReactFlow } from '@xyflow/react'
-import { useEffect } from 'react'
+import type { Node } from '@xyflow/react'
+import { useEffect, useMemo } from 'react'
 import { useERDContentContext } from './ERDContentContext'
 import { useAutoLayout } from './useAutoLayout'
 
@@ -25,9 +25,14 @@ const getHiddenNodeIdsFromUrl = async (): Promise<string[]> => {
   return hiddenNodeIds ? hiddenNodeIds.split(',') : []
 }
 
-export const useInitialAutoLayout = () => {
-  const { getNodes } = useReactFlow()
-  const nodes = getNodes()
+export const useInitialAutoLayout = (nodes: Node[]) => {
+  const tableNodesInitialized = useMemo(
+    () =>
+      nodes
+        .filter((node) => node.type === 'table')
+        .some((node) => node.measured),
+    [nodes],
+  )
 
   const {
     state: { initializeComplete },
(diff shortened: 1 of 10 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/useInitialAutoLayout.ts
```diff
@@ -25,14 +27,18 @@ export const useInitialAutoLayout = () => {
       return
     }
 
+    const tableNodesInitialized = nodes
+      .filter((node) => node.type === 'table')
+      .some((node) => node.measured)
+
```
junkisai, line 33: Before adding the Group Node, nodesInitialized would become true only when the drawing occurred after the conversion to TableNode was completed by `convertDBStructureToNodes()`.

However, with the addition of the Group Node, nodesInitialized started to become true even before the conversion to TableNode was completed. As a result, we created our own custom `nodesInitialized`.

hoshinotsuyoshi (commented):
Awesome work! 🎉 

Now, could you confirm if selecting `All Fields` hides all nodes? 👀

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/useInitialAutoLayout.ts
MH4GF: I think that makes sense!

MH4GF:
The `All Fields` display seems to be corrupted. Please check this problem? @junkisai 

スクリーンショット 2024-12-19 19 20 02

MH4GF (commented):
Implementation looked good!
This is great work! 😄

junkisai:
@MH4GF @hoshinotsuyoshi 
> The All Fields display seems to be corrupted. Please check this problem?

I didn't notice until you pointed it out...
I'll check on it!

MH4GF:
@junkisai 
I know it's a difficult problem, but thank you in advance 🙏🏻

junkisai:
@MH4GF @hoshinotsuyoshi 
I fixed it so that the layout works as expected even when toggling showMode!
Please review it again 🙏 

https://github.com/user-attachments/assets/5b7a92ea-ede3-4c2d-958e-ab19968ef982

MH4GF (approved):
Nice work!! Thank you so much!!!!

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/useInitialAutoLayout.ts
```diff
@@ -25,9 +25,14 @@ const getHiddenNodeIdsFromUrl = async (): Promise<string[]> => {
...
-  const { getNodes } = useReactFlow()
-  const nodes = getNodes()
+export const useInitialAutoLayout = (nodes: Node[]) => {
+  const tableNodesInitialized = useMemo(
+    () =>
+      nodes
+        .filter((node) => node.type === 'table')
+        .some((node) => node.measured),
+    [nodes],
+  )
```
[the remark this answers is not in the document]
junkisai: Yes.
It seems it was retrieving the value of the previous state before the desired value. 🤔

hoshinotsuyoshi:
> I have lightly tried to touch it in the preview environment, and the behavior seems fine 🙌🏻

Me too!

Could it be that nodes with similar rectangle heights are arranged close to each other? ☺️  In any case, awesome!

kukei2

Issue #335, opened 2024-12-20
Source cardinality is not highlighting


sasamuku:
### Self Checks

- [X] This is only for bug report, if you would like to ask a question, please head to [Discussions](../discussions/categories/q-a).
- [X] I have searched for existing issues [search for existing issues](../issues), including closed ones.

### Cloud or Self Hosted

Self Hosted

### Version (only for Self Hosted)

v0.0.11 + 1a8ca43 (2024-12-19)

### Steps to reproduce

1. Select a table
2. If primary key is related to multi foreign keys, it's not highlighting

<img width="644" alt="mastodon-erd" src="https://github.com/user-attachments/assets/230c0f25-349d-490b-950d-f0146033a4c3" />

### Expected Behavior

To be highlighted regardless of the number of foreign keys.

### Actual Behavior

Not highlighted when selected table's primary key has number of foreign keys.

### Additional Context

>SVG doesn't support z-index
ref: https://github.com/xyflow/xyflow/issues/849

`elevateEdgesOnSelect` might be used in this case.

>Enabling this option will raise the z-index of edges connected to a node when selected.
https://reactflow.dev/api-reference/react-flow#edges

_No response_

PR #338, opened 2024-12-20
🐛: Highlight source cardinality with multi-foreign keys

sasamuku:
## Summary

close: https://github.com/liam-hq/liam/issues/335

### before
mastodon-erd

### after

image

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/highlightNodesAndEdges.test.ts
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/highlightNodesAndEdges.test.ts
@@ -36,6 +36,7 @@ const anEdge = (
   target,
   targetHandle,
   animated: false,
+  zIndex: 0,
   data: { isHighlighted: false, ...override?.data },
   ...override,
 })
@@ -180,10 +181,12 @@ describe(highlightNodesAndEdges, () => {
       expect(updatedEdges).toEqual([
         anEdge('users', 'posts', 'users-id', 'posts-user_id', {
           animated: true,
+          zIndex: 1,
           data: { isHighlighted: true },
         }),
         anEdge('users', 'comment_users', 'users-id', 'comment_users-user_id', {
           animated: true,
+          zIndex: 1,
           data: { isHighlighted: true },
         }),
         anEdge(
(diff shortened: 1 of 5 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/convertDBStructureToNodes.ts
```diff
@@ -43,7 +43,7 @@ export const convertDBStructureToNodes = ({
         targetColumnCardinalities: tableColumnCardinalities.get(table.name),
       },
       position: { x: 0, y: 0 },
-      zIndex: 1,
+      zIndex: 2,
```
hoshinotsuyoshi, line 46: 💡 
Personally, I'm not very familiar with such frontend issues, but defining a file like `src/components/ERDRenderer/z-index.ts` or `src/components/z-index.ts` for cross-component definitions might prevent future omissions or oversights. It seems like a good idea. This approach could be suitable if you frequently call from `.ts` or `.tsx` files. 
However, it might be difficult to use from a `.css` file. What do you think? @liam-hq/liam-dev 

e.g.

```typescript
export const zIndex = {
  tableNode: 2,
  highlightEdge: 1,
  unhighlightEdge: 0,
}
```

```typescript
// when using
import { zIndex } from '../z-index'

// ...

zIndex: zIndex.tableNode
```

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/highlightNodesAndEdges.test.ts
```diff
@@ -261,6 +268,7 @@ describe(highlightNodesAndEdges, () => {
           'comment_users-comment_id',
           {
             animated: true,
+            zIndex: 1,
```
hoshinotsuyoshi, line 271: Testing is great!

hoshinotsuyoshi (approved):
I've made one comment, but it's about variable management. Overall, it seems to be fixed. LGTM!

frontend/packages/erd-core/src/components/ERDRenderer/convertDBStructureToNodes.ts
MH4GF: It would be better to define the z-index overlap as a constant 👍🏻 
There is a problem that it cannot be referenced in CSS, but as long as it is used as a property of React Flow, it seems to be fine to define it in TypeScript.

MH4GF (approved):
Awesome work!!

frontend/packages/erd-core/src/components/ERDRenderer/convertDBStructureToNodes.ts
sasamuku: That's true!
zIndex was a magic number and not good.

Is this general to define constants? see below:
1f10080cb927c839851717f0600f75d6b8e94ac7

frontend/packages/erd-core/src/components/ERDRenderer/constants.ts
hoshinotsuyoshi, line 1: I believe that naming the file `constants.ts` is ok. However, I associate `UPPER_CASE (SCREAMING_SNAKE_CASE)` with defining more immutable values. I feel like `camelCase` might be more suitable, but I don't have a strong opinion on this. I would really appreciate any opinions on this matter!  @MH4GF cc: @sasamuku  how do you think?

sasamuku: let me choose `camelCase` here!
c2ffe8dd87d6be3e67e3cdf33bd7fd5ab5ccc096

PR #341, opened 2024-12-20
fix: Render CardinalityMarkers based on loading state

sasamuku:
## Summary

Remove unexpected cardinalities on initial loading.

貼り付けた画像_2024_12_20_16_49

### before

https://github.com/user-attachments/assets/e8c46f9c-d8ad-41bf-a675-21a2ff47c7ff

### after

https://github.com/user-attachments/assets/3769c4bd-588a-4956-a140-711a6ce266e8

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
@@ -158,7 +158,7 @@ export const ERDContentInner: FC<Props> = ({
         selectionOnDrag
         deleteKeyCode={null} // Turn off because it does not want to be deleted
       >
-        <CardinalityMarkers />
+        {!loading && <CardinalityMarkers />}
         <Background
           color="var(--color-gray-600)"
           variant={BackgroundVariant.Dots}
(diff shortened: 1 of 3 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
```diff
@@ -158,7 +158,7 @@ export const ERDContentInner: FC<Props> = ({
         selectionOnDrag
         deleteKeyCode={null} // Turn off because it does not want to be deleted
       >
-        <CardinalityMarkers />
+        {!loading && <CardinalityMarkers />}
```
junkisai, line 161: [nits & imo]
Following the visibility control pattern used for TableNode loading, I think it would be better to apply the same approach to Edge components as well, as this would create more consistency in the implementation.

https://github.com/liam-hq/liam/blob/fix_cardinality_display_during_loading/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableNode.module.css#L11-L13

junkisai (approved):
LGTM 🙌

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
sasamuku: That's good idea 👍 Thanks
b63e2dad86cd5f8509b44cf612c3bde1050e7c71

PR #342, opened 2024-12-20
refactor: Update handleLayout to accept nodes and improve hidden node handling

MH4GF:
## Background

There is a problem with not highlighting when opening with a URL with query parameters.

## Description

In handleLayout, changed to take nodes as arguments.
This PR does not change the behavior. 
In the next PR, we plan to calculate the highlight information and pass it to nodes.


junkisai (approved):
It looked good!

PR #343, opened 2024-12-20
Set envName(`production` or `preview`) to cliVersion 

hoshinotsuyoshi:
## Summary

Integrates the current envName(`production` or `preview`)   into the React component's environment variables.

There is no change to the visual behavior of the ERD HTML.

## Testing

In local , make this diff

```diff
--- a/frontend/packages/erd-core/src/components/ERDRenderer/AppBar/HelpButton/ReleaseVersion.tsx
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/AppBar/HelpButton/ReleaseVersion.tsx
@@ -5,6 +5,8 @@ import styles from './ReleaseVersion.module.css'
 export const ReleaseVersion: FC = () => {
   const { cliVersion } = useCliVersion()

+  console.log('cliVersion:', cliVersion)
+
   // Example output for cliVersion:
   // - Released version:
   //   v0.0.11 (2024-12-19)
```

and `pnpm dev` , 

I found that `envName` set to `cliVersion`

スクリーンショット 2024-12-20 17 10 56

### DIFF
--- a/frontend/packages/cli/vite-plugins/set-env.ts
+++ b/frontend/packages/cli/vite-plugins/set-env.ts
@@ -7,6 +7,7 @@ import { type Plugin, loadEnv } from 'vite'
  * - VITE_CLI_VERSION_IS_RELEASED_GIT_HASH: A flag indicating whether the current GIT hash corresponds to a released tag.
  * - VITE_CLI_VERSION_GIT_HASH: The current GIT commit hash.
  * - VITE_CLI_VERSION_DATE: The commit date of the latest commit.
+ * - VITE_CLI_VERSION_ENV_NAME: Environment name (preview or production).
  *
  * These variables are essential for maintaining version consistency and tracking within the deployment environment.
  */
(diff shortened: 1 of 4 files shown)


frontend/packages/cli/vite-plugins/set-env.ts
```diff
@@ -56,12 +66,18 @@ export function setEnvPlugin(): Plugin {
 
       const packageJsonVersion = env.npm_package_version
       const gitHash = fetchGitHash()
+      const gitBranch = fetchGitBranch()
+
+      // The main branch is considered production, all other branches are treated as previews.
+      // This alignment is done to match the deployment settings to Vercel specified in .github/workflows.
+      const envName = gitBranch === 'main' ? 'production' : 'preview'
```
MH4GF, line 73: I'm not sure if I should write about it here, since it's information only our team uses.
But once it's done, it looks good!

MH4GF (approved):
LGTM 👍🏻

frontend/packages/cli/vite-plugins/set-env.ts
hoshinotsuyoshi: Indeed, there seems to be some merit to that opinion! :)   I believe managing the `envName` value within the CLI package is necessary for our team's dogfooding. Conversely, it might be a good idea to move this `//` comment to the `.github/workflows` side.

PR #344, opened 2024-12-20
fix: Fixed problem with TableNode not being highlighted when opened from URL with query parameter

MH4GF:
Highlighting was not handled during initial display, so it has been added.

https://github.com/user-attachments/assets/496c3af1-4c06-4d22-82bf-5daa7ddc41c0

## Blocking

~Merge after the following PR merged.~

- https://github.com/liam-hq/liam/pull/342


junkisai (approved):
LGTM!
The user experience seems to have improved for those who accessed via the shared link 🎉

PR #345, opened 2024-12-20
📈 add appEnv parameter to logging events for environment tracking.

FunamaYukina:
## Summary

Add ``appEnv`` parameter to log event functions and components.

ss 2575

## Testing

I've tested all clicks and other interactions.


MH4GF (approved):
LGTM 👍🏻

PR #346, opened 2024-12-20
fix: Fixed problem with active tables not being highlighted in LeftPane when opened via query parameter

MH4GF:
https://github.com/user-attachments/assets/95ec6652-e061-460b-81a3-439d0504ea44


frontend/packages/erd-core/src/components/ERDRenderer/LeftPane/TableNameMenuButton/TableNameMenuButton.tsx
```diff
@@ -13,9 +13,6 @@ type Props = {
 }
 
 export const TableNameMenuButton: FC<Props> = ({ node }) => {
-  const {
-    active: { tableName },
-  } = useUserEditingStore()
```
MH4GF, line 18: I've hit the bug a few times, and it seems this one may not re-render. 💭

sasamuku (approved):
LGTM👍👍

PR #350, opened 2024-12-20
refactor: Remove unused isRelatedToTable function and simplify TableNode component logic

MH4GF:
Each of the unnecessary logic was removed.


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/ERDContent.tsx
```diff
@@ -151,13 +133,7 @@ export const ERDContentInner: FC<Props> = ({
     <div className={styles.wrapper} data-loading={loading}>
       {loading && <Spinner className={styles.loading} />}
       <ReactFlow
-        nodes={nodes.map((node) => ({
-          ...node,
-          data: {
-            ...node.data,
-            onClick: handleNodeClick,
```
hoshinotsuyoshi, line 158: yes! L144 already has `handleNodeClick`

hoshinotsuyoshi (approved):
LGTM!

PR #356, opened 2024-12-23
⚒️ Adjust the appearance of the toolbar on mobile devices

FunamaYukina:
## Summary

I've resolved the problem of the missing toolbar on mobile browsers.

Preview: https://liam-erd-sample-acagd5sze-route-06-core.vercel.app/
## Changes

| device/browser | Before  | After |
| ------------- | ------------- | ------------- |
| iPhoneSE<br>(Safari) |  IMG_0083 | IMG_0082  |
| Android<br>(Chrome) | Screenshot_2024-12-23-15-12-26-15_40deb401b9ffe8e1df2f1cc5ba480b12 | Screenshot_2024-12-23-15-04-34-32_40deb401b9ffe8e1df2f1cc5ba480b12  |

 I've checked on Firefox🔥 and Brave🦁 as well, and there seems to be no issue.

### DIFF
--- a/frontend/.changeset/fair-mayflies-exist.md
+++ b/frontend/.changeset/fair-mayflies-exist.md
@@ -1,0 +1,6 @@
+---
+"@liam-hq/erd-core": patch
+"@liam-hq/cli": patch
+---
+
+⚒️ Adjust the appearance of the toolbar on mobile devices.
--- a/frontend/packages/ui/src/markers/CardinalityZeroOrManyLeftMarker.tsx
+++ b/frontend/packages/ui/src/markers/CardinalityZeroOrManyLeftMarker.tsx
@@ -11,7 +11,7 @@ export const CardinalityZeroOrManyLeftMarker: FC<MarkerProps> = ({
   ...props
 }) => {
   return (
-    <svg role="img" aria-label="Zero or Many Left Marker">
+    <svg width="0" height="0" role="img" aria-label="Zero or Many Left Marker">
       <defs>
         <marker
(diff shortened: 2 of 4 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDRenderer.module.css
```diff
@@ -18,7 +18,7 @@
 
 .main {
   width: 100%;
-  height: 100%;
+  height: calc(100svh - var(--default-header-height));
```
FunamaYukina, line 21: I tried using ``vh`` instead of ``svh``, but it didn't display correctly, so I went back to using ``svh``.

MH4GF (changes requested):
The cause of the height overflow is apparently the height of the transparent marker.

スクリーンショット 2024-12-23 16 24 20

https://github.com/liam-hq/liam/blob/c6838375e419b845e65c5e9623e63825c28b52b0/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/CardinalityMarkers.tsx

I'd like to adjust the cause of the overflow. Setting a fixed value for height as a last resort.

FunamaYukina:
@MH4GF 
Thanks for catching that🙏. I've removed the ``width`` and ``height`` attributes from the SVG and tested it on a real device.


ss 2585

MH4GF (approved):
I don't know why it can be displayed with width and height set to 0, but it seems to work fine when called from marker.
LGTM 👍🏻

PR #359, opened 2024-12-23
⚒️　Fixing the toolbar display bug

FunamaYukina:
## Summary

ss 2592

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDRenderer.module.css
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDRenderer.module.css
@@ -45,3 +45,15 @@
 .toolbarWrapper > * {
   pointer-events: auto;
 }
+@media screen and (max-width: 768px) {
+  .toolbarWrapper {
+    transition: opacity 0.3s, visibility 0.3s;
+    opacity: 1;
+    visibility: visible;
+    min-width: 354px;
+  }
+  :global([data-state='expanded']) + .main .toolbarWrapper {
+    opacity: 0;
+    visibility: hidden;
+  }
+}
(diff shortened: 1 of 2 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDRenderer.module.css
```diff
@@ -45,3 +45,7 @@
 .toolbarWrapper > * {
   pointer-events: auto;
 }
+
+:global([data-state='expanded']) + .main .toolbarWrapper {
+  display: none;
+}
```
kumanoayumi, line 51:
```suggestion
@media screen and (max-width: 768px) {
  :global([data-state='expanded']) + .main .toolbarWrapper {
    opacity: 0;
    visibility: hidden;
  }
}
```

kumanoayumi: For devices with widths other than the Mobile version, I would like to see the toolbar displayed, so I would also like the following added to media query of the toolbarWrapper.

```
@media screen and (max-width: 768px) {
  .toolbarWrapper {
    transition: opacity 0.3s, visibility 0.3s;
    opacity: 1;
    visibility: visible;
    min-width: 354px;
  }
}
```

Setting min-width reduces layout wobble when scaling down.

kumanoayumi (changes requested):
Please check the suggest comments! 😸

frontend/packages/erd-core/src/components/ERDRenderer/ERDRenderer.module.css
FunamaYukina: Thank you for your review and code🙏. I've implemented your suggestions, and the toolbar is now functioning much more smoothly✨. Thank you very much for your assistance.

fixed: https://github.com/liam-hq/liam/pull/359/commits/1c20fd1abe6b205c53a123c1b7487afd68e6532e

MH4GF (approved):
LGTM 👍🏻

kumanoayumi (approved):
thx! 🌟 LGTM 👍

PR #363, opened 2024-12-24
fix: Update css module for edge marker

sasamuku:
## Summary

**There was error when compiling on Next.js:**

```
Parsing css source code failed
Selector is not pure (pure selectors must contain at least one local class or id), (lightningcss, Selector(marker, specificity = 0x1))
```

**This error means:**

In CSS Modules, “selectors that do not contain local classes or IDs” will result in an error, so if you write an element selector like this marker { ... } as is, you will get the following error } as it is, the following error will occur.

**Solution:**

Edge markers have different IDs depending on whether they are highlighted or not. Changing the color is implemented by replacing the SVG element because the color of markers can not be dynamically changed by CSS. Therefore, this CSS was not actually applied. Currently it is difficult to make the transition work because the elements are different.

### DIFF
--- a/frontend/.changeset/warm-suits-love.md
+++ b/frontend/.changeset/warm-suits-love.md
@@ -3,4 +3,4 @@
 "@liam-hq/cli": patch
 ---
 
-fix: Update css module for edge marker
+:recycle: Update css module for edge marker
--- a/frontend/.changeset/warm-suits-love.md
+++ b/frontend/.changeset/warm-suits-love.md
@@ -1,0 +1,6 @@
+---
+"@liam-hq/erd-core": patch
+"@liam-hq/cli": patch
+---
+
+fix: Update css module for edge marker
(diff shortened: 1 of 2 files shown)


frontend/.changeset/warm-suits-love.md
```diff
@@ -0,0 +1,6 @@
+---
+"@liam-hq/erd-core": patch
+"@liam-hq/cli": patch
+---
+
+fix: Update css module for edge marker
```
MH4GF, line 6: I thought it was refactoring because there was no user impact. what do you think?

```suggestion
:recycle: Update css module for edge marker
```

MH4GF (approved):
LGTM 👍🏻

frontend/.changeset/warm-suits-love.md
sasamuku: Thanks for good comment!

PR #369, opened 2024-12-24
Set the assetPrefix to be accessible from Service Site

FunamaYukina:
## Summary

In order to allow the service site to access the ``frontend/apps/docs`` directory in this repository, I've configured an ``assetPrefix``.

### DIFF
--- a/frontend/apps/docs/next.config.mjs
+++ b/frontend/apps/docs/next.config.mjs
@@ -2,18 +2,10 @@ import { createMDX } from 'fumadocs-mdx/next'
 
 const withMDX = createMDX()
 
-function isProductionEnvironment() {
-  if (typeof window !== 'undefined') {
-    return window.location.hostname === 'liam-docs-sigma.vercel.app'
-  }
-  return false
-}
 /** @type {import('next').NextConfig} */
 const config = {
   reactStrictMode: true,
-  assetPrefix: isProductionEnvironment()
-    ? 'https://liam-docs-sigma.vercel.app'
-    : '',
+  assetPrefix: process.env.ASSET_PREFIX || '',
 }
 
 export default withMDX(config)
--- a/frontend/apps/docs/next.config.mjs
+++ b/frontend/apps/docs/next.config.mjs
@@ -2,9 +2,18 @@ import { createMDX } from 'fumadocs-mdx/next'
 
 const withMDX = createMDX()
 
+function isProductionEnvironment() {
+  if (typeof window !== 'undefined') {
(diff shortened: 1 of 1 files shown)


frontend/apps/docs/next.config.mjs
```diff
@@ -2,9 +2,18 @@ import { createMDX } from 'fumadocs-mdx/next'
 
 const withMDX = createMDX()
 
+function isProductionEnvironment() {
+  if (typeof window !== 'undefined') {
+    return window.location.hostname === 'liam-docs-sigma.vercel.app'
+  }
+  return false
+}
```
MH4GF, line 10: Can you please explain this to me as I did not understand why this process was necessary?

---

- https://github.com/vercel/next.js/discussions/20650#discussioncomment-4644834
    - In this example, it is always set to `'http://website2.com'`
- https://github.com/vercel/examples/blob/5d40d7600b64f5edb77fd48b4b04fc38e96a894c/solutions/microfrontends/apps/docs/next.config.js#L3
    - The assetPrefix is also a fixed value in the vercel example.

FunamaYukina: I apologize for the lack of explanation🙏. The reason is that setting it to a fixed value causes several path errors when **previewing.** 

ref: https://liam-docs-fk31aq36s-route-06-core.vercel.app/docs

(Additionally, I thought that we wouldn't be able to confirm the changes unless we deployed even during local development.)

So, I decided to only set the assetPrefix for production.

MH4GF: @FunamaYukina 
I see. In this case, it would be better to take it out of an environment variable such as `process.env.ASSET_PREFIX`.
By making it variable-dependent instead of environment-dependent, you can usually run the local environment on its own, but debug assetPrefixes in the local environment only when necessary.

ref: https://12factor.net/config

FunamaYukina: I see👍. Configuring ``ASSET_PREFIX`` exclusively for the production environment seems like a more streamlined approach. Thank you for your suggestion🙏. I will implement this change.

FunamaYukina: I've added the ``ASSET_PREFIX`` to Vercel's settings and updated the code accordingly!
https://github.com/liam-hq/liam/pull/369/commits/7cc97e7de8e0960f655363d3473a38d85906bbd1

MH4GF (approved):
LGTM 👍🏻

Issue #370, opened 2024-12-24
TableDetail table comment components dom error `Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>.`


hoshinotsuyoshi:
### Self Checks

- [X] This is only for bug report, if you would like to ask a question, please head to [Discussions](../discussions/categories/q-a).
- [X] I have searched for existing issues [search for existing issues](../issues), including closed ones.

### Cloud or Self Hosted

Self Hosted

### Version (only for Self Hosted)

v0.0.14 + f83c84d (2024-12-24)

### Steps to reproduce

```
pnpm dev
```

visit http://localhost:5173/?active=users on Chrome

open dev tools (`⌥⌘I`)

### Expected Behavior

no errors

### Actual Behavior

There are error messages

```
Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>.
```

<img width="1248" alt="スクリーンショット 2024-12-24 18 08 31" src="https://github.com/user-attachments/assets/d88c1b6e-3190-4dcc-a9af-4eaff8f16357" />

PR #371, opened 2024-12-24
:bug: fix: Comment component to use `<span>` instead of `<p>`

hoshinotsuyoshi:
## Summary

fix: Comment component to use `<span>` instead of `<p>`

## Related Issue

closes https://github.com/liam-hq/liam/issues/370

### before

`<p>` nesting exists.

スクリーンショット 2024-12-24 18 16 52

### after

スクリーンショット 2024-12-24 18 16 16


FunamaYukina (approved):
I confirmed that the console errors have been resolved in the local environment✅. Thank you!✨

ss 2593

PR #368, opened 2024-12-24
maintenance: move changesets dir to root

MH4GF:
- ref: https://github.com/liam-hq/liam/pull/351

The changesets directory move was omitted. 🙏🏻

PR #373, opened 2024-12-24
fix: fix the path of changeset files

MH4GF:
- ref: https://github.com/liam-hq/liam/pull/368

It appears that Changeset was unable to create a PR because it was merged into the old directory.


hoshinotsuyoshi (approved):
LGTM!! 🙏

PR #387, opened 2024-12-26
🐛 Fixed an issue where markers were not displayed in Safari after selecting TableNode

junkisai:
### [🚸 disable click events on RelationshipEdge](https://github.com/liam-hq/liam/pull/387/commits/30967b634df6105028544a09c8c90248f1861909)

In Reactflow's default settings, Edges were selectable.
Disabled Edge selection as this setting was causing highlighted Edges to become deselected.

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/1784c752-fc2f-481f-9282-cb562ecd8c4e" /> | <video src="https://github.com/user-attachments/assets/ff541495-f5ad-4fbb-a951-98a358efd439" /> | 

### [🐛 move CardinalityMarkers component in ERDRenderer](https://github.com/liam-hq/liam/pull/387/commits/0ea3d67fcad9aba4916880eceb4c5b59f16e5768)

The id specified in the `<marker>` of `<svg>` should be uniquely identifiable, but due to the ERDContent component being rendered in multiple places, this id was no longer unique.

This caused cardinality to stop rendering, so we fixed it by calling it in a place where the id can be uniquely assigned.

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/132a84f1-de20-4428-823a-12860c19e2e7" /> | <video src="https://github.com/user-attachments/assets/aed9fcbf-8620-4781-99ca-29d96c864329" /> |


MH4GF (approved):
LGTM 🚀 
Thanks for noticing!

frontend/packages/erd-core/src/components/ERDRenderer/ERDRenderer.tsx
```diff
@@ -45,6 +46,7 @@ export const ERDRenderer: FC<Props> = ({ defaultSidebarOpen = false }) => {
 
   return (
     <div className={styles.wrapper}>
+      <CardinalityMarkers />
```
sasamuku, line 49: 👍

PR #391, opened 2024-12-26
🐛 update TableHeader styles and add handle positioning for connections

junkisai:
## Summary

I fixed it so that cardinality is also displayed in Table Name mode.

| Before | After |
|--------|--------|
| image | image |


MH4GF (approved):
Nice 👍🏻 
I will do the release work after this.

PR #394, opened 2024-12-26
:wrench: Update changeset config to ignore specific packages

MH4GF:
## Background

Currently, even for applications that are not version controlled, versions have been updated.

image

ref: https://github.com/liam-hq/liam/pull/382

## description

Ignored packages that are not version controlled.

## Testing

You can check the `Version Packages` PR behavior at hand with `pnpm changeset version`.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -7,5 +7,9 @@
   "access": "restricted",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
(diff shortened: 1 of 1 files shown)


hoshinotsuyoshi (approved):
LGTM!

Confirmed `$ pnpm changeset version` does not effect to directory  `frontend/apps/erd-web` 👍

PR #395, opened 2024-12-26
Fix and Refactor cliVersion

sasamuku:
## Summary

* use discriminated union for cliVersion
* fix ReleaseVersion display for CLI
* disable GTM logging for ERD Web
  * In my understanding, GTM logging is just for internal testing
  * we need to discuss logging way to monitor user actions for production🤔

### DIFF
--- a/frontend/packages/cli/src/App.tsx
+++ b/frontend/packages/cli/src/App.tsx
@@ -32,7 +32,7 @@ const cliVersionData = {
   isReleasedGitHash:
     import.meta.env.VITE_CLI_VERSION_IS_RELEASED_GIT_HASH === '1',
   date: import.meta.env.VITE_CLI_VERSION_DATE,
-  displayedOn: 'web',
+  displayedOn: 'cli',
 }
 const cliVersion = v.parse(cliVersionSchema, cliVersionData)
 
--- a/frontend/packages/erd-core/src/components/ERDRenderer/AppBar/CopyLinkButton/CopyLinkButton.tsx
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/AppBar/CopyLinkButton/CopyLinkButton.tsx
@@ -31,12 +31,13 @@ export const CopyLinkButton: FC = () => {
         })
       })
 
-    clickLogEvent({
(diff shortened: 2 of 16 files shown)


frontend/packages/cli/src/App.tsx
```diff
@@ -32,7 +32,7 @@ const cliVersionData = {
   isReleasedGitHash:
     import.meta.env.VITE_CLI_VERSION_IS_RELEASED_GIT_HASH === '1',
   date: import.meta.env.VITE_CLI_VERSION_DATE,
-  displayedOn: 'web',
+  displayedOn: 'cli',
```
MH4GF, line 35: oh!

frontend/packages/erd-core/src/components/ERDRenderer/AppBar/CopyLinkButton/CopyLinkButton.tsx
```diff
@@ -31,12 +31,13 @@ export const CopyLinkButton: FC = () => {
         })
       })
 
-    clickLogEvent({
-      element: 'copyLinkButton',
-      cliVer: cliVersion.version,
-      appEnv: cliVersion.envName,
-    })
-  }, [toast, cliVersion.version, cliVersion.envName])
+    cliVersion.displayedOn === 'cli' &&
```
MH4GF, line 34: Eventually we would like to measure this in the web version, but that can be done later!

frontend/packages/erd-core/src/schemas/cliVersion/schemas.ts
```diff
@@ -1,10 +1,21 @@
...
   version: v.string(),
   gitHash: v.string(),
   envName: v.string(),
   isReleasedGitHash: v.boolean(),
   date: v.string(),
-  displayedOn: v.picklist(['cli', 'web']),
+  displayedOn: v.literal('cli'),
 })
+
+export const cliVersionSchema = v.union([
```
MH4GF, line 18: It would have been better to use `versionSchema`. (A separate PR is fine!)

MH4GF (approved):
LGTM 👍🏻

PR #400, opened 2024-12-27
Improve `erd build` option description

hoshinotsuyoshi:
This is without .changeset, I think that is ok.

## Summary

Improve `erd build` option description

## Testing

```
[liam]$ cd frontend/packages/cli
```

```
[cli]$ node dist-cli/bin/cli.js erd build --help
(node:61123) ExperimentalWarning: WASI is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
Usage: liam erd build [options]

Build ERD html assets

Options:
  --input <path|url>  Path or URL to the .sql file
  --format <format>   Format of the input file (postgres|schemarb)
  -h, --help          display help for command
```

### DIFF
--- a/frontend/packages/cli/src/cli/index.ts
+++ b/frontend/packages/cli/src/cli/index.ts
@@ -18,8 +18,8 @@ program.addCommand(erdCommand)
 erdCommand
   .command('build')
   .description('Build ERD html assets')
-  .option('--input <path>', 'Path to the .sql file')
-  .option('--format <format>', 'Format of the input file')
+  .option('--input <path|url>', 'Path or URL to the .sql file')
+  .option('--format <format>', 'Format of the input file (postgres|schemarb)')
   .action((options) => buildCommand(options.input, distDir, options.format))
 
 export { program }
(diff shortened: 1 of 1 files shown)


frontend/packages/cli/src/cli/index.ts
```diff
@@ -18,8 +18,8 @@ program.addCommand(erdCommand)
 erdCommand
   .command('build')
   .description('Build ERD html assets')
-  .option('--input <path>', 'Path to the .sql file')
-  .option('--format <format>', 'Format of the input file')
+  .option('--input <path|url>', 'Path or URL to the .sql file')
```
MH4GF, line 21: Can I change it like this so it is not limited to sql?

```suggestion
  .option('--input <path|url>', 'Path or URL to the schema file')
```

MH4GF (approved):
thanks 😄

frontend/packages/cli/src/cli/index.ts
hoshinotsuyoshi: oh, I see. nice!

hoshinotsuyoshi:
Now merging. Let me know if you notice anything else!

PR #402, opened 2024-12-27
:hammer: Adjust biome rule`lint/correctness/noNodejsModules` setting

### LINKED ISSUE #401: Adjust biome rule`lint/correctness/noNodejsModules` setting
## Task description

Adjust biome rule`lint/correctness/noNodejsModules` setting.

Disable it by default and enable it individually in each package as needed.

## Background

https://github.com/liam-hq/liam/pull/392/files#r1898284476

> Ignoring seems to be fine.
> 
> The reason I enabled this in `packages/configs/biome.jsonc` was that some packages are designed to run in non-Node.js environments.
> 
> However, given cases like this one, it seems better to disable it by default and enable it individually in each package as needed. I will make the adjustments accordingly.

hoshinotsuyoshi:
## Summary

:wrench: Adjust biome rule`lint/correctness/noNodejsModules` setting

## Related Issue

closes #401

### DIFF
--- a/frontend/packages/db-structure/biome.jsonc
+++ b/frontend/packages/db-structure/biome.jsonc
@@ -8,6 +8,7 @@
   "linter": {
     "rules": {
       "correctness": {
+        "noNodejsModules": "error",
         "useImportExtensions": {
           "level": "error",
           "fix": "safe",
--- a/frontend/apps/erd-web/next.config.ts
+++ b/frontend/apps/erd-web/next.config.ts
@@ -1,4 +1,3 @@
-// biome-ignore lint/correctness/noNodejsModules: This code is only run in the Node.js environment.
 import { execSync } from 'node:child_process'
 import type { NextConfig } from 'next'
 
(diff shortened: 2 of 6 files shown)


frontend/packages/db-structure/biome.jsonc
```diff
@@ -8,6 +8,7 @@
   "linter": {
     "rules": {
       "correctness": {
+        "noNodejsModules": "error",
```
hoshinotsuyoshi, line 11: The `frontend/packages/db-structure/src/schema/` directory (not the parser) likely still requires `noNodejsModules`, I suppose.

FunamaYukina (approved):
Thanks for your quick response 🙏. 
I understand that the ``noNodejsModules`` restriction applies in the following cases.

- frontend/packages/db-structure(not the parser)
- frontend/packages/erd-core

hoshinotsuyoshi:
Yes! The `parser` directory in `db-structure` contains many `ignore` comments. This issue remains unresolved 😅.

PR #403, opened 2024-12-27
✨ detect format

MH4GF:
Added the features to infer format identification and specify it in query parameters!

https://liam-erd-ctjuxynb7-route-06-core.vercel.app/erd/p/raw.githubusercontent.com/mastodon/mastodon/refs/heads/main/db/schema.rb

https://liam-erd-ctjuxynb7-route-06-core.vercel.app/erd/p/raw.githubusercontent.com/ekylibre/ekylibre/refs/heads/main/db/structure.sql

### DIFF
--- a/frontend/apps/docs/content/docs/url-based-access.mdx
+++ b/frontend/apps/docs/content/docs/url-based-access.mdx
@@ -3,3 +3,24 @@ title: URL-Based Access
 ---
 
 TBD
+
+## Detect schema format
+
+You can instantly generate an ER diagram by appending a GitHub file URL to `liambx.com/erd/p/`. For example:
+
+https://liambx.com/erd/p/github.com/mastodon/mastodon/blob/main/db/schema.rb
+
+When Liam ERD fetches your schema file, it automatically attempts to determine the schema format. Here’s how the detection works:
+
+- **File name check**: Certain filenames like `schema.rb` or `Schemafile` are assumed to be **`schemarb`**.
+- **File extension check**: Files ending with **`.rb`** are treated as **`schemarb`**, while files ending with **`.sql`** are treated as **`postgresql`**.
+- For more details, refer to the [detectFormat.ts](https://github.com/liam-hq/liam/blob/main/frontend/packages/db-structure/src/parser/supportedFormat/detectFormat.ts) file in our GitHub repository.
+
+### Override the Detected Format
+If the automatic detection doesn’t match your desired format, you can specify it manually via query parameters. For example:
+
+```
+https://liambx.com/erd/p/example?format=schemarb
+```
+
+Check [**/docs/supported-format**](/docs/supported-format) for a full list of valid format identifiers.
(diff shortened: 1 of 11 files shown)


frontend/apps/docs/content/docs/url-based-access.mdx
```diff
@@ -3,3 +3,24 @@ title: URL-Based Access
 ---
 
 TBD
+
+## Detect schema format
+
+You can instantly generate an ER diagram by appending a GitHub file URL to `liambx.com/erd/p/`. For example:
+
+https://liambx.com/erd/p/github.com/mastodon/mastodon/blob/main/db/schema.rb
```
sasamuku, line 11: This URL is not working now because is is not raw content.
It would be better to correspond to CLI specification? (`--input`)

frontend/apps/docs/content/docs/url-based-access.mdx
```diff
@@ -3,3 +3,24 @@ title: URL-Based Access
...
+
+When Liam ERD fetches your schema file, it automatically attempts to determine the schema format. Here’s how the detection works:
+
+- **File name check**: Certain filenames like `schema.rb` or `Schemafile` are assumed to be **`schemarb`**.
+- **File extension check**: Files ending with **`.rb`** are treated as **`schemarb`**, while files ending with **`.sql`** are treated as **`postgresql`**.
+- For more details, refer to the [detectFormat.ts](https://github.com/liam-hq/liam/blob/main/frontend/packages/db-structure/src/parser/supportedFormat/detectFormat.ts) file in our GitHub repository.
+
+### Override the Detected Format
+If the automatic detection doesn’t match your desired format, you can specify it manually via query parameters. For example:
```
sasamuku, line 20: 👍

frontend/apps/docs/content/docs/url-based-access.mdx
MH4GF: I want to keep it this way, because we want to provide an experience for users that they can see the ER diagram by adding a string in front of URL!

sasamuku: My comment was not well explained 🙏

* `https://github.com/mastodon/mastodon/blob/main/db/schema.rb` is not raw content.
* `https://raw.githubusercontent.com/mastodon/mastodon/refs/heads/main/db/schema.rb` is raw content.

So we need to implement conversion from `github.com` to `raw.githubusercontent.com`.
I think this conversion is good experience for our users.
I want to make sure there is no problem implementing the url conversion in the same way as the CLI.

sasamuku: I found the issue related the above comment 🙏

sasamuku (approved):
Thank you 👍

frontend/apps/docs/content/docs/url-based-access.mdx
MH4GF: Thanks for checking!
When sasamuku-san implemented the conversion process in the CLI before, we decided to remove it, but we ended up putting it back... 🙏🏻

PR #392, opened 2024-12-26
📘 Add robots.txt and sitemap to docs

FunamaYukina:
## Summary

Accessing https://liambx.com/docs routes the path to https://liam-docs-sigma.vercel.app/docs.

| repository | service-site | liam-hq/liam| 
| --- | --- | --- |
| URL |  https://liambx.com/ |  https://liam-docs-sigma.vercel.app| 
| Crawl |  ⭕️ |  ❌| 
| sitemap | has own sitemap<br>and refer to  https://liambx.com/docs/sitemap.xml | has own sitemap (/docs) |
		

This time, I created a sitemap for https://liam-docs-sigma.vercel.app/docs.

preview(sitemap): https://liam-docs-7tfe1sai4-route-06-core.vercel.app/docs/sitemap.xml

## Other Information

I don't think a ``changeset`` is needed for this change, but please let me know if I'm mistaken🙏.

PR #407, opened 2024-12-30
:wrench: to make accesible from service site

MH4GF:
- **:wrench: Add assetPrefix configuration to make it accessible from the service site**
    - ref: https://github.com/liam-hq/liam/pull/369
- **:wrench: Add robots.txt to disallow all web crawlers**
    - ref: https://github.com/liam-hq/liam/pull/392
- **:recycle: remove unnecessary .keep file**
- **:wrench: Add .env.production file with ASSET_PREFIX configuration**


FunamaYukina (approved):
LGTM👍✨

PR #411, opened 2025-01-03
:wrench: setup sentry for erd-web

MH4GF:
I set up sentry for erd-web.


frontend/apps/erd-web/.env.development
```diff
@@ -0,0 +1 @@
+NEXT_PUBLIC_ENV_NAME=development
```
MH4GF, line 1: The following is explained so that I've commit. Only secret is set from the Vercel dashboard.

> Good to know: .env, .env.development, and .env.production files should be included in your repository as they define defaults. 

ref: https://nextjs.org/docs/app/building-your-application/configuring/environment-variables#default-environment-variables

frontend/apps/erd-web/.env.production
```diff
@@ -0,0 +1 @@
+NEXT_PUBLIC_ENV_NAME=production
```
MH4GF, line 1: Since `.env.*` is determined from NODE_ENV, so `.env.preview` cannot be provided.
Therefore, preview settings are set from Vercel.

frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx
```diff
@@ -31,6 +32,7 @@ export default async function Page({
   const { value: dbStructure, errors } = await parse(input, 'schemarb')
   if (errors.length > 0) {
     for (const error of errors) {
+      Sentry.captureException(error)
```
Comment (MH4GF) [inline frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx:35]: Sent to Sentry when parsing fails. This is of great help for debugging.

frontend/apps/erd-web/next.config.ts
```diff
@@ -14,4 +15,42 @@ const nextConfig: NextConfig = {
   },
 }
 
-export default nextConfig
+export default withSentryConfig(nextConfig, {
+  // For all available options, see:
+  // https://github.com/getsentry/sentry-webpack-plugin#options
+
+  org: 'route06cojp',
+  project: 'liam-erd-web',
```
sasamuku, line 23: ask:
Should we embed company name, project name and dsn?
From the viewpoint of OSS, it would be better to be able to pass it as an environment variable 🤔

FunamaYukina (approved):
LGTM! Thank you for your introduction🚀.

frontend/apps/erd-web/next.config.ts
MH4GF: I was aware that these were not confidential information[^1], but looking at examples from other projects[^2], it certainly seemed safer to make them environment variables. I will correct it!

[^1]: https://sentry.zendesk.com/hc/en-us/articles/26741783759899-My-DSN-key-is-publicly-visible-is-this-a-security-vulnerability
[^2]: https://github.com/supabase/supabase/blob/29117959b14ab42ddb5730d0475e37c6fcf79b26/apps/studio/sentry.client.config.ts#L11

MH4GF: fix it: https://github.com/liam-hq/liam/pull/411/commits/ddd36f0a285464549e8c318c37af129013c8abc3
Since it is not confidential information, we will proceed without removing it from the git log.

sasamuku (approved):
LGTM 🚀
Thanks a lot!

PR #412, opened 2025-01-03
:sparkle: Improved error output

MH4GF:
- Output logs are now prefixed with `ERROR:` and `WARN:` and output in color.
- In case of `ERROR` exit 1.

## Testing

```sh
node dist-cli/bin/cli.js erd build --format postgres --input https://raw.githubusercontent.com/metabase/metabase/fe13ca685912aa3cde89a7a7e8c0d4698cf9d060/resources/migrations/initialization/metabase_mysql.sql
```

output
スクリーンショット 2025-01-03 19 54 00

### DIFF
--- a/frontend/packages/cli/src/cli/erdCommand/index.test.ts
+++ b/frontend/packages/cli/src/cli/erdCommand/index.test.ts
@@ -5,7 +5,7 @@ import { buildCommand } from './buildCommand/index.js'
 // Function to set up mocks
 function setupMocks() {
   vi.mock('./buildCommand', () => ({
-    buildCommand: vi.fn(),
+    buildCommand: vi.fn().mockImplementation(() => Promise.resolve([])),
   }))
 }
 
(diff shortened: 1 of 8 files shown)


frontend/packages/cli/src/cli/erdCommand/index.test.ts
```diff
@@ -5,7 +5,7 @@ import { buildCommand } from './buildCommand/index.js'
 // Function to set up mocks
 function setupMocks() {
   vi.mock('./buildCommand', () => ({
-    buildCommand: vi.fn(),
+    buildCommand: vi.fn().mockImplementation(() => Promise.resolve([])),
```
MH4GF, line 8: 📝 Now that the buildCommand function returns an array of errors, the mock also returns an array.

frontend/packages/cli/src/cli/erdCommand/runPreprocess.test.ts
```diff
@@ -47,7 +49,7 @@ describe('runPreprocess', () => {
     },
   )
 
-  it('should throw an error if the format is invalid', async () => {
+  it('should return an error if the format is invalid', async () => {
```
hoshinotsuyoshi, line 52: 👍 👍

hoshinotsuyoshi (approved):
LGTM!

PR #414, opened 2025-01-06
✨ feat: Enable browser back and forward for active table

sasamuku:
## Summary

Enable browser back and forward for active table:

https://github.com/user-attachments/assets/23b34648-37c5-478a-88c7-f56366129a48

### DIFF
--- a/frontend/packages/erd-core/src/stores/userEditing/store.ts
+++ b/frontend/packages/erd-core/src/stores/userEditing/store.ts
@@ -10,6 +10,7 @@ type UserEditingStore = {
   }
   showMode: ShowMode
   hiddenNodeIds: Set<string>
+  isPopstateNavigation: boolean
 }
 
 export const userEditingStore = proxy<UserEditingStore>({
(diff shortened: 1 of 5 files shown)


frontend/packages/erd-core/src/stores/userEditing/store.ts
```diff
@@ -31,7 +34,9 @@ subscribe(userEditingStore.active, () => {
     url.searchParams.delete(activeQueryParam)
   }
 
-  window.history.pushState({}, '', url)
+  if (!isPopstateNavigation) {
+    window.history.pushState({}, '', url)
+  }
```
sasamuku, line 39: 📝 History is not pushed when active table is changed by `popstate` event.
To prevent conflicts between the “go back in history” operation triggered by the `popstate` event and the “advance history” operation triggered by `pushState` on store update.

FunamaYukina (approved):
That's amazing✨! I've confirmed it works correctly on both Chrome and Safari. LGTM,  although I would appreciate additional feedback from other team members.

sasamuku:
If you have any idea on this PR, pls let me know.
This PR is going to be merged because the next PR is blocked.

PR #415, opened 2025-01-06
🐛 Fix unnecessary whitespace generation

FunamaYukina:
## Summary

I've removed the ``min-width`` constraint on the toolbarWrapper to eliminate the extraneous margin. As a result, the toolbarWrapper animation has been slightly modified.

| Before  | After |
| <video src="https://github.com/user-attachments/assets/a855275e-e38e-4a22-8c97-072bbcf60050"> | <video src="https://github.com/user-attachments/assets/3ab03f8f-2e2c-4ed6-8a1c-8c0700351e6f">  |

### DIFF
--- a/.changeset/clever-cherries-tan.md
+++ b/.changeset/clever-cherries-tan.md
@@ -1,0 +1,6 @@
+---
+"@liam-hq/erd-core": patch
+"@liam-hq/cli": patch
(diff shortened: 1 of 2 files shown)


sasamuku (approved):
LGTM!

PR #416, opened 2025-01-06
✨ feat: Add conversion from github.com to raw.githubusercontent.com

sasamuku:
## Summary

Add conversion from `github.com` to `raw.githubusercontent.com`.

Try:
https://liam-erd-kkbljftwm-route-06-core.vercel.app/erd/p/github.com/ekylibre/ekylibre/blob/main/db/structure.sql

### DIFF
--- a/frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx
+++ b/frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx
@@ -20,16 +20,20 @@ const searchParamsSchema = v.object({
   format: v.optional(supportedFormatSchema),
 })
 
-const getContentUrl = (url: string) => {
-  const parsedUrl = new URL(url)
+const resolveContentUrl = (url: string): string | undefined => {
+  try {
+    const parsedUrl = new URL(url)
 
-  if (parsedUrl.hostname === 'github.com' && url.includes('/blob/')) {
+    if (parsedUrl.hostname === 'github.com' && url.includes('/blob/')) {
+      return url
+        .replace('github.com', 'raw.githubusercontent.com')
+        .replace('/blob', '')
+    }
+
     return url
-      .replace('github.com', 'raw.githubusercontent.com')
-      .replace('/blob', '')
+  } catch {
+    return undefined
   }
-
-  return url
 }
 
 export default async function Page({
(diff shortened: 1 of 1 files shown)


frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx
```diff
@@ -20,6 +20,18 @@ const searchParamsSchema = v.object({
   format: v.optional(supportedFormatSchema),
 })
 
+const getContentUrl = (url: string) => {
+  const parsedUrl = new URL(url)
+
+  if (parsedUrl.hostname === 'github.com' && url.includes('/blob/')) {
```
Comment (sasamuku) [inline frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx:26]: To share url replacement logic with `@liam-hq/cli`, we have to add more dependencies between cli and erd-web.
How should we share them without dependencies?🤔

frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx
```diff
@@ -20,6 +20,18 @@ const searchParamsSchema = v.object({
   format: v.optional(supportedFormatSchema),
 })
 
+const getContentUrl = (url: string) => {
+  const parsedUrl = new URL(url)
```
Comment (MH4GF) [inline frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx:24]: Suggest: If a non-url string is passed(ex: liambx.com/erd/p/foobar), a 5xx error will occur here, so it is better to do some error handling.

frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx
Comment (MH4GF) [reply frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx:26]: It's distressing 💭 
It's similar to having a module split like packwerk, so it needs to be dependencies.
Since it is difficult to find the boundaries well, I thought it might be a good idea to write the same logic for both once.

MH4GF (approved):
Thanks!!

frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx
Comment (sasamuku) [reply frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx]: Thanks for your suggestion!

Added display of `notFound` when some error occurs in the input URL.

4165566fd5896d567793b5b8824a43e2cca2d4e3: for parse error like `/erd/p/ hoge/`
f1551ba83155191f3f6a66d942ae1dd8fe2902a1: for fetch error like `/erd/p/123/` (this can be parsed as `https://0.0.0.123/` 😅)

Comment (MH4GF) [reply frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx]: thanks 😄

MH4GF (approved):
LGTM 🚀

PR #427, opened 2025-01-09
💄 Add Title, Favicon, OGP for ERD Web

sasamuku:
## Summary

Add title and favicon for ERD Web:

|pattern|image|
|--|--|
|og:title exists|image|
|no og:title, title exists|image (confirmed with test html)|
|no title, no og:title|image|

For some reason, when I put favicon to `/app`, I faced the following issue:

* https://github.com/vercel/next.js/issues/71609
* workaround: put on `/public`


frontend/apps/erd-web/app/layout.tsx
```diff
@@ -14,8 +14,8 @@ const geistMono = Geist_Mono({
 })
 
 export const metadata: Metadata = {
-  title: 'Create Next App',
-  description: 'Generated by create next app',
+  title: 'Liam ERD',
+  description: 'Automatically generates beautiful and easy-to-read ER diagrams from your schema URL.',
```
sasamuku, line 18: 📝 ref: repository description

frontend/apps/erd-web/public/favicon.ico
hoshinotsuyoshi, line 1: 💭  If we place a favicon at the `/` of `erd-web` and host it at `https://liambx.com/erd/...`, I'm not entirely sure how users would access this `favicon.ico` when browsing the site. Any idea?

But In PR preview environment,  it might be useful.

frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx
```diff
@@ -37,6 +38,44 @@ const resolveContentUrl = (url: string): string | undefined => {
...
+
+  const res = await fetch(projectUrl).catch(() => null)
+
+  const projectName = await (async () => {
+    if (res?.ok) {
+      const html = await res.text()
+      const ogTitleMatch = html.match(
+        /<meta property="og:title" content="([^"]+)" \/>/,
+      )
+      const htmlTitleMatch = html.match(/<title>([^<]+)<\/title>/)
```
Comment (hoshinotsuyoshi) [inline frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx:60]: nits

```javascript
const ogTitleMatch = html.match(
  /<meta property="og:title" content="([^"]+)" \/>/,
)
const htmlTitleMatch = html.match(/<title>([^<]+)<\/title>/)
```

might work in most cases, but I’m slightly concerned about its heavy reliance on the specific structure of GitHub’s HTML strings.

For instance, it may fail to handle cases such as:
- When the `content` attribute appears before the `property` attribute (attribute order variation).
- When single quotes (`'`) are used instead of double quotes (`"`).
- When the closing `/>` lacks the slash (`>` instead of `/>`).

------

Parsing it as a DOM would be ideal, but if you want it to work in both Node.js and the browser, additional considerations might be needed. For now, I think keeping it as is would be fine.

frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx
```diff
@@ -37,6 +38,44 @@ const resolveContentUrl = (url: string): string | undefined => {
...
 }
 
+export async function generateMetadata({
+  params,
+}: PageProps): Promise<Metadata> {
+  const parsedParams = v.safeParse(paramsSchema, await params)
+  if (!parsedParams.success) return notFound()
+
+  const rawUrl = parsedParams.output.slug.join('/')
+  if (!rawUrl) notFound()
```
Comment (hoshinotsuyoshi) [inline frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx:48]: nits: L88 uses the type var `joinedPath`. It might be better to use the same expression here as well.

frontend/apps/erd-web/public/favicon.ico
hoshinotsuyoshi, line 1: ask:

When an end-user accesses an ERD under `https://liambx.com/erd/p/...`, I believe this favicon.ico won't actually be visible to the end-user. Is that understanding correct?

For example, consider the URL:  
`https://liambx.com/erd/p/raw.githubusercontent.com/mastodon/mastodon/refs/heads/main/db/schema.rb`.  

When accessing this page, the favicon is sourced from `https://liambx.com/favicon.ico`. My understanding is that this behavior will remain unchanged even after this PR is merged. Is that ok?

sasamuku: Yeah true.
There is no need to replace favicon, I think.
I will revert it.

sasamuku: You are totally right👍

frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx
Comment (sasamuku) [reply frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx:48]: I'll fix this 👍

frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx
Comment (sasamuku) [reply frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx:60]: It will be fine that this code works on Node.js not both.
I found `jsdom` that extract DOM and works on Node.js.
I will try to handle this in the other PR.

frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx
Comment (sasamuku) [reply frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx:48]: c991f352f6301adf49a1b50e8dc232ff2b0e0076

frontend/apps/erd-web/public/favicon.ico
sasamuku: reverted
bfdabd3b283b0cea1cb0fee51bf2555c06e81372

frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx
```diff
@@ -77,6 +114,13 @@ export default async function Page({
     notFound()
   }
 
+  const input = `
+  create_table "companies", id: :uuid, default: -> { "gen_random_uuid()" }, comment: "store companies", force: :cascade do |t|
+  t.string "name", null: false
+  t.string "address", null: false
+end
+  `
```
Comment (hoshinotsuyoshi) [inline frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx:122]: temp test?  @sasamuku

Comment (sasamuku) [reply frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx:122]: OMG 😇😇
I'll remove this!

Comment (sasamuku) [reply frontend/apps/erd-web/app/erd/p/[...slug]/page.tsx]: removed
07dd8c57feeb662fbb5fd895338b5ae9f876512c

hoshinotsuyoshi (approved):
LGTM! Nice work!

PR #432, opened 2025-01-09
💄 Move react flow attribution from bottom-right to bottom-left

sasamuku:
## Summary

Move react flow attribution from bottom-right to bottom-left.

|before|after|
|--|--|
|image|image|

Reason:
When embedding an ER diagram in an iframe in LP, we would like to display a link to go to the path of the ER diagram on bottom-right, but it overlaps with the React Flow logo (attribution).


MH4GF (approved):
Thanks 👍🏻

PR #438, opened 2025-01-10
💄 Style to ErrorDisplay component

hoshinotsuyoshi:
This PR focuses on styling. Please review it from that perspective 🙏 .

## Summary

💄 Style to ErrorDisplay component

## Testing

local: http://localhost:3001/erd/p/raw.githubusercontent.com/metabase/metabase/fe13ca685912aa3cde89a7a7e8c0d4698cf9d060/resources/migrations/initialization/metabase_mysql.sql

preview: https://liam-erd-41czdfe8k-route-06-core.vercel.app/erd/p/raw.githubusercontent.com/metabase/metabase/fe13ca685912aa3cde89a7a7e8c0d4698cf9d060/resources/migrations/initialization/metabase_mysql.sql

スクリーンショット 2025-01-10 9 42 01

## Other Information

No changeset added.

### DIFF
--- a/frontend/apps/docs/content/docs/parser/troubleshooting.mdx
+++ b/frontend/apps/docs/content/docs/parser/troubleshooting.mdx
@@ -1,0 +1,5 @@
+---
+title: Troubleshooting
+---
+
+TBD
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ErrorDisplay.module.css
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ErrorDisplay.module.css
@@ -118,6 +118,7 @@
   line-height: 160%;
 }
 
+/* TODO: This component has been partially adjusted for responsiveness, but confirm the designer's intentions and make any necessary corrections if needed. */
 @media screen and (max-width: 768px) {
   .message1 {
     gap: var(--spacing-1, 4px);
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ErrorDisplay.module.css
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ErrorDisplay.module.css
@@ -1,8 +1,121 @@
-/* TODO: styling */
 .wrapper {
-  position: relative;
-  width: 100%;
+  font-family: var(----main-font, Inter);
+  display: flex;
+  padding: 0px var(--spacing-6, 24px);
+  flex-direction: column;
+
   height: 100%;
+
+  justify-content: center;
+  align-items: center;
+  gap: var(--spacing-10, 40px);
+
   background-color: var(--global-background);
-  color: blanchedalmond;
+}
+
+.main {
+  display: flex;
+  padding: var(--spacing-8, 32px);
+  align-items: flex-start;
+  gap: var(--spacing-6, 24px);
+
+  border-left: 1px solid var(--callout-warning-border, rgba(255, 215, 72, 0.2));
+  background: var(--callout-warning-background, rgba(255, 215, 72, 0.1));
+}
+
+.iconWrapper {
+  display: flex;
+  width: 24px;
+  padding-top: var(--spacing-1, 4px);
+  align-items: center;
+}
+
+.inner {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: flex-start;
+  gap: var(--spacing-8, 32px);
+}
+
+.message1 {
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  gap: var(--spacing-2, 8px);
+}
+
+.message1Title {
+  color: var(--callout-warning-text, #c3b476);
+  font-family: var(--message-font, Montserrat);
+  font-size: var(--font-size-8, 20px);
+  font-weight: 700;
+  line-height: 160%;
+}
+
+.message1Sentence {
+  color: var(--callout-warning-text, #c3b476);
+  font-size: var(--font-size-5, 14px);
+  font-weight: 400;
+  line-height: 160%;
+}
+
+.message2 {
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  gap: var(--spacing-2, 8px);
+}
+
+.message2Title {
+  color: var(--callout-warning-text, #c3b476);
+  font-size: var(--font-size-7, 18px);
+  font-weight: 700;
+  line-height: 160%;
+}
+
+.message2Sentence {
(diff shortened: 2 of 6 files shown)


frontend/apps/docs/content/docs/parser/troubleshooting.mdx
hoshinotsuyoshi, line 1: `Parser > Troubleshooting` .

--------

スクリーンショット 2025-01-10 16 01 14

frontend/packages/erd-core/src/components/ERDRenderer/ErrorDisplay.module.css
```diff
@@ -1,8 +1,161 @@
-/* TODO: styling */
 .wrapper {
-  position: relative;
-  width: 100%;
+  font-family: var(----main-font, Inter);
```
FunamaYukina, line 2: Maybe this one? 🤔

```suggestion
  font-family: var(---main-font, Inter);
```
ss 2655

hoshinotsuyoshi: @FunamaYukina  sorry, `--` is maybe correct. I'll check now 🙇  🙇

hoshinotsuyoshi, line 2:
```suggestion
  font-family: var(--main-font, Inter);
```

hoshinotsuyoshi: ✔️  `$ git grep -- '---main-font'`  --> no match.

FunamaYukina: Oh, Sorry! My suggestion was incorrect.😵

hoshinotsuyoshi: Thank you for the catch!

frontend/packages/erd-core/src/components/ERDRenderer/ErrorDisplay.tsx
```diff
@@ -7,18 +8,70 @@ type Props = {
...
+              <p>
+                Move ahead with your project: You can still create it! The
+                unrecognized statements will just be skipped.
+              </p>
+            </div>
+            <a
+              href="https://liambx.com/docs/parser/troubleshooting"
+              target="_blank"
+              className={styles.callout}
+              rel="noreferrer"
```
MH4GF, line 49: I would like to remove it here because I would like to have the referrer's information.

ref: [別タブへのリンク記述「target=”_blank”とrel=”noopener noreferrer”」の見直しを - 株式会社真摯](https://cinci.jp/blog/20240723-review-target-blank-noopener-noreferrer-practices)

frontend/packages/erd-core/src/components/ERDRenderer/ErrorDisplay.module.css
```diff
@@ -1,8 +1,161 @@
...
+
+.message3Sentence {
+  color: var(--callout-warning-text, #c3b476);
+
+  font-size: var(--font-size-5, 14px);
+  font-weight: 400;
+  line-height: 160%;
+}
+
+@media screen and (max-width: 768px) {
```
MH4GF, line 121: Thank you for the responsive support!
However, the design for the mobile version is not yet complete and will be reimplemented later. But looks like it will stay that way for now.

MH4GF (approved):
Thanks!!

frontend/packages/erd-core/src/components/ERDRenderer/ErrorDisplay.tsx
hoshinotsuyoshi: @MH4GF  Thank you for pointing that!

> noreferrer

I’ve also been wondering whether it might be better to remove `noreferrer`. However, in the `erd-core` package, I think we should proceed with caution. Currently, the `ErrorDisplay.tsx` component should not affect the HTML output of the CLI, but depending on future developments, it might end up being used . In such cases, there’s a possibility that the URLs of environments hosted by end users could be exposed in the `referrer` header. I’d like to discuss this further.

frontend/packages/erd-core/src/components/ERDRenderer/ErrorDisplay.module.css
hoshinotsuyoshi: You're absolutely right! There’s a chance the web designer's intentions and design choices might not be fully captured, so I’ll add a comment to the CSS to reflect this.

frontend/packages/erd-core/src/components/ERDRenderer/ErrorDisplay.tsx
MH4GF: @hoshinotsuyoshi 

> In such cases, there’s a possibility that the URLs of environments hosted by end users could be exposed in the referrer header. I’d like to discuss this further.

This is true. We want to be cautious about this. Right now it looks like we're going to stay noreferrer. do you think?

hoshinotsuyoshi: note: AppBar has already noreferer.


```
[liam]$ git grep -l noreferrer
frontend/packages/erd-core/src/components/ERDRenderer/AppBar/AppBar.tsx
frontend/packages/erd-core/src/components/ERDRenderer/AppBar/GithubButton/GithubButton.tsx
frontend/packages/erd-core/src/components/ERDRenderer/AppBar/HelpButton/HelpButton.tsx
frontend/packages/erd-core/src/components/ERDRenderer/AppBar/ReleaseNoteButton/ReleaseNoteButton.tsx
frontend/packages/erd-core/src/components/ERDRenderer/ErrorDisplay.tsx
```

----


@MH4GF 

> Right now it looks like we're going to stay noreferrer.

Good. That sounds fine and reasonable for now, but I agree—it does feel like we might want to differentiate in the near future.

I haven’t thought about it in depth, but from the perspective of `erd-core`, would it make sense to use `noreferrer` when used from the CLI, and omit it when used in `erd-web`? The boundary for this might be around `ERDRenderer`. If we’re passing it as props, that’s where it could be handled, though using context might be a better approach.

hoshinotsuyoshi: For now, I think I’ll go ahead and merge this PR! :)

PR #435, opened 2025-01-09
🚸 Add ErrorDisplay component for handling and displaying errors in ERDViewer (first step)

hoshinotsuyoshi:
## Summary

Component for erd-web app.

- Introduced the `ErrorDisplay` component to provide a visual representation of errors in the ERD renderer.

## Testing

display: 

https://liam-erd-ajtacqg7o-route-06-core.vercel.app/erd/p/raw.githubusercontent.com/metabase/metabase/fe13ca685912aa3cde89a7a7e8c0d4698cf9d060/resources/migrations/initialization/metabase_mysql.sql

スクリーンショット 2025-01-09 20 14 13

PR #440, opened 2025-01-10
🐛 Fix argument type of errors

hoshinotsuyoshi:
## Summary

follow up https://github.com/liam-hq/liam/pull/435 .

`ERDRenderer` was also used by cli package. 🙏  https://github.com/liam-hq/liam/blob/f575795ac23844fccf5f89226e5336207cc32c23/frontend/packages/cli/src/App.tsx#L50 

## Testing

https://liam-erd-sample-fx2yr05sz-route-06-core.vercel.app/ shows contents.


MH4GF (approved):
LGTM 👍🏻

PR #443, opened 2025-01-10
publish erd core 2

MH4GF:
- **✨ Add Vite configuration and update package structure for erd-core**
- **🔧 Update build task dependencies in turbo.json to include 'gen'**
- **🔧 Update .syncpackrc to include 'module' and 'types' in package configuration**


MH4GF:
close because it is no longer necessary to publish。

PR #445, opened 2025-01-10
✨ Layout is now adjusted when tables are increased or decreased.

MH4GF:
Layout is now executed when hidden changes, such as when filtering from Related Tables.

https://github.com/user-attachments/assets/e9896864-cb36-4c2b-9164-561fdd6f0ab1


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/useSyncHiddenNodesChange.ts
hoshinotsuyoshi, line 1: In Safari, when reloading the page, the spinner seems to appear twice. 😭

- https://liam-erd-sample.vercel.app/ - works fine  
- https://liam-erd-sample-kzjrcxemu-route-06-core.vercel.app/ - the spinner shows twice.


https://github.com/user-attachments/assets/43893657-c819-49cd-af7a-b9b96ff5c535

MH4GF: Thanks, I see this flickering is happening in Chrome as well.
Since it cannot be resolved successfully, I will consider a different approach!

PR #451, opened 2025-01-14
✅ Add nullable test for Prisma

sasamuku:
## Summary

Add nullable test for Prisma.

### DIFF
--- a/frontend/packages/db-structure/src/parser/prisma/index.test.ts
+++ b/frontend/packages/db-structure/src/parser/prisma/index.test.ts
@@ -48,5 +48,26 @@ describe(processor, () => {
 
       expect(value).toEqual(expected)
     })
+
+    it('nullable', async () => {
+      const { value } = await processor(`
+        model users {
+          id   Int    @id @default(autoincrement())
+          description String?
+        }
+      `)
+
+      const expected = userTable({
+        columns: {
+          description: aColumn({
+            name: 'description',
+            type: 'String',
+            notNull: false,
+          }),
+        },
+      })
+
+      expect(value).toEqual(expected)
+    })
   })
 })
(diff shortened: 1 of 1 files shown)


frontend/packages/db-structure/src/parser/prisma/index.test.ts
```diff
@@ -7,8 +7,8 @@ describe(processor, () => {
   const userTable = (override?: Partial<Table>) =>
     aDBStructure({
       tables: {
-        User: aTable({
-          name: 'User',
+        users: aTable({
```
sasamuku, line 10: 📝 Prisma recommends PascalCase but we use snake_case to use common test cases for schema.rb, sql.

>A common approach for naming tables/collections in databases for example is to use plural form and [snake_case](https://en.wikipedia.org/wiki/Snake_case) notation. However, we recommended a different [naming convention (singular form, PascalCase)](https://www.prisma.io/docs/orm/reference/prisma-schema-reference#naming-conventions).

frontend/packages/db-structure/src/parser/prisma/index.test.ts
```diff
@@ -46,7 +46,26 @@ describe(processor, () => {
...
+          id   Int    @id @default(autoincrement())
+          description String?
+        }
+      `)
+
+      const expected = userTable({
+        columns: {
+          description: aColumn({
+            name: 'description',
+            type: 'String',
```
sasamuku, line 64: 📝 We need to convert column types to use the following common case. Might be dealt with in the next PR.

https://github.com/liam-hq/liam/blob/f769a6ce4777748ba5d6033f8ad22cf5f7d8714d/frontend/packages/db-structure/src/parser/__tests__/testcase.ts#L54

MH4GF (approved):
LGTM 👍🏻

Issue #337, opened 2024-12-20
[Feature Request] reflect show mode via query parameter


MH4GF:
### Discussed in https://github.com/liam-hq/liam/discussions/336

<div type='discussions-op-text'>

<sup>Originally posted by **MH4GF** December 20, 2024</sup>
Currently the default show mode is `Table name only`. However, there are cases where you may want to switch defaults.
Active and hidden tables can be reflect via query parameters. I would also like to reflect show mode from the query parameter.</div>

PR #452, opened 2025-01-14
✨ Implement show mode handling in URL parameters and state management

FunamaYukina:
## Summary

In thisPR, the code has been updated to manage the ``showMode`` state and synchronize it with the URL query parameters.

Preview: https://liam-erd-sample-ggnkr3clv-route-06-core.vercel.app/?showMode=TABLE_NAME&active=status_edits

## Related Issue

close: https://github.com/liam-hq/liam/issues/337

####  1. Managing showMode State:

The  ``showMode`` state is now monitored, and any changes to it are reflected in the URL query parameters.
A  ``previousShowMode`` variable is used to track the previous state and detect changes.

####  2. Synchronizing showMode with URL:

When  ``showMode`` changes, the URL query parameter  ``showMode`` is updated accordingly.
If showMode is set, it is added to the URL; otherwise, it is removed.

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/useInitialAutoLayout.ts
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/useInitialAutoLayout.ts
@@ -1,5 +1,10 @@
 import type { QueryParam } from '@/schemas/queryParam'
-import { addHiddenNodeIds, updateActiveTableName } from '@/stores'
+import type { ShowMode } from '@/schemas/showMode'
+import {
+  addHiddenNodeIds,
+  updateActiveTableName,
+  updateShowMode,
+} from '@/stores'
 import { decompressFromEncodedURIComponent } from '@/utils'
 import { type Node, useReactFlow } from '@xyflow/react'
 import { useEffect, useMemo } from 'react'
@@ -24,6 +29,14 @@ const getHiddenNodeIdsFromUrl = async (): Promise<string[]> => {
     : undefined
 
   return hiddenNodeIds ? hiddenNodeIds.split(',') : []
+}
+
+const getShowModeFromUrl = (): string | undefined => {
+  const urlParams = new URLSearchParams(window.location.search)
+  const showModeQueryParam: QueryParam = 'showMode'
+  const showMode = urlParams.get(showModeQueryParam)
+
+  return showMode || 'TABLE_NAME'
 }
 
 export const useInitialAutoLayout = (
@@ -66,6 +79,9 @@ export const useInitialAutoLayout = (
         shouldFitViewToActiveTable && activeTableName
           ? { maxZoom: 1, duration: 300, nodes: [{ id: activeTableName }] }
           : undefined
+
+      const showMode = getShowModeFromUrl()
+      updateShowMode(showMode as ShowMode)
 
       if (tableNodesInitialized) {
         handleLayout(updatedNodes, updatedEdges, fitViewOptions)
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/useInitialAutoLayout.ts
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/useInitialAutoLayout.ts
@@ -37,10 +37,6 @@ const getShowModeFromUrl = (): ShowMode => {
   const showModeQueryParam: QueryParam = 'showMode'
   const showMode = urlParams.get(showModeQueryParam)
 
-  if (!showMode) {
-    return 'TABLE_NAME'
-  }
-
   const result = v.safeParse(showModeSchema, showMode)
   if (result.success && result.output) {
     return result.output
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/useInitialAutoLayout.ts
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/useInitialAutoLayout.ts
@@ -1,5 +1,5 @@
 import type { QueryParam } from '@/schemas/queryParam'
-import type { ShowMode } from '@/schemas/showMode'
+import { type ShowMode, showModeSchema } from '@/schemas/showMode'
 import {
   addHiddenNodeIds,
   updateActiveTableName,
@@ -8,6 +8,7 @@ import {
 import { decompressFromEncodedURIComponent } from '@/utils'
 import { type Node, useReactFlow } from '@xyflow/react'
 import { useEffect, useMemo } from 'react'
+import * as v from 'valibot'
 import { useERDContentContext } from './ERDContentContext'
 import { highlightNodesAndEdges } from './highlightNodesAndEdges'
 import { useAutoLayout } from './useAutoLayout'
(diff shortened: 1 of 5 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/useInitialAutoLayout.ts
```diff
@@ -26,6 +31,14 @@ const getHiddenNodeIdsFromUrl = async (): Promise<string[]> => {
   return hiddenNodeIds ? hiddenNodeIds.split(',') : []
 }
 
+const getShowModeFromUrl = (): string | undefined => {
```
MH4GF, line 34: If it is done this way, the cast by as in line 84 should be eliminated:

```suggestion
const getShowModeFromUrl = (): ShowMode | undefined => {
```

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/usePopStateListener.ts
```diff
@@ -1,15 +1,23 @@
...
   useEffect(() => {
     const handlePopState = async () => {
       const url = new URL(window.location.href)
       const tableName = url.searchParams.get('active')
+      const showMode = url.searchParams.get('showMode')
 
       updateIsPopstateInProgress(true)
       updateActiveTableName(tableName ?? undefined)
 
+      updateShowMode(showMode as ShowMode)
```
MH4GF, line 19: The `as` here is dangerous. Because you can put any string in the query parameter.
It is a good to parse here with valibot, and if the value is invalid, put in the default value.

frontend/packages/erd-core/src/stores/userEditing/store.ts
```diff
@@ -52,3 +52,18 @@ subscribe(userEditingStore.hiddenNodeIds, async () => {
 
   window.history.pushState({}, '', url)
 })
+
+let previousShowMode = userEditingStore.showMode
```
MH4GF, line 56: Why should the previous value be retained?
I thought that since it is executed when there is a change in the value, it would never match the previous value.

MH4GF (commented):
Thanks 👍🏻 
I wrote some comments!

frontend/packages/erd-core/src/stores/userEditing/store.ts
FunamaYukina: In this case, since we're using the following code, we'll also subscribe to changes other than userEditingStore.showMode.💦
```
subscribe(userEditingStore, () => {
```

not
```
subscribe(userEditingStore.showcase, () => {
```
↑ I tried this approach as well, but I couldn't resolve the following error.
```
Argument of type 'string' is not assignable to parameter of type 'object'.
```

Therefore, by using ``previousShowMode``, we update the URL query parameter **only when userEditingStore.showMode changes**. This prevents unnecessary calls to ``window.history.pushState.``

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/useInitialAutoLayout.ts
FunamaYukina: Thank you for your feedback. You're absolutely right.
I've made the following adjustments, but I've also added some code to validate the received showMode. Do you think it's a bit redundant?

https://github.com/liam-hq/liam/pull/452/commits/0890fafc470fb8f7501fdb7e3a14d838592fc8b5


```ts
const getShowModeFromUrl = (): ShowMode => {
  const urlParams = new URLSearchParams(window.location.search)
  const showModeQueryParam: QueryParam = 'showMode'
  const showMode = urlParams.get(showModeQueryParam)

  if (!showMode) {
    return 'TABLE_NAME'
  }

  const result = v.safeParse(showModeSchema, showMode)
  if (result.success && result.output) {
    return result.output
  }

  return 'TABLE_NAME'
}

```

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/usePopStateListener.ts
FunamaYukina: Thank you for your feedback🙏. I've made the same adjustments to this one(https://github.com/liam-hq/liam/pull/452/commits/0890fafc470fb8f7501fdb7e3a14d838592fc8b5 ) as well.
https://github.com/liam-hq/liam/pull/452/commits/168010704d67685b61b7306768863717a0e3d072

frontend/packages/erd-core/src/stores/userEditing/store.ts
MH4GF: I see, for that purpose, I thought it would be a good to try valtio's subscribeKey first!
https://valtio.dev/docs/api/utils/subscribeKey

If this still does not work, the current code looks good!

FunamaYukina: Thank you! I'll give it a try🙏.

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/useInitialAutoLayout.ts
```diff
@@ -26,6 +32,23 @@ const getHiddenNodeIdsFromUrl = async (): Promise<string[]> => {
...
+const getShowModeFromUrl = (): ShowMode => {
+  const urlParams = new URLSearchParams(window.location.search)
+  const showModeQueryParam: QueryParam = 'showMode'
+  const showMode = urlParams.get(showModeQueryParam)
+
+  if (!showMode) {
+    return 'TABLE_NAME'
+  }
+
+  const result = v.safeParse(showModeSchema, showMode)
```
MH4GF, line 44: Since safeParse can contain any value, it would be simpler to do the following:

```suggestion
  const result = v.safeParse(showModeSchema, showMode)
```

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/usePopStateListener.ts
```diff
@@ -1,15 +1,36 @@
...
 
+      if (!showMode) {
+        updateShowMode('TABLE_NAME')
+        setTimeout(() => {
+          updateIsPopstateInProgress(false)
+        }, 0)
+        return
+      }
+
+      const result = v.safeParse(showModeSchema, showMode)
```
MH4GF, line 28: This one would be better too:

```suggestion
      const result = v.safeParse(showModeSchema, showMode)
```

frontend/packages/erd-core/src/stores/userEditing/store.ts
FunamaYukina: I was able to fix the code using the subscribeKey! Thank you.
https://github.com/liam-hq/liam/pull/452/commits/fc6a5ddc1f46be4b870c8b9702b132e6381dabae

FunamaYukina:
@MH4GF 
I apologize, but I've found another issue🙏. It might take some time to fix, so would it be okay if I created a separate PR for it?🙇‍♀️

The problem is that you can't go back more than once in your browser.
Specifically, if you:
1. Change the showMode in the ERD multiple times(ex. ``KEY_ONLY -> ALL FIELDS -> TABLE NAME``), and then
2. Try to go back more than once in your browser,
3. You can't go back any further.(we can see ``ALL FIELDS``,but we cant go back ``KEY_ONLY``)

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/usePopStateListener.ts
FunamaYukina: Thank you so much! fixed: https://github.com/liam-hq/liam/pull/452/commits/e99b44698436eb30771ba9abe295732711e8d710

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/useInitialAutoLayout.ts
FunamaYukina: Thank you so much🙏! fixed: https://github.com/liam-hq/liam/pull/452/commits/e99b44698436eb30771ba9abe295732711e8d710

MH4GF:
@FunamaYukina history operation is currently unstable, so a separate issue is fine, of course!

MH4GF (approved):
LGTM 👍🏻

FunamaYukina:
@MH4GF Got it.Thank you🙏! I'll address this in a new PR.

FunamaYukina:
Thank you for your review!

PR #453, opened 2025-01-15
Fix docs search

hoshinotsuyoshi:
## Summary  
  

The `apps/doc` app now serves as the origin for `https://liambx.com/docs`.  

Requests to `/docs/:path*` are directed to the origin `apps/doc` app. However, it seems that the RootProvider of fumadocs defaults the search API path to `/api/search` unless a specific search option is provided. To address this, we’ve adjusted the `apps/doc` app to handle search under the `/docs/` path. This change is intended to ensure that search functionality works seamlessly even when accessed via the service site.  

Although integration tests have not been conducted, the `apps/doc` app functions correctly on its own. Therefore, I’d like to proceed with testing this approach first.  

## Testing

✔️  I checked in preview env! : https://liam-docs-p5mz4kkex-route-06-core.vercel.app/docs


MH4GF (approved):
This implementation would work in a local environment. LGTM 👍🏻

PR #458, opened 2025-01-15
docs: Fix a path

hoshinotsuyoshi:
## Summary

I made two minor fixes:

* d6797474 : docs: Fix a path
* c8cb0163 : add a comment for a .ts file which is referred from docs


MH4GF (approved):
Oh...Thanks!!

PR #459, opened 2025-01-15
✨️  Add interactive init command with Inquirer and optional GitHub Actions workflow

hoshinotsuyoshi:
## Summary

1. **Interactive `init` Command**:  
   - Utilizes Inquirer.js to guide users interactively through the setup process.
   - Supports selection of database or ORM technologies like PostgreSQL, Ruby on Rails (`schema.rb`), and Prisma.
   - Provides an option to generate a GitHub Actions workflow for automating ERD builds.

2. **Dependency Updates**:  
   - Added `inquirer` (v12.3.2) as a new dependency for interactive prompts.
   - Updated the lockfile (`pnpm-lock.yaml`) to reflect new packages and transitive dependencies.


frontend/packages/cli/.gitignore
```diff
@@ -30,3 +30,4 @@ dist-ssr
 dist-cli
 public/schema.json
 .rollup.cache
+.github/workflows
```
MH4GF, line 33: Is this line necessary?

frontend/packages/cli/src/cli/initCommand/index.ts
```diff
@@ -1,20 +1,132 @@
...
+  ])
+
+  const selectedFormat = formatMap[answers.dbOrOrm] || 'postgresql'
+
+  console.info(`
+For more details about Liam ERD usage and advanced configurations, check out:
+${yocto.blueBright('https://liambx.com/docs')}
+`)
+
+  console.info('\n--- Output ---')
```
MH4GF, line 74: Looks like this line doesn't need to be there.

MH4GF (approved):
LGTM 🚀 
Small comment added!

frontend/packages/cli/.gitignore
hoshinotsuyoshi: No, that's right. I'll remove it!

frontend/packages/cli/src/cli/initCommand/index.ts
hoshinotsuyoshi: It does seem like this line doesn't need to be there. I agree it feels fine to remove it for now."

PR #454, opened 2025-01-15
✨ Add support for Prisma format in parser

sasamuku:
## Summary

Support for Prisma format in parser.

`@prisma/internals` is CommonJS, we had some problems with `cli` and `erd-web`.
This PR contains workarounds for those problems.

PR #464, opened 2025-01-15
🔧 Disable turbopack option from dev command for ERD Web

sasamuku:
## Summary

Removed `--turbopack` flag from the dev script as it may bypass webpack configurations defined in `next.config.ts`, specifically our webpack alias settings for `@prisma/internals`. This ensures consistent behavior between development and production environments.

>Note: Production builds with next build are not yet supported.
https://turbo.build/pack/docs

## Related Issue

- https://github.com/liam-hq/liam/pull/454

### DIFF
--- a/frontend/apps/erd-web/package.json
+++ b/frontend/apps/erd-web/package.json
@@ -21,7 +21,7 @@
   },
   "scripts": {
     "build": "next build",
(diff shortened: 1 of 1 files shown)


hoshinotsuyoshi (approved):
LGTM!

PR #469, opened 2025-01-16
✨ Enhance format detection by adding support for prisma

sasamuku:
## Summary

It is not affect production deployment and can detect Prisma as format (but can not dispay ERD yet).

### DIFF
--- a/frontend/packages/db-structure/src/parser/supportedFormat/detectFormat.ts
+++ b/frontend/packages/db-structure/src/parser/supportedFormat/detectFormat.ts
@@ -6,11 +6,13 @@ import type { SupportedFormat } from './index.js'
 const nameToFormatMap: Record<string, SupportedFormat> = {
   'schema.rb': 'schemarb',
   schemafile: 'schemarb',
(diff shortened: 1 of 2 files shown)


frontend/packages/db-structure/src/parser/supportedFormat/detectFormat.ts
MH4GF, line 1: Can you add a test case? I think we can easily prepare it for you.

https://github.com/liam-hq/liam/blob/f5ee4ea06fb4b900c4e6c972d09106a2551d4ecb/frontend/packages/db-structure/src/parser/supportedFormat/detectFormat.test.ts

MH4GF (approved):
Thanks!!

frontend/packages/db-structure/src/parser/supportedFormat/detectFormat.ts
sasamuku: Thanks✨
Added test:

aafbbe2f6a28bb587daaf6e9a68364209a5cb95c

PR #470, opened 2025-01-16
💄 Update LiamLogo component to use currentColor for fill attributes

FunamaYukina:
## Summary

The LiamLogo component has been updated to use the ``currentColor`` value for its fill property.

ss 2679


junkisai (approved):
LGTM!

PR #472, opened 2025-01-16
🐛 Update TableCounter to count only table nodes

FunamaYukina:
## Summary

I've updated TableCounter to only count nodes that are strictly tables. Previously, ``non-related-table-group`` nodes were being included, leading to incorrect counts.

ss 2676


hoshinotsuyoshi (approved):
LGTM!

PR #489, opened 2025-01-17
Add permissions to the GitHub Actions workflows

masutaka:
I have already changed the `GITHUB_TOKEN` permissions to read-only at the organizational level. This follows the [Principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege).

In this pull request, I will make changes to workflows that should clearly be granted permissions.

Since all the tests in this pull request have passed, I think it is ok. If permissions-related problems occur in the future, we can deal with them on a case-by-case basis.

### DIFF
--- a/.github/workflows/frontend-ci.yml
+++ b/.github/workflows/frontend-ci.yml
@@ -12,6 +12,9 @@ jobs:
   setup-job:
     runs-on: ubuntu-latest
     timeout-minutes: 3
+    permissions:
+      contents: read
+      pull-requests: read
     outputs:
       has-changes: ${{ steps.changes.outputs.has-changes }}
     steps:
--- a/.github/workflows/vercel-deploy-erd.yml
+++ b/.github/workflows/vercel-deploy-erd.yml
@@ -15,6 +15,9 @@ concurrency:
 jobs:
   setup-deployment:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
     outputs:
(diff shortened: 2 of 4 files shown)


.github/workflows/frontend-ci.yml
```diff
@@ -12,6 +12,9 @@ jobs:
   setup-job:
     runs-on: ubuntu-latest
     timeout-minutes: 3
+    permissions:
+      contents: read
+      pull-requests: read
```
masutaka, line 17: FYI:
The `dorny/paths-filter` action requires `pull-requests: read permission`.
https://github.com/dorny/paths-filter?tab=readme-ov-file#supported-workflows

.github/workflows/vercel-deploy-erd.yml
```diff
@@ -15,6 +15,9 @@ concurrency:
 jobs:
   setup-deployment:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
```
masutaka, line 20: ditto

.github/workflows/vercel-deploy.yml
```diff
@@ -15,6 +15,9 @@ concurrency:
 jobs:
   setup-deployment:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
```
masutaka, line 20: ditto

MH4GF (approved):
LGTM 👍🏻

PR #498, opened 2025-01-17
🔧 Update Next.js configuration to exclude '@prisma/internals'

sasamuku:
## Summary

Exclude `@prisma/internals` from the bundle
to prevent `Error: Cannot find module 'fs'` errors in the build process.

✅ schema.rb
https://liam-erd-2e02911f4-route-06-core.vercel.app/erd/p/github.com/mastodon/mastodon/blob/main/db/schema.rb

✅ sql
https://liam-erd-2e02911f4-route-06-core.vercel.app/erd/p/github.com/ekylibre/ekylibre/blob/main/db/structure.sql

✅ Prisma 🎉
https://liam-erd-2e02911f4-route-06-core.vercel.app/erd/p/github.com/langfuse/langfuse/blob/main/packages/shared/prisma/schema.prisma

### DIFF
--- a/frontend/apps/erd-web/next.config.ts
+++ b/frontend/apps/erd-web/next.config.ts
@@ -35,6 +35,17 @@ const nextConfig: NextConfig = {
       config.externals['@prisma/schema-files-loader'] =
         '@prisma/schema-files-loader'
     }
+    config.plugins.push({
+      // biome-ignore lint/suspicious/noExplicitAny: webpack types are incomplete so we need to use any here
+      apply: (compiler: any) => {
+        compiler.hooks.afterEmit.tap('InstallPrismaInternals', () => {
+          execSync('node scripts/install-prisma-internals.mjs', {
+            stdio: 'inherit',
+            cwd: __dirname,
+          })
+        })
+      },
+    })
 
     return config
   },
--- a/frontend/apps/erd-web/next.config.ts
+++ b/frontend/apps/erd-web/next.config.ts
@@ -6,13 +6,36 @@ const gitCommitHash = execSync('git rev-parse --short HEAD').toString().trim()
 const releaseDate = new Date().toISOString().split('T')[0]
 
 const nextConfig: NextConfig = {
-  // NOTE: Exclude '@prisma/internals' from the client-side bundle
-  // This module is server-side only and should not be included in the client build
+  // NOTE: Exclude Prisma-related packages from the bundle
+  // These packages are installed separately in the node_modules/@prisma directory
+  // Excluding them prevents `Error: Cannot find module 'fs'` errors in the build process
   webpack: (config) => {
-    config.resolve.alias = {
-      ...config.resolve.alias,
-      '@prisma/internals': false,
+    if (Array.isArray(config.externals)) {
+      config.externals.push(
+        '@prisma/debug',
+        '@prisma/engines',
+        '@prisma/engines-version',
+        '@prisma/fetch-engine',
+        '@prisma/generator-helper',
+        '@prisma/get-platform',
+        '@prisma/internals',
+        '@prisma/prisma-schema-wasm',
+        '@prisma/schema-files-loader',
+      )
+    } else {
+      config.externals['@prisma/debug'] = '@prisma/debug'
+      config.externals['@prisma/engines'] = '@prisma/engines'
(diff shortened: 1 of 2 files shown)


frontend/apps/erd-web/next.config.ts
```diff
@@ -6,13 +6,47 @@ const gitCommitHash = execSync('git rev-parse --short HEAD').toString().trim()
...
+      config.externals['@prisma/internals'] = '@prisma/internals'
+      config.externals['@prisma/prisma-schema-wasm'] =
+        '@prisma/prisma-schema-wasm'
+      config.externals['@prisma/schema-files-loader'] =
+        '@prisma/schema-files-loader'
     }
+    config.plugins.push({
+      // biome-ignore lint/suspicious/noExplicitAny: webpack types are incomplete so we need to use any here
+      apply: (compiler: any) => {
+        compiler.hooks.afterEmit.tap('InstallPrismaInternals', () => {
```
sasamuku, line 41: 📝 
Used `afterEmit` hooks because executing the script via `postbuild` does not work well.
https://webpack.js.org/api/compiler-hooks/#afteremit

>`postbuild` does not work well.

When executing the script via `postbuild`, modules can not be found with the following error.

```ts
Error: Cannot find module '@prisma/internals'
```

That's why I've integrated the script with build process by using hook.

frontend/apps/erd-web/next.config.ts
```diff
@@ -6,13 +6,47 @@ const gitCommitHash = execSync('git rev-parse --short HEAD').toString().trim()
...
-  // NOTE: Exclude '@prisma/internals' from the client-side bundle
-  // This module is server-side only and should not be included in the client build
+  // NOTE: Exclude Prisma-related packages from the bundle
+  // These packages are installed separately in the node_modules/@prisma directory
+  // Excluding them prevents `Error: Cannot find module 'fs'` errors in the build process
   webpack: (config) => {
-    config.resolve.alias = {
-      ...config.resolve.alias,
-      '@prisma/internals': false,
+    if (Array.isArray(config.externals)) {
```
sasamuku, line 13: 📝
Why not using `serverExternalPackages`?
https://nextjs.org/docs/app/api-reference/config/next-config-js/serverExternalPackages

The dependency on `@prisma/internals` is as follows, and it is likely that `serverExternalPackages` did not work because of the distant dependencies.

```
@liam-hq/erd-web -> @liam-hq/db-structure -> @prisma/internals
```

frontend/apps/erd-web/scripts/install-prisma-internals.mjs
```diff
@@ -0,0 +1,34 @@
...
+  if (!fs.existsSync(dotNextPath)) {
+    fs.mkdirSync(dotNextPath, { recursive: true })
+  }
+  if (!fs.existsSync(nodeModulesPath)) {
+    fs.mkdirSync(nodeModulesPath, { recursive: true })
+  }
+
+  process.chdir(nodeModulesPath)
+
+  execSync('npm install @prisma/internals', { stdio: 'inherit' })
```
MH4GF, line 22: Q: Couldn't you just run it through `pnpm`?

MH4GF (approved):
This is great job 🚀🚀🚀
Thanks!!!

frontend/apps/erd-web/scripts/install-prisma-internals.mjs
sasamuku: hmm, I've tried `pnpm` but the following error occurred.

```ts
 ERR_PNPM_INCLUDED_DEPS_CONFLICT  modules directory was installed with optionalDependencies, dependencies, devDependencies. Current install wants optionalDependencies, dependencies.
node:internal/errors:984
  const err = new Error(message);
              ^

Error: Command failed: pnpm install @prisma/internals
```

sasamuku: Could be similar to this issue.
https://github.com/pnpm/pnpm/issues/4705

>This has nothing to do with peer dependencies. It's about running scripts in a different environment than they were installed with, leading to the error:

MH4GF: OK, I think this is good for now!

PR #516, opened 2025-01-20
📝 Remove Explore Diagrams section from documentation

FunamaYukina:
## Summary

As the galaxy page is not ready for launch, we have temporarily removed the related navigation.
Furthermore, there were no other navigation paths to the galaxy page.
ss 2706


MH4GF (approved):
Thanks!

PR #517, opened 2025-01-20
📝 Update docs/cli/ci-cd.mdx

hoshinotsuyoshi:
## Summary

📝 Update docs/cli/ci-cd.mdx

preview: https://liam-docs-h8vdfew31-route-06-core.vercel.app/docs/cli/ci-cd

### DIFF
--- a/frontend/apps/docs/content/docs/cli/ci-cd.mdx
+++ b/frontend/apps/docs/content/docs/cli/ci-cd.mdx
@@ -1,6 +1,163 @@
 ---
 title: CI/CD Integration
-description: Explains how to integrate Liam ERD into your build pipeline, automatically update diagrams, and publish them on every commit.
 ---
 
-TBD
+import { File, Folder, Files } from "fumadocs-ui/components/files";
+
+## Overview
+
+By integrating Liam ERD into your CI/CD pipeline, you can **automatically generate ER diagrams on every commit and host their latest versions**.  
+This ensures that your entire team and all stakeholders can consistently access the most up-to-date database structure.
+
+### Why is Hosting the Latest ER Diagram Important?
+
+- **Streamlined Onboarding**  
+  When new or cross-functional team members need to understand the database structure, there’s no need to dig through source code or spreadsheets. The information is consistently available and always current.
+- **Improved Communication**  
+  Visual representations of the data structure make it easier to discuss, clarify, and align on database design among developers, product managers, customer support, data analysts, and more.
+- **Drawbacks of Manual Updates**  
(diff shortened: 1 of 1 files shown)


frontend/apps/docs/content/docs/cli/ci-cd.mdx
```diff
@@ -1,6 +1,163 @@
...
+    <Folder name="assets" defaultOpen>
+      <File name="favicon-xxxxxxxx.ico" />
+      <File name="index-xxxxxxxx.css" />
+      <File name="index-xxxxxxxx.js" />
+    </Folder>
+    <File name="favicon.ico" />
+    <File name="index.html" />
+    <File name="schema.json" />
+  </Folder>
+</Files>
```
MH4GF, line 84: nice 👍🏻

frontend/apps/docs/content/docs/cli/ci-cd.mdx
```diff
@@ -1,6 +1,163 @@
...
+
+## Common Pitfalls
+
+### Security and Privacy
+- If you are using a static hosting service like Cloudflare Pages, Vercel, or Netlify, **avoid exposing your private repository schema** by default.
+- Configure organizational authentication so that **only internal members** can view it.
+
+### Performance with Large Schemas
+- If your schema contains a large number of tables, initial load times can increase.
+
```
MH4GF, line 155: This one doesn't seem to need explaining. What do you think?

MH4GF (approved):
There is some duplication with other pages, but I don't think we need to worry about that right now.
It's great that it's so polite and the user can see everything on this page!
LGTM 👍🏻

frontend/apps/docs/content/docs/cli/ci-cd.mdx
hoshinotsuyoshi: @MH4GF My concern was that it might become unstable when handling a very large number of tables (such as 1000+).  

However, since there's no concrete workaround at the moment, it might be fine to leave it out. I'll remove it!

MH4GF: > However, since there's no concrete workaround at the moment, it might be fine to leave it out.

I think so too 💭 Let's discuss solutions in the future.

hoshinotsuyoshi, line 155:
```suggestion
```

hoshinotsuyoshi:
> There is some duplication with other pages, but I don't think we need to worry about that right now.

Oh, I missed this! Yes, I think it's fine to have some duplication for now!

PR #514, opened 2025-01-20
💄  Update Open Graph image

FunamaYukina:
## Summary

I've updated the OG images for the docs, erd-web, and cli applications.

### docs

preview: https://liam-docs-g94cbkqn1-route-06-core.vercel.app/docs

- Add a new OG image
- Replace the TOP page image with an OG image.
ss 2708

### erd-web

preview: https://liam-erd-hh7qhu5ey-route-06-core.vercel.app/erd/p/github.com/mastodon/mastodon/blob/main/db/schema.rb

- Add a new OG image
ss 2709

### CLI

preview: https://liam-erd-sample-em1z6ea99-route-06-core.vercel.app/

- Add a new OG image

ss 2710

PR #518, opened 2025-01-20
🐛 Add environment variable for base URL and update metadata generation

FunamaYukina:
## Summary

In [the previous PR](https://github.com/liam-hq/liam/pull/514), the image path for the docs was not set up correctly, so I've added it now.

ss 2717

Although I've updated the paths in this PR, the preview environment continues to use the production paths.💭

preview: https://liam-docs-3w1g3rdn9-route-06-core.vercel.app/docs

ss 2718

### DIFF
--- a/frontend/apps/docs/.env.production
+++ b/frontend/apps/docs/.env.production
@@ -1,0 +1,1 @@
+NEXT_PUBLIC_BASE_URL="https://liam-docs-sigma.vercel.app"
--- a/frontend/apps/docs/app/docs/.env.development
+++ b/frontend/apps/docs/app/docs/.env.development
@@ -1,0 +1,1 @@
+NEXT_PUBLIC_BASE_URL="http://localhost:3002"
--- a/frontend/apps/docs/app/docs/.env.development
+++ b/frontend/apps/docs/app/docs/.env.development
@@ -1,1 +1,0 @@
-NEXT_PUBLIC_BASE_URL="http://localhost:3002"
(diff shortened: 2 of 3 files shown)


MH4GF:
@FunamaYukina 
Q: Was this problem only in the development environment? Was it also in the production environment?

MH4GF:
@FunamaYukina OK, I understand that you set the environment variable from the Vercel dashboard.
I mistakenly thought the problem was in the development environment because only .env.development was added.

`NEXT_PUBLIC_BASE_URL` is not secret variables, How about adding .env.production and commit?

FunamaYukina:
@MH4GF 
My apologies🙇‍♀️. I overlooked mentioning that I've configured environment variables on Vercel. Given that this information isn't classified, I've created an ``env.production`` file as suggested.（I have removed the ``environment variable`` from Vercel.）

https://github.com/liam-hq/liam/pull/518/commits/3f05b5172e6425aa7e2840dc114d3d8f15ac9606

I've set default values as I determined that the ``env.development`` file was redundant.
Additionally, I've fixed the incorrect path to the env file.

MH4GF:
@FunamaYukina 
It's not a bad idea to set it up in Vercel!
But I think it would be easier to understand this if it were clearly stated in the code. Thanks!!

MH4GF (approved):
LGTM 👍🏻 Thanks for correcting!

FunamaYukina:
Thank you for your review!!✨

PR #519, opened 2025-01-20
🐛 Fix prisma relationship direction

sasamuku:
## Summary

Fixed Prisma relationship direction.

|before|after|
|--|--|
|image|image|

### DIFF
--- a/frontend/packages/db-structure/src/parser/prisma/parser.ts
+++ b/frontend/packages/db-structure/src/parser/prisma/parser.ts
@@ -37,13 +37,19 @@ async function parsePrismaSchema(schemaString: string): Promise<ProcessResult> {
 
   for (const model of dmmf.datamodel.models) {
     for (const field of model.fields) {
-      if (field.relationName) {
+      if (
+        field.relationName &&
+        (field.relationFromFields?.length ?? 0) > 0 &&
+        (field.relationToFields?.length ?? 0) > 0 &&
+        field.relationFromFields?.[0] &&
+        field.relationToFields?.[0]
+      ) {
         const relationship: Relationship = {
           name: field.relationName,
-          primaryTableName: model.name,
-          primaryColumnName: field.relationFromFields?.[0] || '',
-          foreignTableName: field.type,
-          foreignColumnName: field.relationToFields?.[0] || '',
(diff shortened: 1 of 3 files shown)


frontend/packages/db-structure/src/parser/prisma/parser.ts
```diff
@@ -37,13 +37,19 @@ async function parsePrismaSchema(schemaString: string): Promise<ProcessResult> {
 
   for (const model of dmmf.datamodel.models) {
     for (const field of model.fields) {
-      if (field.relationName) {
+      if (
+        field.relationName &&
+        (field.relationFromFields?.length ?? 0) > 0 &&
+        (field.relationToFields?.length ?? 0) > 0 &&
+        field.relationFromFields?.[0] &&
+        field.relationToFields?.[0]
```
sasamuku, line 45: 📝 
Both the `users` model and the `posts` model have fields related to relations.
Since the relation of the model with foreign key (which means `posts`) has information about primaryKey and foreignKey, I added an if condition to use that one.


```js
// users.posts
{
  name: 'posts',
  kind: 'object',
  isList: true,
  isRequired: true,
  isUnique: false,
  isId: false,
  isReadOnly: false,
  hasDefaultValue: false,
  type: 'posts',
  nativeType: null,
  relationName: 'postsTousers',
  relationFromFields: [],
  relationToFields: [],
  isGenerated: false,
  isUpdatedAt: false
}
```

```js
// posts.user
{
  name: 'user',
  kind: 'object',
  isList: false,
  isRequired: true,
  isUnique: false,
  isId: false,
  isReadOnly: false,
  hasDefaultValue: false,
  type: 'users',
  nativeType: null,
  relationName: 'postsTousers',
  relationFromFields: [ 'userId' ],
  relationToFields: [ 'id' ],
  isGenerated: false,
  isUpdatedAt: false
}
```

hoshinotsuyoshi (approved):
I also tested at local like below:

```
 node dist-cli/bin/cli.js erd build --input https://github.com/langfuse/langfuse/blob/ddee827e22d6d4633ad7431d5f8063aaff305b47/packages/shared/prisma/schema.prisma --format prisma
```

LGTM!

PR #522, opened 2025-01-21
✨️  Support parsing default values in Prisma schema 

hoshinotsuyoshi:
## Summary

Support parsing default values in Prisma schema
- Added `@prisma/generator-helper` as a dependency.
    - for  `extractDefaultValue` return-type 
- Implemented `extractDefaultValue` function to correctly parse default values in Prisma models.
- Added test cases for parsing default values of type `String`, `Int`, and `Boolean`.

## Testing

- erd-web works 
     - preview: https://liam-erd-8a8524u0h-route-06-core.vercel.app/erd/p/github.com/langfuse/langfuse/blob/cf29c6b7e447cf1aec7a8d50ae6a877c3844b7cd/packages/shared/prisma/schema.prisma

### DIFF
--- a/frontend/packages/db-structure/src/parser/prisma/index.test.ts
+++ b/frontend/packages/db-structure/src/parser/prisma/index.test.ts
@@ -70,6 +70,72 @@ describe(processor, () => {
       expect(value).toEqual(expected)
     })
 
+    it('default value as string', async () => {
+      const { value } = await processor(`
+        model users {
+          id   Int    @id @default(autoincrement())
+          description String @default("user's description")
+        }
+      `)
+
+      const expected = userTable({
+        columns: {
+          description: aColumn({
+            name: 'description',
+            type: 'String',
+            default: "user's description",
+            notNull: true,
+          }),
+        },
+      })
+
+      expect(value).toEqual(expected)
+    })
+
+    it('default value as integer', async () => {
+      const { value } = await processor(`
+        model users {
+          id   Int    @id @default(autoincrement())
+          age  Int    @default(30)
+        }
+      `)
+
+      const expected = userTable({
+        columns: {
+          age: aColumn({
+            name: 'age',
+            type: 'Int',
+            default: 30,
+            notNull: true,
+          }),
+        },
+      })
+
+      expect(value).toEqual(expected)
+    })
+
+    it('default value as boolean', async () => {
+      const { value } = await processor(`
+        model users {
+          id     Int     @id @default(autoincrement())
+          active Boolean @default(true)
+        }
+      `)
+
+      const expected = userTable({
+        columns: {
+          active: aColumn({
+            name: 'active',
+            type: 'Boolean',
+            default: true,
+            notNull: true,
+          }),
+        },
+      })
+
+      expect(value).toEqual(expected)
+    })
+
     it('relationship', async () => {
       const { value } = await processor(`
         model users {
--- a/frontend/packages/db-structure/src/parser/prisma/parser.ts
+++ b/frontend/packages/db-structure/src/parser/prisma/parser.ts
@@ -73,12 +73,15 @@ async function parsePrismaSchema(schemaString: string): Promise<ProcessResult> {
 function extractDefaultValue(field: DMMF.Field) {
   const value = field.default?.valueOf()
   const defaultValue = value === undefined ? null : value
-  // NOTE: When `@default(autoincrement())` is specified, defaultValue becomes
-  // an object in the form of `{"name":"autoincrement","args":[]}`.
-  // For now, to align with other parsers, return null if the value is an object.
-  return typeof defaultValue === 'object' && defaultValue !== null
-    ? null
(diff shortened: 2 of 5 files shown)


frontend/packages/db-structure/src/parser/prisma/index.test.ts
```diff
@@ -70,6 +70,72 @@ describe(processor, () => {
       expect(value).toEqual(expected)
     })
 
+    it('default value as string', async () => {
+      const { value } = await processor(`
+        model users {
+          id   Int    @id @default(autoincrement())
+          description String @default("user's description")
+        }
```
hoshinotsuyoshi, line 78: sql(postgres):

```sql
CREATE TABLE "users" (
    "id" SERIAL NOT NULL,
    "description" TEXT DEFAULT 'user''s description',

    CONSTRAINT "users_pkey" PRIMARY KEY ("id")
);
```

frontend/packages/db-structure/src/parser/prisma/index.test.ts
```diff
@@ -70,6 +70,72 @@ describe(processor, () => {
...
+
+      expect(value).toEqual(expected)
+    })
+
+    it('default value as integer', async () => {
+      const { value } = await processor(`
+        model users {
+          id   Int    @id @default(autoincrement())
+          age  Int    @default(30)
+        }
```
hoshinotsuyoshi, line 100: sql(postgres):

```sql
CREATE TABLE "users3" (
    "id" SERIAL NOT NULL,
    "age" INTEGER NOT NULL DEFAULT 30,

    CONSTRAINT "users3_pkey" PRIMARY KEY ("id")
);
```

frontend/packages/db-structure/src/parser/prisma/index.test.ts
```diff
@@ -70,6 +70,72 @@ describe(processor, () => {
...
+
+      expect(value).toEqual(expected)
+    })
+
+    it('default value as boolean', async () => {
+      const { value } = await processor(`
+        model users {
+          id     Int     @id @default(autoincrement())
+          active Boolean @default(true)
+        }
```
hoshinotsuyoshi, line 122: sql(postgres):

```sql
CREATE TABLE "users" (
    "id" SERIAL NOT NULL,
    "active" BOOLEAN NOT NULL DEFAULT true,

    CONSTRAINT "users_pkey" PRIMARY KEY ("id")
);
```

frontend/packages/db-structure/src/parser/prisma/parser.ts
```diff
@@ -68,4 +70,15 @@ async function parsePrismaSchema(schemaString: string): Promise<ProcessResult> {
   }
 }
 
+function extractDefaultValue(field: DMMF.Field) {
+  const value = field.default?.valueOf()
+  const defaultValue = value === undefined ? null : value
+  // NOTE: When `@default(autoincrement())` is specified, defaultValue becomes
+  // an object in the form of `{"name":"autoincrement","args":[]}`.
+  // For now, to align with other parsers, return null if the value is an object.
```
hoshinotsuyoshi, line 78: How about this?

sasamuku: How about accepting `FieldDefaultScalar` only for now?
note comment and implementation will be more simple.

frontend/packages/db-structure/src/parser/prisma/parser.ts
```diff
@@ -68,4 +70,15 @@ async function parsePrismaSchema(schemaString: string): Promise<ProcessResult> {
...
+  return typeof defaultValue === 'object' && defaultValue !== null
+    ? null
+    : defaultValue
+}
```
sasamuku, line 82: >How about accepting FieldDefaultScalar only for now?

Example: 

```suggestion
  return typeof defaultValue === 'string' ||
    typeof defaultValue === 'number' ||
    typeof defaultValue === 'boolean'
    ? defaultValue
    : null
}
```

MH4GF (approved):
The most important thing is to have a test case, and it looks good because it is coming through 👍🏻 

I'll let you be the judge on this point:
https://github.com/liam-hq/liam/pull/522/files#r1923368735

frontend/packages/db-structure/src/parser/prisma/parser.ts
hoshinotsuyoshi: Thank you. I think `FieldDefaultScalar[]` is minor.  So I think we accept only `FieldDefaultScalar` for now.

sasamuku: There is no critical issue, the above comment is nits 👍

frontend/packages/db-structure/src/parser/prisma/parser.ts
hoshinotsuyoshi: Nice!

PR #523, opened 2025-01-21
📝 Add prisma schema header to tests for improved parsing validation

sasamuku:
## Summary

There are several type attributes supported by specific db engine such as `@db.BigInt`.
ref: https://www.prisma.io/docs/orm/reference/prisma-schema-reference#model-field-scalar-types

If we use default connector, these attributes will raise validation error:

```ts
GetDmmfError: Prisma schema validation - (get-dmmf wasm)
Error code: P1012
error: Native type BigInt is not supported for Default connector.
  -->  schema.prisma:6
   | 
 5 |         model users {
 6 |           id   BigInt    @id @db.BigInt @default(autoincrement())
   | 
```

So I've added the postgresql provider on each test cases.
ref: https://www.prisma.io/docs/orm/overview/databases/postgresql

### DIFF
--- a/frontend/packages/db-structure/src/parser/prisma/index.test.ts
+++ b/frontend/packages/db-structure/src/parser/prisma/index.test.ts
@@ -27,9 +27,22 @@ describe(processor, () => {
       },
     })
 
+  const prismaSchemaHeader = `
+    generator client {
+      provider = "prisma-client-js"
+    }
+
+    datasource db {
+      provider = "postgresql"
+      url = env("DATABASE_URL")
+    }
+  `
+
   describe('should parse prisma schema correctly', () => {
     it('not null', async () => {
       const { value } = await processor(`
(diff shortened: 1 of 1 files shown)


frontend/packages/db-structure/src/parser/prisma/index.test.ts
hoshinotsuyoshi, line 1: I added other tests in https://github.com/liam-hq/liam/pull/522  ,  so I will add these headers as well:

https://github.com/liam-hq/liam/pull/531 https://github.com/liam-hq/liam/pull/531/commits/fbbb03a54891128bf85a3588eaee920128e65974

hoshinotsuyoshi: And some refactor added in https://github.com/liam-hq/liam/pull/531 . Please check it out!

PR #525, opened 2025-01-21
🐛 docs: Modifying styles for mobile devices

FunamaYukina:
## Summary

I've resolved the mobile design issues. I've removed the global ``--fd-sidebar-width`` setting and reverted it to the default ``260px``. Since the sidebar was designed with a width of ``260px``, this should address the problem.

| Before  | After |
| ss 2728 | ss 2727 |

## Testing

https://liam-docs-rokwzylu4-route-06-core.vercel.app/docs

### DIFF
--- a/frontend/apps/docs/app/global.css
+++ b/frontend/apps/docs/app/global.css
@@ -61,8 +61,6 @@
     --fd-accent-foreground: var(--liam-gray-500);
 
     --fd-ring: 221 83% 53%;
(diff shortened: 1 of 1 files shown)


junkisai (approved):
The default value was 16.25rem, so I didn't need to overwrite it. 🙏

FunamaYukina:
Oh, I see!

PR #526, opened 2025-01-21
✨ Add support for table comments in Prisma schema

sasamuku:
## Summary

Add support for table comments in Prisma schema.

> [!NOTE]
>Prisma does not yet support table and column comments.
> - https://github.com/prisma/prisma/issues/8703
>I mean, Prisma comment feature is only attached within the AST and is not applied to user database.
> ```js
> /// This comment will get attached to the `User` node in the AST
> model User {
>   /// This comment will get attached to the `id` node in the AST
>   id     Int   @default(autoincrement())
>   // This comment is just for you
>   weight Float /// This comment gets attached to the `weight` node
> }
> ```
> https://www.prisma.io/docs/orm/prisma-schema/overview#comments

### DIFF
--- a/frontend/packages/db-structure/src/parser/prisma/index.test.ts
+++ b/frontend/packages/db-structure/src/parser/prisma/index.test.ts
@@ -92,6 +92,21 @@ describe(processor, () => {
       expect(value).toEqual(expected)
     })
 
+    it('table comment', async () => {
+      const { value } = await processor(`
+        /// store our users.
+        model users {
+          id   Int    @id @default(autoincrement())
+        }
+      `)
+
+      const expected = userTable({
+        comment: 'store our users.',
+      })
+
+      expect(value).toEqual(expected)
+    })
+
     it('relationship', async () => {
       const { value } = await processor(`
         model users {
--- a/frontend/packages/db-structure/src/parser/prisma/parser.ts
+++ b/frontend/packages/db-structure/src/parser/prisma/parser.ts
@@ -30,7 +30,7 @@ async function parsePrismaSchema(schemaString: string): Promise<ProcessResult> {
     tables[model.name] = {
       name: model.name,
       columns,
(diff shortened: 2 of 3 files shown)


frontend/packages/db-structure/src/parser/prisma/index.test.ts
```diff
@@ -70,6 +70,21 @@ describe(processor, () => {
       expect(value).toEqual(expected)
     })
 
+    it('table comment', async () => {
+      const { value } = await processor(`
+        /// store our users.
```
hoshinotsuyoshi, line 75: I see!

frontend/packages/db-structure/src/parser/prisma/parser.ts
```diff
@@ -30,7 +30,7 @@ async function parsePrismaSchema(schemaString: string): Promise<ProcessResult> {
     tables[model.name] = {
       name: model.name,
       columns,
-      comment: null,
+      comment: model.documentation ?? null,
```
hoshinotsuyoshi, line 33: > I mean, Prisma comment feature is only attached within the AST and is not applied to user database.


I think using `model.documentation` for comments makes sense. What do you think? @MH4GF cc: @sasamuku

sasamuku: I think so, better than nothing.
I would like to someone else's thought👍

MH4GF: It makes sense because it seems to be information that users want to see!

hoshinotsuyoshi (approved):
LGTM!

FunamaYukina:
There was a merge conflict after I merged https://github.com/liam-hq/liam/pull/524 🙏

MH4GF (approved):
LGTM 👍🏻

PR #527, opened 2025-01-21
🔧 Setup GTM for ERD Web

sasamuku:
## Summary

Setup GTM for Liam ERD Web.

There is Hydration Error in local development:

```
Console Error

Hydration failed because the server rendered HTML didn't match the client. As a result this tree will be regenerated on the client. This can happen if a SSR-ed Client Component used

- A server/client branch `if (typeof window !== 'undefined')`.
- Variable input such as `Date.now()` or `Math.random()` which changes each time it's called.
- Date formatting in a user's locale which doesn't match the server.
- External changing data without sending a snapshot of it along with the HTML.
- Invalid HTML tag nesting.

It can also happen if the client has a browser extension installed which messes with the HTML before React loaded.

See more info here: https://nextjs.org/docs/messages/react-hydration-error
```

I've tried `next/script` [ref](https://morganfeeney.com/guides/nextjs/how-to-integrate-google-tag-manager-with-nextjs), but the error still occur.

However, getting the tags seems to be no problem:

image

So, as I go through the GTM settings, we look for the cause of the error.


frontend/apps/erd-web/app/layout.tsx
```diff
@@ -2,6 +2,7 @@ import type { Metadata } from 'next'
 import { Geist, Geist_Mono } from 'next/font/google'
 import type React from 'react'
 import './globals.css'
+import { GoogleTagManager } from '@next/third-parties/google'
```
sasamuku, line 5: ref: https://nextjs.org/docs/app/building-your-application/optimizing/third-party-libraries#google-tag-manager

frontend/apps/erd-web/app/layout.tsx
```diff
@@ -35,6 +36,9 @@ export default function RootLayout({
 }>) {
   return (
     <html lang="en">
+      {process.env.NEXT_PUBLIC_ENV_NAME === 'development' && (
```
sasamuku, line 39: 📝 We have to prepare consent component, GTM is available on only development for now.

MH4GF (approved):
> So, as I go through the GTM settings, we look for the cause of the error.

I understand, LGTM!

hoshinotsuyoshi (approved):

PR #528, opened 2025-01-21
🐛 Change the Rive src attribute to an absolute path.

FunamaYukina:
## Summary

We're encountering an issue where Mr. Jack is not showing up on https://liambx.com/docs. 

ss 2731

The problem lies in the riv file, which is currently referencing ``https://liambx.com/rivs`` instead of the correct path ``https://liam-docs-sigma.vercel.app/rivs``. To fix this, I've updated the source to use an absolute path. While I can't verify this until after deployment, I expect that this change will resolve the display issue.


MH4GF (approved):
LGTM 👍🏻

PR #529, opened 2025-01-21
✅ Add tests for unique fields in Prisma model

FunamaYukina:
## Summary

I've added tests for the unique constraint, which is already supported in the Prisma parser.


sasamuku (approved):
LGTM!

PR #533, opened 2025-01-22
🐛 fix: exclude model type from columns in Prisma parser 

hoshinotsuyoshi:
## Summary

Exclude model type from columns

| before | after |
|--------|--------|
| before  | after |

## Testing

see preview https://liam-erd-o43orap6s-route-06-core.vercel.app/erd/p/github.com/langfuse/langfuse/blob/cf29c6b7e447cf1aec7a8d50ae6a877c3844b7cd/packages/shared/prisma/schema.prisma

### DIFF
--- a/frontend/packages/db-structure/src/parser/prisma/parser.ts
+++ b/frontend/packages/db-structure/src/parser/prisma/parser.ts
@@ -13,9 +13,11 @@ async function parsePrismaSchema(schemaString: string): Promise<ProcessResult> {
   const relationships: Record<string, Relationship> = {}
   const errors: Error[] = []
 
+  const modelNames = dmmf.datamodel.models.map((model) => model.name)
   for (const model of dmmf.datamodel.models) {
     const columns: Columns = {}
     for (const field of model.fields) {
+      if (modelNames.includes(field.type)) continue
       const defaultValue = extractDefaultValue(field)
       columns[field.name] = {
         name: field.name,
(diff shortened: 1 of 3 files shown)


frontend/packages/db-structure/src/parser/prisma/parser.ts
```diff
@@ -13,9 +13,11 @@ async function parsePrismaSchema(schemaString: string): Promise<ProcessResult> {
   const relationships: Record<string, Relationship> = {}
   const errors: Error[] = []
 
+  const modelNames = dmmf.datamodel.models.map((model) => model.name)
   for (const model of dmmf.datamodel.models) {
     const columns: Columns = {}
     for (const field of model.fields) {
+      if (modelNames.includes(field.type)) continue
```
sasamuku, line 20: The following code might work 👀
Please check it out!

```suggestion
  for (const model of dmmf.datamodel.models) {
    const columns: Columns = {}
    for (const field of model.fields) {
      if (field.relationName) continue
```

hoshinotsuyoshi: https://github.com/liam-hq/liam/pull/533/commits/519b09aeeae87b5e2b98dca3aa37dd2bc259d73d I've checked it works. No difference in schema.json

sasamuku (approved):
Very helpful.
LGTM!

PR #535, opened 2025-01-22
🐛 Change the Rive src attribute to an absolute path.(light ver)

FunamaYukina:
## Summary

We previously encountered an issue where Mr. Jack was not being displayed. This was resolved in pull request #528.
While that PR only addressed **JackAnimationForDark**, we've now made the same fix for **JackAnimationForLight** as well.
（As **JackAnimationForLight** is not currently in use, we are unable to verify its functionality at this time.）


hoshinotsuyoshi (approved):
LGTM!

PR #537, opened 2025-01-22
📝 Add message from Mr.Jack

kumanoayumi:
## Summary

I added message from Mr.Jack.

https://github.com/user-attachments/assets/e8184b8a-7f26-4210-9a39-bf336d5e4e64

## Changes

- Add Mr. Jack gif animation asssts
- Update README.md


FunamaYukina (commented):
It's super cute🫶! I think it would look great even bigger.🤔

When the width is 32, the result will be as follows:
ss 2737

I'd like to hear what the other members think!

MH4GF:
Great!!!😄 

> When the width is 32, the result will be as follows:

I also seemed to be able to make it a little bigger. That would be more eye-catching!

kumanoayumi:
@FunamaYukina @MH4GF Thank you for feedback,
https://github.com/user-attachments/assets/93510bef-9ec5-4f25-94bc-f8a4457f65d2

With inline, the row height will be a little high, but I'll try to increase the size to 40px!

FunamaYukina (approved):
Thank you!! Looks Great to me.🥰

MH4GF (approved):
Nice job 😆 LGTM 🚀

PR #539, opened 2025-01-22
Fix package-install tab content design

FunamaYukina:

### DIFF
--- a/frontend/apps/docs/content/docs/index.mdx
+++ b/frontend/apps/docs/content/docs/index.mdx
@@ -4,6 +4,7 @@ description: Liam ERD is a tool that effortlessly generates beautiful and easy-t
 ---
 
 import { Package, Globe } from "lucide-react";
+import { Tab, Tabs } from 'fumadocs-ui/components/tabs'; // For package-install code blocks
 
 ![Liam logo](/images/liam_erd.png)
 


frontend/apps/docs/content/docs/index.mdx
```diff
@@ -4,6 +4,7 @@ description: Liam ERD is a tool that effortlessly generates beautiful and easy-t
 ---
 
 import { Package, Globe } from "lucide-react";
+import { Tab, Tabs } from 'fumadocs-ui/components/tabs'; // For package-install code blocks
```
FunamaYukina, line 7: I considered adding an ignore comment for Biome, but given that the Biome check was successful, I opted to keep only the comment.

https://github.com/liam-hq/liam/blob/8305af4b543df264775be0cb1ea4ae1af4ad0239/frontend/apps/docs/biome.jsonc#L9

sasamuku:
Other pages also have a problem?👀
image
https://liam-docs-9uhdhmfsu-route-06-core.vercel.app/docs/cli

FunamaYukina:
@sasamuku Oh!! Thanks for letting me know🙏. I'll check it out.

FunamaYukina:
I made the same correction to the CLI page!

https://liam-docs-pa9xqad2y-route-06-core.vercel.app/docs/cli

FunamaYukina:
@junkisai 
I thought the same! The text color in the TabContent is also different. 
I'll try implementing a copy button as an option in the Tab component! 👍Thank you for your support!

FunamaYukina:
I've created a ``CopyButton`` component and integrated it into the ``Tabs`` component so that users can easily copy the content of each tab.I referred to the code in the codeblock on fumadocs👀.（https://github.com/fuma-nama/fumadocs/blob/dev/packages/ui/src/components/codeblock.tsx）

ss 2754
ss 2755

junkisai (approved):
Thank you for adjusting the code!
It looks good and has been modified according to the design. 🎉

PR #301, opened 2024-12-18
:children_crossing: Sort table nodes by name in LeftPane component

hoshinotsuyoshi:
## Summary

Sort  var `tableNodes` .

PR #544, opened 2025-01-22
Support glob pattern as input paths

yukukotani:
## Summary
Add glob pattern support for the `--input` option in `erd build` command, allowing users to specify multiple schema files using glob patterns.

## Related Issue
N/A

## Changes
- Modified `getInputContent.ts` to support glob patterns for file path resolution
- Updated the `--input` option description to indicate glob pattern support
- Added glob package as a dependency
- Added comprehensive test cases for glob pattern functionality:
  - Multiple file matching and content combination
  - Error handling for non-existent files
  - Error handling for patterns with no matches

## Testing

  ```bash
  # Single file (existing behavior)
  liam erd build --input schema.sql --format postgresql

  # Multiple files using glob
  liam erd build --input "db/migrate/*.sql" --format postgresql

  # Multiple directories
  liam erd build --input "{db1,db2}/*.sql" --format postgresql
  ```

## Other Information
- The implementation maintains backward compatibility with existing file path and URL support
- All changes are generated by Cursor Composer (wow!)

### DIFF
--- a/frontend/packages/cli/src/cli/erdCommand/getInputContent.ts
+++ b/frontend/packages/cli/src/cli/erdCommand/getInputContent.ts
@@ -1,5 +1,6 @@
 import fs from 'node:fs'
 import { URL } from 'node:url'
+import { glob } from 'glob'
 
 function isValidUrl(url: string): boolean {
   try {
@@ -15,11 +16,22 @@ function isGitHubFileUrl(url: string): boolean {
   return parsedUrl.hostname === 'github.com' && url.includes('/blob/')
 }
 
-function readLocalFile(filePath: string): string {
-  if (!fs.existsSync(filePath)) {
-    throw new Error('Invalid input path. Please provide a valid file.')
+async function readLocalFiles(pattern: string): Promise<string> {
+  const files = await glob(pattern)
+  if (files.length === 0) {
+    throw new Error('No files found matching the pattern. Please provide valid file(s).')
   }
-  return fs.readFileSync(filePath, 'utf8')
+
+  const contents = await Promise.all(
+    files.map(async (filePath) => {
+      if (!fs.existsSync(filePath)) {
+        throw new Error(`File not found: ${filePath}`)
+      }
+      return fs.readFileSync(filePath, 'utf8')
+    }),
+  )
+
+  return contents.join('\n')
 }
 
 async function downloadGitHubRawContent(githubUrl: string): Promise<string> {
@@ -40,7 +52,7 @@ async function downloadFile(url: string): Promise<string> {
 
 export async function getInputContent(inputPath: string): Promise<string> {
   if (!isValidUrl(inputPath)) {
-    return readLocalFile(inputPath)
+    return await readLocalFiles(inputPath)
   }
 
   return isGitHubFileUrl(inputPath)
--- a/frontend/packages/cli/src/cli/erdCommand/getInputContent.ts
+++ b/frontend/packages/cli/src/cli/erdCommand/getInputContent.ts
@@ -19,7 +19,9 @@ function isGitHubFileUrl(url: string): boolean {
 async function readLocalFiles(pattern: string): Promise<string> {
   const files = await glob(pattern)
   if (files.length === 0) {
-    throw new Error('No files found matching the pattern. Please provide valid file(s).')
+    throw new Error(
+      'No files found matching the pattern. Please provide valid file(s).',
+    )
   }
 
   const contents = await Promise.all(
--- a/.changeset/chilled-dancers-search.md
+++ b/.changeset/chilled-dancers-search.md
@@ -1,0 +1,5 @@
+---
+"@liam-hq/cli": minor
+---
+
+✨ Support glob pattern as input paths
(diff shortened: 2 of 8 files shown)


yukukotani:
Feel free to pend or close this since it's not well discussed.

MH4GF:
@yukukotani Thanks for your contribution! You are the first person outside the team!😆

I am interested in a use case. The one I immediately thought of was drizzle's migration file, but is there one you imagined?

The content looks good to me. Please check the CI!

yukukotani:
@MH4GF 
What an honor to be the first!

Actually, we have an internal CLI that concat multiple sql files and pass it to sqldef. But yeah, this feature would work well for drizzle too.

I've checked lint and build are passing in my local. Please re-check!

MH4GF:
@yukukotani 

> Actually, we have an internal CLI that concat multiple sql files and pass it to sqldef. But yeah, this feature would work well for drizzle too.

I see, you are managing sql files by splitting them up. sqldef means that the CREATE TABLE statement is written declaratively. If so, it would certainly work with this tool 👍🏻 

Thanks for the correction! I'll review soon!

MH4GF (approved):
LGTM 👍🏻 Thanks 🎉 
One more member, please review who it is.

hoshinotsuyoshi:
Regarding the unstable order of `glob` results and parallel file reading using `Promise.all`, which may cause the order of SQL statements in the return value of `getInputContent` to change:

I considered a case where we prepare `1.sql` and `2.sql` and let the CLI built with this PR read them.

```sql
$ cat 1.sql
CREATE TABLE "public"."users" (
    "id" bigserial NOT NULL,
    PRIMARY KEY ("id")
);
```

```sql
$ cat 2.sql
CREATE TABLE "public"."posts" (
    "id" bigserial NOT NULL,
    "user_id" integer,
    PRIMARY KEY ("id")
);

ALTER TABLE ONLY "public"."posts" ADD CONSTRAINT "posts_user_id_fkey" FOREIGN KEY ("user_id") REFERENCES "public"."users" ("id") ON UPDATE NO ACTION ON DELETE NO ACTION;
```

I have confirmed that even if the order of `1.sql` and `2.sql` changes, it does not seem to affect the visual appearance of the resulting HTML.

### **NOTE:**  
In practice, on my local macOS, the order is mostly stable. However, to verify this further, I experimented with different input arguments—passing `--input {1.sql,2.sql}`, `{2.sql,1.sql}`, and `*.sql` to observe the behavior.

### **NOTE:**  

The only difference in `schema.json` is the order of the `tables` field. But it is not problem. ref https://github.com/liam-hq/liam/pull/301

<details><summary>Diff reference</summary>

```diff
--- 1.schema.json	2025-01-23 12:34:56
+++ 2.schema.json	2025-01-23 12:36:58
@@ -1,7 +1,7 @@
 {
   "tables": {
-    "posts": {
-      "name": "posts",
+    "users": {
+      "name": "users",
       "columns": {
         "id": {
           "name": "id",
\ No newline at end of file
@@ -12,23 +12,13 @@
           "unique": false,
           "notNull": true,
           "comment": null
-        },
-        "user_id": {
-          "name": "user_id",
-          "type": "int4",
-          "default": null,
-          "check": null,
-          "primary": false,
-          "unique": false,
-          "notNull": false,
-          "comment": null
         }
       },
       "comment": null,
       "indices": {}
     },
-    "users": {
-      "name": "users",
+    "posts": {
+      "name": "posts",
       "columns": {
         "id": {
           "name": "id",
\ No newline at end of file
@@ -39,6 +29,16 @@
           "unique": false,
           "notNull": true,
           "comment": null
+        },
+        "user_id": {
+          "name": "user_id",
+          "type": "int4",
+          "default": null,
+          "check": null,
+          "primary": false,
+          "unique": false,
+          "notNull": false,
+          "comment": null
         }
       },
       "comment": null,
\ No newline at end of file
```

</details>

---

Given this, I think the unstable order should be acceptable.  🚀

frontend/packages/cli/src/cli/erdCommand/getInputContent.ts
```diff
@@ -15,11 +16,24 @@ function isGitHubFileUrl(url: string): boolean {
   return parsedUrl.hostname === 'github.com' && url.includes('/blob/')
 }
 
-function readLocalFile(filePath: string): string {
-  if (!fs.existsSync(filePath)) {
-    throw new Error('Invalid input path. Please provide a valid file.')
+async function readLocalFiles(pattern: string): Promise<string> {
+  const files = await glob(pattern)
+  if (files.length === 0) {
```
hoshinotsuyoshi, line 21: ❤️

hoshinotsuyoshi:
In the near future, we may need to include the filename along with the line number and other details in error reports when a parsing error occurs. However, since the current error reports are not very detailed, let's consider this when making them more informative.

hoshinotsuyoshi (approved):
LGTM!

yukukotani:
@hoshinotsuyoshi Thanks for your review, and good catch on the order of statements!

In the case of the unstable order causes some problem, we can just sort the result of glob. `Promise.all` is not the cause as [it preserves the order.](https://www.karltarvas.com/order-of-resolved-values-in-promise-all.html)

hoshinotsuyoshi:
@yukukotani 
Oh, I see! I didn’t realize that `Promise.all` preserves the order like that. Thanks for pointing it out!  

And yes, that makes sense—if sorting in `glob` ever becomes necessary, it should be easy to address.

PR #545, opened 2025-01-22
apps/docs: Fix nested `<p>` tags issue in Callout component at .mdx file

hoshinotsuyoshi:
## Summary

preview: https://liam-docs-q12li259w-route-06-core.vercel.app/docs/cli/ci-cd

Tiny fix.
fix nested `<p>` tags issue in Callout component at .mdx file.

before:

after:


junkisai (approved):
I see, it seems like we have to be careful about how I use it when writing documents. 👀

PR #546, opened 2025-01-23
🐛 apps/docs: Fix RSC payload failure

hoshinotsuyoshi:
## Summary

The link might be on the same domain, but it's likely a different application than `docs`, so we don't use Link.

https://liambx.com/docs

current error:


MH4GF (approved):
👍🏻

PR #551, opened 2025-01-23
🔥 Remove debug documentation file

FunamaYukina:
## Summary

With the release of Liam ERD, I eliminated the debug page.
This was added in #232.

ss 2756


hoshinotsuyoshi (approved):
> added in https://github.com/liam-hq/liam/pull/232.

Thank you! I can confirm that only one file should be removed.

PR #567, opened 2025-01-27
💄 Add an override for the Fumadocs `package installation` component.

FunamaYukina:
## Summary

The background color of the areas using the Fumadocs package-install component was inconsistent with the design. I've corrected it to match the design.

### Before
ss 2763

### Design
ss 2764

### After

ss 2766


MH4GF (approved):
This seems to be the only way... LGTM 👍🏻

PR #569, opened 2025-01-27
🐛 fix iframe detection to avoid window defined error

sasamuku:
## Summary

Fixed `ReferenceError: window is not defined`.

```ts
 ⨯ ReferenceError: window is not defined
    at window (app/erd/p/[...slug]/erdViewer.tsx:42:19)
  40 |   }
  41 |   const version = v.parse(versionSchema, versionData)
> 42 |   const isIframe = window !== window.parent
     |                   ^
  43 |
  44 |   return (
  45 |     <div style={{ height: '100vh' }}> {
  digest: '999588878'
}
```

I've confirmed no error on my local environment ✅

### DIFF
--- a/frontend/apps/erd-web/app/erd/p/[...slug]/erdViewer.tsx
+++ b/frontend/apps/erd-web/app/erd/p/[...slug]/erdViewer.tsx
@@ -27,11 +27,11 @@ export default function ERDViewer({
   errorObjects,
   defaultSidebarOpen,
 }: ERDViewerProps) {
-  const [isIframe, setIsIframe] = useState(false)
+  const [isShowCookieConsent, setShowCookieConsent] = useState(false)
 
   useEffect(() => {
     initDBStructureStore(dbStructure)
-    setIsIframe(window !== window.parent)
+    setShowCookieConsent(window === window.parent)
   }, [dbStructure])
 
   const versionData = {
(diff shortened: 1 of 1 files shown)


frontend/apps/erd-web/app/erd/p/[...slug]/erdViewer.tsx
```diff
@@ -27,8 +27,11 @@ export default function ERDViewer({
   errorObjects,
   defaultSidebarOpen,
 }: ERDViewerProps) {
+  const [isIframe, setIsIframe] = useState(false)
```
Comment (MH4GF) [inline frontend/apps/erd-web/app/erd/p/[...slug]/erdViewer.tsx:30]: In this implementation, May appear CookieConsent and disappear momentarily depending of useEffect timing because default value is false.
What do you think that default value changes to true? but it is not readable, I like following:

```ts
const [isShowCookieConsent, setShowCookieConsent] = useState(false)

~~~

setShowCookieConsent(window === window.parent)

~~~

process.env.NEXT_PUBLIC_ENV_NAME !== 'production' && isShowCookieConsent &&
```

MH4GF (commented):
I want to discuss it!

frontend/apps/erd-web/app/erd/p/[...slug]/erdViewer.tsx
Comment (sasamuku) [reply frontend/apps/erd-web/app/erd/p/[...slug]/erdViewer.tsx]: Thanks, It would be better👍
Fixed:
https://github.com/liam-hq/liam/pull/569/commits/d757fe1f31de3ddf9fedf15189a6bd8201bf824e

MH4GF (approved):
LGTM 👍🏻

PR #572, opened 2025-01-27
🔧 Update Prisma column types to match PostgreSQL standards

FunamaYukina:
## Summary

I updated LiamERD to display PostgreSQL data types by referring to https://www.prisma.io/docs/orm/reference/prisma-schema-reference.

I've made it so that enum values are returned as ``enum`` types.

| Before  | After |
|   |   |

### DIFF
--- a/frontend/packages/db-structure/src/parser/prisma/parser.ts
+++ b/frontend/packages/db-structure/src/parser/prisma/parser.ts
@@ -25,7 +25,7 @@ async function parsePrismaSchema(schemaString: string): Promise<ProcessResult> {
       const defaultValue = extractDefaultValue(field)
       columns[field.name] = {
         name: field.name,
-        type: field.type,
+        type: convertToPostgresColumnType(field.type, dmmf.datamodel.enums),
         default: defaultValue,
         notNull: field.isRequired,
         unique: field.isUnique,
@@ -154,6 +154,40 @@ function normalizeConstraintName(constraint: string): ForeignKeyConstraint {
       return 'SET_DEFAULT'
     default:
       return 'NO_ACTION'
+  }
+}
+
+function convertToPostgresColumnType(
+  type: string,
+  enums: readonly DMMF.DatamodelEnum[],
+): string {
+  const enumNames = enums.map((e) => e.name)
+
+  if (enumNames.includes(type)) {
+    return 'enum'
+  }
+
+  switch (type) {
+    case 'String':
+      return 'text'
+    case 'Boolean':
+      return 'boolean'
+    case 'Int':
+      return 'integer'
+    case 'BigInt':
+      return 'bigint'
+    case 'Float':
+      return 'double precision'
+    case 'DateTime':
+      return 'timestamp'
+    case 'Json':
+      return 'jsonb'
+    case 'Decimal':
+      return 'decimal'
+    case 'Bytes':
+      return 'bytea'
+    default:
+      return ''
   }
 }
 
--- a/frontend/packages/db-structure/src/parser/prisma/parser.ts
+++ b/frontend/packages/db-structure/src/parser/prisma/parser.ts
@@ -25,7 +25,7 @@ async function parsePrismaSchema(schemaString: string): Promise<ProcessResult> {
       const defaultValue = extractDefaultValue(field)
       columns[field.name] = {
         name: field.name,
-        type: convertToPostgresColumnType(field.type, dmmf.datamodel.enums),
+        type: convertToPostgresColumnType(field.type),
         default: defaultValue,
         notNull: field.isRequired,
         unique: field.isUnique,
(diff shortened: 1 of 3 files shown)


frontend/packages/db-structure/src/parser/prisma/parser.ts
```diff
@@ -157,4 +157,38 @@ function normalizeConstraintName(constraint: string): ForeignKeyConstraint {
...
+    case 'BigInt':
+      return 'bigint'
+    case 'Float':
+      return 'double precision'
+    case 'DateTime':
+      return 'timestamp'
+    case 'Json':
+      return 'jsonb'
+    case 'Decimal':
+      return 'decimal'
```
FunamaYukina, line 186: Although the exact data type is decimal(65,30), we've used the generic 'decimal' for brevity.
https://www.prisma.io/docs/orm/reference/prisma-schema-reference#decimal

frontend/packages/db-structure/src/parser/prisma/parser.ts
```diff
@@ -157,4 +157,38 @@ function normalizeConstraintName(constraint: string): ForeignKeyConstraint {
...
 
+function convertToPostgresColumnType(
+  type: string,
+  enums: readonly DMMF.DatamodelEnum[],
+): string {
+  const enumNames = enums.map((e) => e.name)
+
+  if (enumNames.includes(type)) {
+    return 'enum'
+  }
```
MH4GF, line 168: In current postgres parser also display real enum value (e.g.  `AnnotationQueueStatus` ).
Need to figure out how to display of enum, I think we can delete this line now and display it as is. What do you think?

frontend/packages/db-structure/src/parser/prisma/parser.ts
```diff
@@ -157,4 +157,38 @@ function normalizeConstraintName(constraint: string): ForeignKeyConstraint {
   }
 }
 
+function convertToPostgresColumnType(
```
MH4GF, line 160: Can you please specify in code comments that we are following this document?
https://www.prisma.io/docs/orm/reference/prisma-schema-reference#model-field-scalar-types

frontend/packages/db-structure/src/parser/prisma/parser.ts
```diff
@@ -157,4 +157,38 @@ function normalizeConstraintName(constraint: string): ForeignKeyConstraint {
...
+): string {
+  const enumNames = enums.map((e) => e.name)
+
+  if (enumNames.includes(type)) {
+    return 'enum'
+  }
+
+  switch (type) {
+    case 'String':
+      return 'text'
```
MH4GF, line 172: If `@db.Char(x)` is defined, it must be referenced.

```hcl
model Model {
  /* ... */
  myField String @db.Char(x)
}
```

But I don't think this needs to be addressed now.

frontend/packages/db-structure/src/parser/prisma/parser.ts
MH4GF: The policy should have been to display actual values whenever possible, even when parsing schema.rb.
So I thought it was ok to display `decimal(65,30)`.

MH4GF (commented):
Thanks!

frontend/packages/db-structure/src/parser/prisma/parser.ts
FunamaYukina: I see your point. I had been under the impression that displaying the ``PostgreSQL`` data type was the standard approach. However, I believe that showing the enum's display name is a more user-friendly solution.👍

frontend/packages/db-structure/src/parser/prisma/parser.ts
FunamaYukina: Yes.You're absolutely right. We also don't support ``real`` types, ``UUID`` types, and others. 
I'll try developing them in the next PR!

frontend/packages/db-structure/src/parser/prisma/parser.ts
FunamaYukina: This has been fixed in commit: https://github.com/liam-hq/liam/pull/572/commits/23d43907fde4a8de8744bfec0f16557b758c9f1e

frontend/packages/db-structure/src/parser/prisma/parser.ts
FunamaYukina: Sorry about that! I've added a comment to the reference.
https://github.com/liam-hq/liam/pull/572/commits/6c16e00647db190b6b37a0cdb597fd55fa6504ed

frontend/packages/db-structure/src/parser/prisma/parser.ts
FunamaYukina: Thank you!That makes sense.  I have updated the code to support the data types as defined in the reference.
https://github.com/liam-hq/liam/pull/572/commits/6d2686608289738edc0bb0f1573384c00c96c334

frontend/packages/db-structure/src/parser/prisma/parser.ts
MH4GF: @FunamaYukina No need to perfect it now. I'll leave it up to you to decide how far you want to go!

MH4GF (approved):
LGTM 👍🏻

FunamaYukina:
Thank you for your review!✨

PR #577, opened 2025-01-27
✨ Add initial tbls parser

sasamuku:
## Summary

- Add initial tbls parser
- Add format for `schema.json`

#### ERD Web:

image

https://liam-erd-b9jxja3n1-route-06-core.vercel.app/erd/p/github.com/k1LoW/tbls/blob/v1.81.0/sample/mysql/schema.json

#### CLI:

image

```sh
$ ./dist-cli/bin/cli.js erd build --input https://github.com/k1LoW/tbls/blob/v1.81.0/sample/mysql/schema.json --format tbls
```

### DIFF
--- a/.changeset/early-files-talk.md
+++ b/.changeset/early-files-talk.md
@@ -1,0 +1,6 @@
+---
+"@liam-hq/db-structure": patch
+"@liam-hq/cli": patch
+---
+
+✨ Add initial tbls parser
--- a/frontend/packages/db-structure/src/parser/supportedFormat/detectFormat.ts
+++ b/frontend/packages/db-structure/src/parser/supportedFormat/detectFormat.ts
@@ -7,6 +7,7 @@ const nameToFormatMap: Record<string, SupportedFormat> = {
   'schema.rb': 'schemarb',
   schemafile: 'schemarb',
   'prisma.schema': 'prisma',
+  'schema.json': 'tbls',
 }
 
 const extensionToFormatMap: Record<string, SupportedFormat> = {
--- a/frontend/packages/db-structure/src/parser/tbls/parser.ts
+++ b/frontend/packages/db-structure/src/parser/tbls/parser.ts
@@ -1,0 +1,96 @@
+import type {
+  Columns,
+  Indices,
+  Relationship,
+  Tables,
+} from '../../schema/index.js'
+import { aColumn, aRelationship, aTable, anIndex } from '../../schema/index.js'
+import type { ProcessResult, Processor } from '../types.js'
+import { defaultRelationshipName } from '../utils/index.js'
+import schema from './schema.generated.js'
+
+async function parseTblsSchema(schemaString: string): Promise<ProcessResult> {
+  const parsedSchema = JSON.parse(schemaString)
+  const result = schema.safeParse(parsedSchema)
+
+  if (!result.success) {
+    return {
+      value: {
+        tables: {},
+        relationships: {},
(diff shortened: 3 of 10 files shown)


frontend/packages/db-structure/src/parser/supportedFormat/detectFormat.ts
```diff
@@ -7,6 +7,7 @@ const nameToFormatMap: Record<string, SupportedFormat> = {
   'schema.rb': 'schemarb',
   schemafile: 'schemarb',
   'prisma.schema': 'prisma',
+  'schema.json': 'tbls',
```
sasamuku, line 10: 📝 `schema.json` is a general name.
I think it's alright to handle it when namespace conflicts in Liam.

frontend/packages/db-structure/src/parser/tbls/parser.ts
```diff
@@ -0,0 +1,96 @@
...
+  const parsedSchema = JSON.parse(schemaString)
+  const result = schema.safeParse(parsedSchema)
+
+  if (!result.success) {
+    return {
+      value: {
+        tables: {},
+        relationships: {},
+      },
+      errors: [new Error('Invalid schema format')],
```
MH4GF, line 22: suggestion: 

```ts
      errors: [new Error(`Invalid schema format: ${result.errors}`)],
```

(I don't know what the error type is, so I can't get this to work.)

MH4GF (approved):
LGTM 👍🏻

frontend/packages/db-structure/src/parser/tbls/parser.ts
sasamuku: Thanks, fixed: a580fb6a158b3eed2de753a78c84882126af8ee1

frontend/packages/db-structure/src/parser/supportedFormat/detectFormat.ts
MH4GF: This is troubling, but I think this is not a problem because it is just inference logic.

.changeset/early-files-talk.md
```diff
@@ -0,0 +1,6 @@
+---
+"@liam-hq/db-structure": patch
+"@liam-hq/cli": patch
```
hoshinotsuyoshi, line 3: Since last Friday, we've been shipping new features as **`minor`** instead of `patch`.   https://github.com/liam-hq/liam/releases/tag/%40liam-hq%2Fcli%400.2.0

For this PR, would `minor` be appropriate? Or should we consider this an `initial` implementation?   @MH4GF  @sasamuku 

```suggestion
"@liam-hq/db-structure": minor
"@liam-hq/cli": minor
```

MH4GF: @hoshinotsuyoshi Thanks!
I thought that `patch` was still good enough for now. 
The intent is available to the user but unstable and treated as experimental.

I would like to make it `minor` at a time when we have completed some implementation!

hoshinotsuyoshi: > The intent is available to the user but unstable and treated as experimental.

That makes sense! Thanks for the clarification. @MH4GF

hoshinotsuyoshi (approved):
LGTM!

.changeset/early-files-talk.md
sasamuku: Thanks, I understood 👍

PR #578, opened 2025-01-27
🔧 prisma: Supports column data types more accurately

FunamaYukina:
## Summary

While our PR for #572 previously supported only the following columns, we've enhanced it to provide more accurate support. For example, we now support columns like ``serial``, ``uuid``, and ``time``

ss 2769

Support for added columns: 
```
char
varchar
bit
varbit
uuid
xml
inet
citext
smallint
smallserial
serial
oid
bigserial
real
money
timestamptz
date
time
timetz
json
```

You can find the UUID and serial number within the ``cal.com`` data schema.
Preview: https://liam-erd-ncg6vs35h-route-06-core.vercel.app/erd/p/github.com/calcom/cal.com/blob/main/packages/prisma/schema.prisma?showMode=ALL_FIELDS&active=EventType

ss 2772

### DIFF
--- a/.changeset/funny-squids-yawn.md
+++ b/.changeset/funny-squids-yawn.md
@@ -1,0 +1,6 @@
+---
+"@liam-hq/db-structure": patch
+"@liam-hq/cli": patch
+---
+
+🔧 Implement convertToPostgresColumnType function for PostgreSQL type conversion and update parser to utilize it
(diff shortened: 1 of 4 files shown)


MH4GF (approved):
Thanks!

FunamaYukina:
Thank you for your review! 💫

PR #579, opened 2025-01-27
📈 Remove Google Tag Manager

junkisai:
## Summary

I have removed the Google Tag Manager tag that was embedded to track how the beta version is being used.


MH4GF (approved):
👍🏻

PR #580, opened 2025-01-27
🔧 Update cli package.json

hoshinotsuyoshi:
Added keywords

### DIFF
--- a/.syncpackrc
+++ b/.syncpackrc
@@ -1,6 +1,12 @@
 {
+  "formatRepository": false,
   "sortFirst": [
     "name",
+    "description",
+    "keywords",
+    "homepage",
+    "bugs",
+    "repository",
     "license",
     "private",
     "version",
--- a/frontend/packages/cli/package.json
+++ b/frontend/packages/cli/package.json
@@ -1,5 +1,18 @@
 {
   "name": "@liam-hq/cli",
+  "description": "Command-line tool designed to generate a web application that displays ER diagrams. See https://liambx.com/docs/cli",
+  "keywords": [
+    "data-visualization",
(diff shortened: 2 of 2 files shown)


.syncpackrc
```diff
@@ -1,6 +1,12 @@
 {
+  "formatRepository": false,
```
hoshinotsuyoshi, line 2: nits

see https://jamiemason.github.io/syncpack/config/format-repository/

hoshinotsuyoshi: When `true`,  

```json
  "repository": {
    "type": "git",
    "url": "https://github.com/liam-hq/liam.git"
  }
```  

becomes  

```json
  "repository": "liam-hq/liam"
```  

This is also valid (see [npm docs](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#repository) ), but I thought providing the full URL explicitly would be more convenient.

frontend/packages/cli/package.json
```diff
@@ -1,5 +1,18 @@
 {
   "name": "@liam-hq/cli",
+  "description": "Command-line tool designed to generate a web application that displays ER diagrams. See https://liambx.com/docs/cli",
+  "keywords": [
+    "data-visualization",
+    "entity-relationship-diagram",
+    "er-diagram",
+    "erd"
```
hoshinotsuyoshi, line 8: Picked from the [GitHub topics of liam-hq/liam](https://github.com/liam-hq/liam) —do we need more?

MH4GF: This is OK for now!

MH4GF (approved):
LGTM 👍🏻

PR #583, opened 2025-01-27
♻️ Add createParserTestCases function and update parser tests

FunamaYukina:
## Summary

To improve consistency, I've standardized the tests using a single test case. However, variations in parser behavior, like ``index names`` and ``default values``, necessitate the redefinition of ``userTable`` for each parser.

### DIFF
--- a/.changeset/nice-starfishes-collect.md
+++ b/.changeset/nice-starfishes-collect.md
@@ -1,6 +1,0 @@
----
-"@liam-hq/db-structure": patch
-"@liam-hq/cli": patch
----
-
-♻️ Add createParserTestCases function and update parser tests
--- a/.changeset/nice-starfishes-collect.md
+++ b/.changeset/nice-starfishes-collect.md
@@ -1,0 +1,6 @@
+---
+"@liam-hq/db-structure": patch
+"@liam-hq/cli": patch
+---
+
+♻️ Add createParserTestCases function and update parser tests
--- a/frontend/packages/db-structure/src/parser/__tests__/testcase.ts
+++ b/frontend/packages/db-structure/src/parser/__tests__/testcase.ts
@@ -1,35 +1,13 @@
 import {
   type Table,
   aColumn,
-  aDBStructure,
-  aTable,
+  type aDBStructure,
   anIndex,
 } from '../../schema/index.js'
 
-const userTable = (override?: Partial<Table>) =>
-  aDBStructure({
-    tables: {
-      users: aTable({
-        name: 'users',
-        columns: {
-          id: aColumn({
-            name: 'id',
-            type: 'bigserial',
-            notNull: true,
-            primary: true,
-            unique: true,
-          }),
-          ...override?.columns,
-        },
-        indices: {
-          ...override?.indices,
-        },
-        comment: override?.comment ?? null,
-      }),
-    },
-  })
-
-export const parserTestCases = {
+export const createParserTestCases = (
+  userTable: (override?: Partial<Table>) => ReturnType<typeof aDBStructure>,
+) => ({
   'table comment': userTable({
     comment: 'store our users.',
   }),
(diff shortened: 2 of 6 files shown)


frontend/packages/db-structure/src/parser/__tests__/testcase.ts
```diff
@@ -1,35 +1,13 @@
...
-          ...override?.indices,
-        },
-        comment: override?.comment ?? null,
-      }),
-    },
-  })
-
-export const parserTestCases = {
+export const createParserTestCases = (
+  userTable: (override?: Partial<Table>) => ReturnType<typeof aDBStructure>,
```
MH4GF, line 9: suggestion: It becomes more obvious if you do the following:

```suggestion
  userTable: (override?: Partial<Table>) => DBStructure,
```

.changeset/nice-starfishes-collect.md
MH4GF, line 1: suggestion: This file is not needed because it don't impact users.

FunamaYukina:
📝 
The tests for tbls also need to be fixed before merging.

frontend/packages/db-structure/src/parser/prisma/index.test.ts
MH4GF, line 1: nice work 🚀

frontend/packages/db-structure/src/parser/prisma/index.test.ts
```diff
@@ -12,11 +13,11 @@ describe(_processor, () => {
           columns: {
             id: aColumn({
               name: 'id',
-              type: 'serial',
+              type: 'bigserial',
               default: 'autoincrement()',
```
MH4GF, line 17: Q: I want to know the reason for splitting the userTable function into separate files. Am I understanding this `default` difference correctly?

frontend/packages/db-structure/src/parser/prisma/parser.ts
```diff
@@ -34,7 +34,7 @@ async function parsePrismaSchema(schemaString: string): Promise<ProcessResult> {
         ),
         default: defaultValue,
         notNull: field.isRequired,
-        unique: field.isUnique,
+        unique: field.isId || field.isUnique,
```
MH4GF, line 37: 👍🏻

MH4GF (approved):
LGTM! I made some nits comments.

frontend/packages/db-structure/src/parser/prisma/index.test.ts
FunamaYukina: You're absolutely right. The biggest reason is that Prisma automatically sets the default to ``autoincrement()``.
And this default value is necessary to differentiate between ``bigint`` and ``bigserial``.

ss 2773


https://www.prisma.io/docs/orm/reference/prisma-schema-reference#postgresql-3

MH4GF: OK, I understand!

.changeset/nice-starfishes-collect.md
FunamaYukina: Yes! Thank you for your advice.🙏
deleted: https://github.com/liam-hq/liam/pull/583/commits/83299025399a017e625d20fa11a96f89081161b2

frontend/packages/db-structure/src/parser/__tests__/testcase.ts
FunamaYukina: Oh, thank you so much.🙏
Fixed: https://github.com/liam-hq/liam/pull/583/commits/83736cfd8d6cca0eaf46d2582618d5bc1161be08

MH4GF (approved):
LGTM 👍🏻

FunamaYukina:
Thank you for your review!
The ``tbls`` tests passed, so it seems like we're good to go.

PR #586, opened 2025-01-28
🔧 Add Google Tag Manager tracking for production

sasamuku:
## Summary

Forgot to set environment variables 😅


MH4GF (approved):
👍🏻

junkisai (commented):
Since GMT_ID is not a value that changes depending on the environment, I thought it might be okay to define it as a constant in the code. 👀 

https://github.com/liam-hq/liam/blob/add_environment_variable_gtm_id_for_prod/frontend/apps/erd-web/app/layout.tsx#L42

MH4GF:
> Since GMT_ID is not a value that changes depending on the environment, I thought it might be okay to define it as a constant in the code. 👀

That makes sense. 👍🏻

sasamuku:
>Since GMT_ID is not a value that changes depending on the environment, I thought it might be okay to define it as a constant in the code.

Okay, Let's do that👍

PR #588, opened 2025-01-28
✨ Add support for unique column constraints in tbls parser

sasamuku:
## Summary

Support unique column constraints (only single column not multi column)

PR #590, opened 2025-01-28
✨ Add support for primary key constraints in `tbls` parser

FunamaYukina:
## Summary

I added support for primary key constraints to the tbls parser!


frontend/packages/db-structure/src/parser/tbls/index.test.ts
```diff
@@ -14,7 +14,7 @@ describe(processor, () => {
               name: 'id',
               type: 'int',
               notNull: true,
-              primary: false, // TODO: should be true
+              primary: false,
```
FunamaYukina, line 17: I was aiming for ``primary: true``, but since I have to add ``constraints`` to every test, I've decided to add a separate test just for primary keys.

frontend/packages/db-structure/src/parser/tbls/parser.ts
```diff
@@ -55,11 +55,18 @@ async function parseTblsSchema(schemaString: string): Promise<ProcessResult> {
     )
 
     for (const tblsColumn of tblsTable.columns) {
+      const isPrimaryKey = !!tblsTable.constraints?.some(
+        (constraint) =>
+          constraint.type === 'PRIMARY KEY' &&
+          constraint.columns?.includes(tblsColumn.name),
+      )
```
sasamuku, line 62: imo:
It is redundant because the process is repetitive: num of column * num of constraints
How about using `Set`? 

ref:

- https://github.com/liam-hq/liam/pull/588

FunamaYukina: Oh, Thank you for your explanation! I'll try fixing it using the ``Set`` method.

FunamaYukina: Thank you for your support. I believe the performance has improved significantly.✨
https://github.com/liam-hq/liam/pull/590/commits/c761ef3244cae2d9a94c38c6e9839d64816bec6d

sasamuku (approved):
LGTM🚀

PR #596, opened 2025-01-28
🔧 Add Cursor AI development rules configuration file

MH4GF:
A file has been added to allow the user to set custom instructions for using cursor as an editor.
I have added it with reference to https://github.com/PatrickJS/awesome-cursorrules, but you may continue to modify it as you see fit.

ref: https://docs.cursor.com/context/rules-for-ai#cursorrules


junkisai (approved):
I learned about cursorrules!😄

PR #598, opened 2025-01-28
✨ Add support for default in `tbls` parser

FunamaYukina:
## Summary

Added support for default values in the tbls parser.
Since ``tblsColumn.default`` in tbls's ``schema.json`` provides values like booleans and numbers as strings (e.g., ``'1'`` or ``'true'``), we have added processing to appropriately convert them to numbers or booleans before returning.


MH4GF (approved):
LGTM 👍🏻

PR #599, opened 2025-01-28
✅ tbls: Add tests for table and column comments.

FunamaYukina:
## Summary

Since table and column comments were already supported by the tbls parser, I added tests for them!


sasamuku (approved):
LGTM🚀🚀

PR #603, opened 2025-01-28
💄 Add Row3 icons to the table detail component and update the header style.

FunamaYukina:
## Summary

I added icons to the headers of both the Table and Column components within the ``TableDetailDrawer``.

| Before  | After |
| ss 2780 | ss 2779 |


junkisai (approved):
LGTM👌

PR #607, opened 2025-01-28
Toggle function for sections within the ``TableDetailDrawer``

FunamaYukina:
## Summary

I've added a toggle feature to each section of the ``TableDetailDrawer``. For now, I've implemented it only for the ``Columns`` section.（So that I can apply it to other sections after the review.）

Preview: https://liam-erd-sample-b9tglfoys-route-06-core.vercel.app/?active=announcements

https://github.com/user-attachments/assets/f97763a9-fadb-4073-b4d6-b959274fe3eb

In the future,

ss 2783

## Changes

https://github.com/liam-hq/liam/pull/607/commits/aecb91a55e2b2b9516a70c6834f7be6d936ac6fe : ✨ Implement collapsible columns in ERD table detail view
https://github.com/liam-hq/liam/pull/607/commits/1a656aa24093125ca30be4d3fc11a0b4b4998c92 : ♻️ Unify the structure and style of each section within the TableDetailDrawer.

## Other Information

I'd like you to check out the animation part!

PR #616, opened 2025-01-29
✨  Add toggle function for all sections in TableDetailDrawer

FunamaYukina:
## Summary

In PR #607, we added a toggle feature to the Columns section. Now, we've extended this functionality to include all remaining sections: ``Indices``, ``Unique``, and ``RelatedTables``.

ss 2788

## Testing

https://github.com/user-attachments/assets/04dbad17-77b0-4e1c-a726-89a949049925

Since the ｀``RelatedTable`` is fixed at the bottom, it may behave in the following way.

https://github.com/user-attachments/assets/3aea6007-0545-452c-964a-94641a9ad50a

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableDetail/Columns/Columns.tsx
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableDetail/Columns/Columns.tsx
@@ -6,7 +6,7 @@ import {
   Rows3 as Rows3Icon,
 } from '@liam-hq/ui'
 import clsx from 'clsx'
-import { type FC, useState } from 'react'
+import { type FC, type MouseEvent, useState } from 'react'
 import styles from './Columns.module.css'
 import { ColumnsItem } from './ColumnsItem'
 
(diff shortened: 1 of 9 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableDetail/Columns/Columns.tsx
```diff
@@ -16,17 +16,18 @@ type Props = {
 
 export const Columns: FC<Props> = ({ columns }) => {
   const [isClosed, setIsClosed] = useState(false)
-  const handleClose = () => {
+  const handleClose = (event: MouseEvent) => {
+    event.stopPropagation()
```
FunamaYukina, line 20: Since the ``handleClose`` function is also used in the header, I added the above code to prevent double event firing.

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableDetail/Indices/Indices.tsx
```diff
@@ -1,30 +1,56 @@
...
       </div>
       {Object.keys(indices).length !== 0 && (
-        <ul className={styles.list}>
-          {Object.entries(indices).map(([key, index]) => (
-            <li key={key} className={styles.listItem}>
-              {index.name}
-            </li>
-          ))}
-        </ul>
+        <div className={clsx(!isClosed && styles.visible, styles.listWrapper)}>
```
FunamaYukina, line 44: I had to add an extra ``div`` to fix an animation issue caused by the ``padding``.
If we add a class ``visible`` to the ``list`` class, it would look like this:


https://github.com/user-attachments/assets/b7365149-98ea-4578-a6ed-bf3d629b06ff

MH4GF (approved):
It looks good to me, but there are exactly the same codes, so I think it would be better to extract the common component like `CollaspableHeader`.

FunamaYukina:
I'll make a new PR for the shared component. Thanks for reviewing!

PR #617, opened 2025-01-29
💄 Refactor ErrorDisplay for responsive styling

sasamuku:
## Summary

Refactored ErrorDisplay for responsive styling.

|🎨 Design|💻 ParseError|💻 NetworkError|
|--|--|--|
|image|image|image|

🙅‍♂️ The position of error messages has not been as designed yet.

### Next work

The position of error messages will be corrected in the next PR.
It may be necessary to change the structure of the component slightly.

PR #502, opened 2025-01-17
🚸 Updated CLI help text to dynamically display supported formats

hoshinotsuyoshi:
## Summary

- 🚸 Updated CLI help text to dynamically display supported formats https://github.com/liam-hq/liam/pull/502/commits/bd2a4ca71204ed8eb694e19c08547738574f78d3
    - **main part of this PR**
    - Added `prisma`
    - Enhanced test coverage for the `--format` option in `erdCommand`. 
 - fix(cli): improve format validation error message for prisma https://github.com/liam-hq/liam/pull/502/commits/f4b9f73d987c01cafe831aca5f76a6aeee3316d3
    - **Improves error message**
    - Updated `runPreprocess` to provide more detailed error messages when an invalid format is specified.

PR #446, opened 2025-01-14
🚸 Enhanced error rendering in the ErrorDisplay component, adding detailed error summaries

hoshinotsuyoshi:
## Summary

Enhanced error rendering in the `ErrorDisplay` component, adding detailed error summaries.

- Replaced `ProcessError` with `ErrorObject` across components and modules.
- Fixed typos in error messages, replacing "parcing" with "parsing" in test files and logic.

Now 

details-summary

## Testing

You can check the differrence:

* main branch: 
    * https://liam-erd-web.vercel.app/erd/p/raw.githubusercontent.com/metabase/metabase/fe13ca685912aa3cde89a7a7e8c0d4698cf9d060/resources/migrations/initialization/metabase_mysql.sql
* this PR branch:
    * preview: https://liam-erd-2hxkzepwq-route-06-core.vercel.app/erd/p/raw.githubusercontent.com/metabase/metabase/fe13ca685912aa3cde89a7a7e8c0d4698cf9d060/resources/migrations/initialization/metabase_mysql.sql
    * or local: http://localhost:3001/erd/p/raw.githubusercontent.com/metabase/metabase/fe13ca685912aa3cde89a7a7e8c0d4698cf9d060/resources/migrations/initialization/metabase_mysql.sql

PR #619, opened 2025-01-29
💄 Change position of info icon to error display for mobile view

sasamuku:
## Summary

Changed position of InfoIcon on ErrorDisplay.

The placement of the InfoIcon has been implemented in a tricky way.
Please let me know if there is a better way to meet design.

### Before

|🎨 Design|💻 ParseError|💻 NetworkError|
|--|--|--|
|image|image|image|

### After

|🎨 Design|💻 ParseError|💻 NetworkError|
|--|--|--|
|image|image|image|

- https://github.com/liam-hq/liam/pull/617

### DIFF
--- a/.changeset/strange-weeks-change.md
+++ b/.changeset/strange-weeks-change.md
@@ -1,0 +1,6 @@
+---
+"@liam-hq/erd-core": patch
+"@liam-hq/cli": patch
+---
+
+💄 Change position of info icon to error display for mobile view
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ErrorDisplay/ErrorDisplay.module.css
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ErrorDisplay/ErrorDisplay.module.css
@@ -21,11 +21,7 @@
 }
 
 .iconWrapper {
-  display: flex;
-  min-width: 24px;
-  min-height: 24px;
-  padding-top: var(--spacing-1, 4px);
-  align-items: center;
+  display: none;
 }
 
 @media screen and (min-width: 768px) {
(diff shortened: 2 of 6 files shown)


MH4GF:
@sasamuku Am I correct in my understanding that the title and text should be aligned at the start on PC, but not on SP?

PC
Screenshot_2025-01-30_at_11_39_48

SP
Screenshot_2025-01-30_at_11_40_38

You are right, it is complicated. 
It would have been more beneficial not to overcomplicate things here, so it seemed better to check with the our designer about her intentions and change it simple!

hoshinotsuyoshi:
Oh, my bad. 🙇‍♂️  It probably started around #446 or #502, but I just noticed it after seeing [this comment](https://github.com/liam-hq/liam/pull/619#issuecomment-2623396610).

In browsers other than Chrome (e.g., Safari), the title message appears to be using a **serif** font-family.  

- **Main branch:**  https://liambx.com/erd/p/bad-place)](https://liambx.com/erd/p/bad-place
- **Preview:** https://liam-erd-j4d9d6viv-route-06-core.vercel.app/erd/p/bad-place

I believe using a non-serif font is correct. Could you confirm? 🙏  @sasamuku

sasamuku:
@MH4GF 
>Am I correct in my understanding that the title and text should be aligned at the start on PC, but not on SP?

Right, I considered changing a component structure, but it was difficult.

> it seemed better to check with the our designer about her intentions and change it simple!

Sounds nice👍 I'll check with designer.

sasamuku:
@hoshinotsuyoshi 
>I believe using a non-serif font is correct. Could you confirm?

Thanks for letting me know! I'll check it.

sasamuku:
### Finally

`InfoIcon` is hidden on mobile to avoid complexity:

|🎨 Design|💻 ParseError|💻 NetworkError|
|--|--|--|
|image|image|image|

sasamuku:
@MH4GF Could you please re-review?🙏

hoshinotsuyoshi (approved):
LGTM!

PR #620, opened 2025-01-29
💄 Update styles for responsive design and sticky headers in TableDetail components

FunamaYukina:
## Summary

For mobile devices, the ``RelatedTables`` are no longer fixed to the bottom but scroll along with the content.
The header of each section has also been made sticky on scroll.

Since there's a lot of duplicate code, I'll refactor it in the next PR.🙏

| Before  | After |
|  RelatedTables is fixed to the buttom ss 2790 | RelatedTables is not fixed  ss 2791 |
| The section header is scrollable. ss 2793 | The section header is not scrollable. ss 2794 |

### DIFF
--- a/.changeset/quick-llamas-chew.md
+++ b/.changeset/quick-llamas-chew.md
@@ -1,0 +1,6 @@
+---
+"@liam-hq/erd-core": patch
+"@liam-hq/cli": patch
+---
+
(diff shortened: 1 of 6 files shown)


sasamuku (approved):
LGTM👍

PR #615, opened 2025-01-29
💄 Enhance mobile Toolbar animation behavior and improve UI responsiveness

hoshinotsuyoshi:
## Summary

- 💄 Enhance mobile Toolbar animation behavior and improve UI responsiveness
- and some refactors

### Changes

* 738e66db : 💄 Enhance mobile Toolbar animation behavior and improve UI responsiveness **main change**
  - Refactored `expand`, `collapse`, and `fadeInContent` animations for smoother transitions.
  - Introduced `.initialOpen` class to differentiate initial and interactive states.
  - Updated `MobileToolbar.tsx`:
    - Added `useEffect` to reset `hasInteracted` on window resize, improving responsiveness.
    - Adjusted class application logic to distinguish between initial and user-interacted states.
  - Minor cleanup: removed redundant padding values in CSS.
* d0bd6bab : 💄 tweak width
    - fixes wrong setting 🙏 
* ef47ff1f : 💄 fix font-family
    - fixes wrong setting 🙏 
* 73f31eb7 : 💄 fix to use `align-items: center`
    - fixes wrong setting 🙏 
* 97108c82  : ♻️ refactor: combine Toolbar components
    - fixes two `biome-ignore`
    - see also https://github.com/liam-hq/liam/pull/592#discussion_r1933058816

##### before

see example https://liambx.com/erd/p/github.com/mastodon/mastodon/blob/e2f085e2b2cec08dc1f5ae825730c2a3bf62e054/db/schema.rb

##### after

https://github.com/user-attachments/assets/c93c0754-9cd4-4873-b7ef-a1f0df6e08e9

##### before

When viewport width changes, animation fired

https://github.com/user-attachments/assets/6a6102db-ed4d-4014-88aa-3343da5d7afc

##### after

Even when viewport width changes, animation does not be fired

https://github.com/user-attachments/assets/af3977b5-6f85-4fa1-b156-c2a7752fa518

----

- https://liam-erd-sample-g2z50k2vy-route-06-core.vercel.app/?showMode=ALL_FIELDS
- https://liam-erd-qy0kd5nf4-route-06-core.vercel.app/erd/p/raw.githubusercontent.com/mastodon/mastodon/refs/heads/main/db/schema.rb

## Testing

I've checked preview envs.

- https://liam-erd-sample-6aoivhxo9-route-06-core.vercel.app/
- https://liam-erd-jwsewat7s-route-06-core.vercel.app/erd/p/github.com/mastodon/mastodon/blob/main/db/schema.rb

PR #592, opened 2025-01-28
✨️ Add new Mobilie Toolbar

hoshinotsuyoshi:
## Summary

https://github.com/user-attachments/assets/aef8520d-f046-461b-8b4e-a17a4782bcff

- f6bc6908 : ♻️ refactor: Rename Toolbar -> DesktopToolbar
- 0dc25721 : ♻️ refactor: move `<DesktopToolbar />` position
    Repositioned `<DesktopToolbar />` to ensure the upcoming Mobile Toolbar
    can be displayed above the LeftPane in the next commit.
- d7bbb7f9 : ✨️ Add new Mobilie Toolbar
    - Introduced `MobileToolbar.tsx` for improved usability on mobile devices.
    - Applied conditional rendering in `ERDRenderer.tsx` to display either `DesktopToolbar` or `MobileToolbar` based on screen size.
    - Added styles for `MobileToolbar` with animations for expand/collapse behavior.
    - Enhanced `FitviewButton` and `TidyUpButton` to support children as button labels.
    - Modified `IconButton` to accept children for better accessibility.
    - Added `Ellipsis` icon to support toolbar toggling.
    - Defined a new CSS variable `--z-index-toolbar` for consistent layering.

> [!WARNING]
> - `ShowMode` optimization is not yet.
> - some TODO remains.
>    - two biome-ignores
>    - three `!important`
>       - https://github.com/liam-hq/liam/pull/592/files#r1931680348

PR #621, opened 2025-01-29
♻️ Move ERDRenderer/ERDContent/Toolbar -> ERDRenderer/Toolbar

hoshinotsuyoshi:
## Summary

♻️ Move ERDRenderer/ERDContent/Toolbar -> ERDRenderer/Toolbar  

- ref. https://github.com/liam-hq/liam/pull/592#discussion_r1933038207  
   - > It is better to move Toolbar components to the same level as ERDContent 🙏🏻  
- ref. https://github.com/liam-hq/liam/commit/97108c82728cc94081ed5aa5f070b99602a299aa   ( previous PR #615 )
   - This commit only resolves `biome-ignore`  

----

> It is better to move Toolbar components to the same level as ERDContent 🙏🏻  

This PR might align with that comment's intention. What do you think about the changes in this PR?  @MH4GF


frontend/packages/erd-core/src/components/ERDRenderer/ERDRenderer.tsx
```diff
@@ -11,10 +11,10 @@ import { toggleLogEvent } from '@/features/gtm/utils'
 import { useVersion } from '@/providers'
 import { useDBStructureStore, useUserEditingStore } from '@/stores'
 import { CardinalityMarkers } from './CardinalityMarkers'
-import { Toolbar } from './ERDContent'
 import { ErrorDisplay } from './ErrorDisplay'
 import { RelationshipEdgeParticleMarker } from './RelationshipEdgeParticleMarker'
 import { TableDetailDrawer, TableDetailDrawerRoot } from './TableDetailDrawer'
+import { Toolbar } from './Toolbar'
```
hoshinotsuyoshi, line 17: I think this is fine.

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/index.ts
```diff
@@ -1,3 +1,3 @@
 export * from './ERDContent'
 export * from './TableNode'
-export * from './Toolbar'
+export * from './useAutoLayout'
```
hoshinotsuyoshi, line 3: Needed for  `TidyUpButton` .

MH4GF (approved):
LGTM 👍🏻

Issue #333, opened 2024-12-20
It is possible to initiate a meaningless connection


MH4GF:
### Self Checks

- [X] This is only for bug report, if you would like to ask a question, please head to [Discussions](../discussions/categories/q-a).
- [X] I have searched for existing issues [search for existing issues](../issues), including closed ones.

### Cloud or Self Hosted

Self Hosted

### Version (only for Self Hosted)

v0.0.12 + f4c6be3 (2024-12-20)

### Steps to reproduce

https://github.com/user-attachments/assets/8ed62d83-a90f-4763-a0de-e39a46ed5e39

### Expected Behavior

It is not possible to create connection.

### Actual Behavior

It is possible to initiate a meaningless connection.

PR #636, opened 2025-02-01
🐛 remove meaningless connection

ya2s:
## Summary

Rather than changing the isConnectable of each handle, the overall nodesConnectable is changed.
I think it should be changed when individual response is required in the future.
At the same time, the style of the cursor was changed.

CleanShot 2025-02-02 at 00 21 31

## Related Issue

resolve: #333


sasamuku (approved):
LGTM👍

Issue #574, opened 2025-01-27
The type name of a Table Node is not displayed when opened from a URL where active is set as a query parameter.


MH4GF:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](../discussions/categories/q-a).
- [x] I have searched for existing issues [search for existing issues](../issues), including closed ones.

### Cloud or Self Hosted

Cloud

### Steps to reproduce

access to:
https://liambx.com/erd/p/raw.githubusercontent.com/liam-hq/liam/b7b9228de8840eb4ac3ca8342ebbda6d611d7ae5/frontend/packages/db-structure/src/parser/schemarb/input/schema1.in.rb?showMode=ALL_FIELDS&active=users

<img width="1074" alt="Image" src="https://github.com/user-attachments/assets/fff80984-3a98-45ec-8e62-f3fed30dfbfd" />

### Expected Behavior

Display type name of an active table node.

<img width="1074" alt="Image" src="https://github.com/user-attachments/assets/b03e76e0-6155-48bb-adeb-dcfab0562adc" />

### Actual Behavior

Don't display type name of an active table node.

PR #637, opened 2025-02-02
🐛 set the 'data-erd' attribute in the active table to make type names visible

tnyo43:
## Summary

The table type names in a highlighted node are shown [by setting the `data-erd` attribute to `"'table-node-highlighted'"`](https://github.com/liam-hq/liam/blob/18e75a33134cbbfb98ef55e84419b2e7a93dc15a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableColumnList/TableColumn/TableColumn.module.css#L50-L52).
In this PR, I will make it set that data attribute to the node which is active highlighted to display the table type names as well.

## Related Issue

resolve: https://github.com/liam-hq/liam/issues/574

## Changes

- set `data-erd="table-node-highlighted"` to active highlighted nodes

## Testing

1. run `pnpm run dev`
2. go to http://localhost:3001/erd/p/raw.githubusercontent.com/liam-hq/liam/b7b9228de8840eb4ac3ca8342ebbda6d611d7ae5/frontend/packages/db-structure/src/parser/schemarb/input/schema1.in.rb?showMode=ALL_FIELDS&active=users
3. table field type names of "users" are displayed in the first render

https://github.com/user-attachments/assets/1ed1ae40-8efd-45e2-9657-a1a5fffbb3fb

## Other Information

I previously tried [another way to solve this bug](https://github.com/liam-hq/liam/pull/637/commits/241907fb1776a643d84bf718fadc15a299c9e152) once. However I reverted it due to [the reason I mentioned in the commit message](https://github.com/liam-hq/liam/pull/637/commits/4ccea838127a1f3bfbc1628d9602ee0afa566d39).
I'm very new to this project and, unfotrunately, not familiar with the domain knowledge. I appreciate if you check that approach as well and let me know if you prefer it 😉

### DIFF
--- a/.changeset/silly-impalas-add.md
+++ b/.changeset/silly-impalas-add.md
@@ -1,0 +1,6 @@
+---
+"@liam-hq/erd-core": patch
+"@liam-hq/cli": patch
+---
+
+🐛 modify function to treat active highlighted nodes as also highlighted nodes
--- a/.changeset/silly-impalas-add.md
+++ b/.changeset/silly-impalas-add.md
@@ -1,6 +1,0 @@
----
-"@liam-hq/erd-core": patch
-"@liam-hq/cli": patch
----
-
-🐛 modify function to treat active highlighted nodes as also highlighted nodes
--- a/.changeset/slimy-terms-jump.md
+++ b/.changeset/slimy-terms-jump.md
@@ -1,0 +1,6 @@
+---
+"@liam-hq/erd-core": patch
+"@liam-hq/cli": patch
+---
+
+🐛 Fixed the 'data-erd' attribute in the active table to make type names visible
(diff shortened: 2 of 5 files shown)


MH4GF (approved):
LGTM 🚀 Thanks for fixing it! 😄

MH4GF:
@tnyo43 

> I previously tried [another way to solve this bug](https://github.com/liam-hq/liam/pull/637/commits/241907fb1776a643d84bf718fadc15a299c9e152) once. However I reverted it due to [the reason I mentioned in the commit message](https://github.com/liam-hq/liam/pull/637/commits/4ccea838127a1f3bfbc1628d9602ee0afa566d39).

Your understanding is correct, isActiveHighlighted and isHighlighted are exclusive. I would like to refactor it so that both states do not exist, but that can be done another time 👍🏻

PR #645, opened 2025-02-03
Add playwright tests

FunamaYukina:
## Summary

I've integrated Playwright for our end-to-end testing needs. 
You can run the tests using the ``pnpm test:e2e`` command.(Only on ``erd-core`` and ``cli`` packages)

Currently, our test suite only verifies the visibility of the title. I'll be adding more tests in a subsequent PR to cover other functionalities.


.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,51 @@
+name: E2E Tests
+on:
+  workflow_run:
+    workflows:
+      - "Vercel Deployment"
+      - "Vercel Deployment erd-sample"
+    types:
+      - completed
+
```
FunamaYukina, line 9: This E2E test GitHub Action will start after the completion of the following workflows: ``Vercel Deployment erd-sample`` and ``Vercel Deployment``. The reason is that we need to access the deployed URL to perform the E2E tests.

This condition cannot be tested on this branch because it requires a merge to the main branch. However, I've confirmed that the test passes with a fixed URL.

with a fixed URL pattern: https://github.com/liam-hq/liam/blob/daeaf24255a64729b6a5d8520338e8e79f263282/.github/workflows/e2e-test.yml#L20

GitHub action log:
https://github.com/liam-hq/liam/actions/runs/13127470815/job/36626423092

.github/workflows/_run_e2e_test.yml
```diff
@@ -0,0 +1,45 @@
...
+        uses: actions/checkout@v4
+
+      - name: Setup pnpm
+        uses: ./.github/actions/pnpm-setup
+
+      - name: Configure Git safe directory
+        run: git config --global --add safe.directory /__w/liam/liam
+
+      - name: Install Playwright browsers
+        run: pnpm exec playwright install --with-deps
```
MH4GF, line 32: This might be unnessesary because it uses docker image of playwright 💭

frontend/packages/erd-core/package.json
```diff
@@ -19,6 +19,8 @@
     "@biomejs/biome": "1.9.3",
     "@liam-hq/configs": "workspace:*",
     "@liam-hq/db-structure": "workspace:*",
+    "@playwright/test": "1.50.0",
+    "@types/node": "22.9.0",
```
FunamaYukina, line 23: I've added "@types/node" to page.test.ts to use environment variables.

FunamaYukina:
Due to the excessive number of unnecessary commits, I'll rebase this PR to provide a more focused change set🙏. Thank you for taking the time to review.

PR #650, opened 2025-02-03
🎨 Fix sticky header z-index for mobile layout

MH4GF:


hoshinotsuyoshi:
just a note 📝 

 It seems that issues like this are easier to spot on devices with smaller screen heights. I confirmed that on desktop Chrome, selecting the iPhone SE preset (375x667) and then rotating the screen reproduces the issue.

hoshinotsuyoshi (approved):
LGTM!

PR #651, opened 2025-02-03
✨ Add close functionality for mobile toolbar when clicking outside

sasamuku:
## Summary

https://github.com/user-attachments/assets/8ef60a23-539e-42e4-8e2b-e3f68c55fb66

### DIFF
--- a/docs/packages-license.md
+++ b/docs/packages-license.md
@@ -2,7 +2,7 @@
 
 
 ## Summary
-* 1010 MIT
+* 1011 MIT
 * 77 Apache 2.0
 * 64 ISC
 * 27 Simplified BSD
@@ -1899,6 +1899,17 @@ LGPL-3.0-or-later permitted
 
 <a name="@radix-ui/react-primitive"></a>
 ### @radix-ui/react-primitive v2.0.0
+#### 
+
+##### Paths
+* /home/runner/work/liam/liam
+
+<a href="http://opensource.org/licenses/mit-license">MIT</a> permitted
+
+
+
+<a name="@radix-ui/react-radio-group"></a>
+### @radix-ui/react-radio-group v1.2.2
 #### 
 
 ##### Paths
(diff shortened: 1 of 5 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/Toolbar/MobileToolbar/MobileToolbar.tsx
```diff
@@ -19,53 +20,85 @@ export const MobileToolbar: FC = () => {
...
+      if (
+        toolbarRef.current &&
+        !toolbarRef.current.contains(event.target) &&
+        isOpen
+      ) {
+        event.preventDefault()
+        event.stopPropagation()
+
+        setIsOpen(false)
+        setIsShowModeMenu(false)
```
hoshinotsuyoshi, line 38: That sounds good.

Also,

```suggestion
```

^ This might be a reasonable approach as well. What do you think?

The close button function, `toggleOpenClose`, currently only calls `setIsOpen(false)`.

hoshinotsuyoshi (approved):
> close functionality for mobile toolbar when clicking outside

Thank you!!  Looks good to me. A small comment left.

frontend/packages/erd-core/src/components/ERDRenderer/Toolbar/MobileToolbar/MobileToolbar.tsx
sasamuku: Thanks for your comment.

>The close button function, toggleOpenClose, currently only calls setIsOpen(false).

It would be better to chose the same way as the close button.

docs/packages-license.md
```diff
@@ -1908,6 +1908,17 @@ LGPL-3.0-or-later permitted
...
+### @radix-ui/react-radio-group v1.2.2
+#### 
+
+##### Paths
+* /home/runner/work/liam/liam
+
+<a href="http://opensource.org/licenses/mit-license">MIT</a> permitted
+
+
+
```
hoshinotsuyoshi, line 1921: Oops, my mistake! I should’ve included this diff in my PR. 🙇 (All good in the end, though!)

PR #657, opened 2025-02-04
🎨 Make SidebarFooter appear after scrolling when table list overflows

hoshinotsuyoshi:
## Summary

🎨 Make SidebarFooter appear after scrolling when table list overflows 

Refactored Sidebar layout to allow the table list to occupy more space initially. The entire sidebar is now scrollable, and SidebarFooter is displayed only after scrolling.

before:

https://github.com/user-attachments/assets/ef4aed8b-adc0-46a9-a5fe-ac93dc22d525

after:

https://github.com/user-attachments/assets/4accbb26-2939-4a47-9229-95f834b7e88e

### DIFF
--- a/.changeset/twenty-pigs-accept.md
+++ b/.changeset/twenty-pigs-accept.md
@@ -1,0 +1,6 @@
+---
+"@liam-hq/erd-core": patch
+"@liam-hq/cli": patch
+---
+
+🎨 Make SidebarFooter appear after scrolling when table list overflows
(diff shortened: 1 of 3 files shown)


MH4GF:
@hoshinotsuyoshi 
In this approach, the table count in the PC layout is hidden too. I would like to sticky that one as it is now..
https://github.com/user-attachments/assets/9588f7dc-fcc5-472d-9013-fe674826bd68

This problem is probably caused by the contents of mobile layout being in the SidebarFooter. I think that it can be solved by moving to the SidebarContent.

hoshinotsuyoshi:
@MH4GF  Thank you! how about this? https://github.com/liam-hq/liam/pull/657/commits/ec8d04458a44e0913764acc0f2e73e3540dc4a26


https://github.com/user-attachments/assets/c8280276-a7ef-425b-b2f2-b82fddc07613

MH4GF (approved):
LGTM 👍🏻 Thank you for correcting!

PR #628, opened 2025-01-30
💄 Adjust UI of LeftPane on mobile widths

junkisai:
## Summary

The information that was displayed on the AppBar in mobile width has been adjusted to be displayed on the LeftPane side.

## Related Issue

N/A

## Changes

N/A

## Testing

**@liam-hq/cli**

| mobile | desktop |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/052de265-f206-424f-8899-db6b589b72cc" /> | スクリーンショット 2025-01-31 11 34 48 | 

**@liam-hq/erd-web**

| mobile | desktop |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/412733df-c6cf-4d9b-b436-ce04a854e3b0" /> | スクリーンショット 2025-01-31 11 34 27 | 

## Other Information

N/A

PR #659, opened 2025-02-04
🎨 Enhance MenuItemLink with improved external link handling

junkisai:
## Summary

I've made adjustments to prevent the 'noreferrer' attribute from being added when linking to the same domain, while taking into account both the CLI and web versions.

ref. https://github.com/liam-hq/liam/pull/628#discussion_r1940525096

## Testing

| cli ver (noreferrer) | web ver (without noreferrer) |
|--------|--------|
| スクリーンショット 2025-02-04 15 04 45 | スクリーンショット 2025-02-04 15 05 08 |

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/LeftPane/LeftPane.tsx
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/LeftPane/LeftPane.tsx
@@ -24,41 +24,49 @@ import { MenuItemLink, type Props as MenuItemLinkProps } from './MenuItemLink'
 import { TableCounter } from './TableCounter'
 import { TableNameMenuButton } from './TableNameMenuButton'
 
-const menuItemLinks: MenuItemLinkProps[] = [
-  {
-    label: 'Release Notes',
-    href: 'https://github.com/liam-hq/liam/releases',
-    isExternalLink: true,
-    icon: <Megaphone className={styles.icon} />,
-  },
-  {
-    label: 'Documentation',
-    href: 'https://liambx.com/docs',
-    isExternalLink: true,
-    icon: <BookText className={styles.icon} />,
-  },
-  {
-    label: 'Community Forum',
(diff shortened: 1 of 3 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/LeftPane/MenuItemLink/MenuItemLink.tsx
```diff
@@ -5,14 +5,16 @@ import styles from './MenuItemLink.module.css'
 export type Props = {
   label: string
   href: string
-  isExternalLink?: boolean
+  target: '_self' | '_blank'
```
hoshinotsuyoshi, line 8: 📝 It seems unlikely that the argument will be called with `_self` for now, but both the intention and implementation look solid, I think. 👍

frontend/packages/erd-core/src/components/ERDRenderer/LeftPane/LeftPane.tsx
```diff
@@ -24,42 +24,50 @@ import { MenuItemLink, type Props as MenuItemLinkProps } from './MenuItemLink'
...
+        label: 'Community Forum',
+        href: 'https://github.com/liam-hq/liam/discussions',
+        noreferrer: true,
+        target: '_blank',
+        icon: <MessagesSquare className={styles.icon} />,
+      },
+      {
+        label: 'Go to Homepage',
+        href: 'https://liambx.com/',
+        noreferrer: version.displayedOn === 'cli',
```
hoshinotsuyoshi, line 56: > In the web version, the transition is within the same domain, so I want to be able to track the referral source in GA4 without adding `noreferrer`.
Therefore, I thought it would be a good idea to add props for whether to open in a separate tab and whether it is the same domain.

thanks!

hoshinotsuyoshi (approved):
LGTM!

PR #661, opened 2025-02-04
🐛 Refactor ERD renderer toolbar positioning and visibility

sasamuku:
## Summary

Refactor to change position of Toolbar along with Sidebar.

### DesktopToolbar

https://github.com/user-attachments/assets/9d722862-88db-4c1b-bbd6-97a9370f09fe

### MobileToolbar

https://github.com/user-attachments/assets/1b2a185a-9c14-4dad-990f-a269fe698bf8

### DIFF
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDRenderer.module.css
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDRenderer.module.css
@@ -33,7 +33,7 @@
 .toolbarWrapper {
   position: absolute;
   left: 0;
-  bottom: 2rem;
+  bottom: 0;
   right: 0;
   display: grid;
   justify-content: center;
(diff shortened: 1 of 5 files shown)


frontend/packages/erd-core/src/components/ERDRenderer/ERDRenderer.module.css
```diff
@@ -46,19 +46,8 @@
...
   pointer-events: auto;
 }
-@media screen and (max-width: 768px) {
-  .toolbarWrapper {
-    transition: opacity 0.6s;
-    opacity: 1;
-    visibility: visible;
-  }
 
-  :global([data-state='expanded']) + .main .toolbarWrapper {
```
sasamuku, line 56: 📝 This is added in https://github.com/liam-hq/liam/pull/359, but this will be unnecessary because the Toolbar is now displayed according to the screen width.

MH4GF (approved):
LGTM 👍🏻

PR #665, opened 2025-02-04
🐛 Adjust the item height calculation in the Columns component.

FunamaYukina:
## Summary

ss 2806


frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableDetail/Columns/Columns.tsx
```diff
@@ -10,8 +10,8 @@ type Props = {
 }
 
 export const Columns: FC<Props> = ({ columns }) => {
-  // NOTE: 131.5px is the height of one item in the list
-  const contentMaxHeight = Object.keys(columns).length * 131.5
+  // NOTE: 300px is the height of one item in the list(when comments are lengthy)
+  const contentMaxHeight = Object.keys(columns).length * 300
```
sasamuku, line 14: Q: I couldn't understand why the height is 300 👀

FunamaYukina: Ah, sorry for the lack of explanation.🙏
 When testing with a large number of column comments, the height was around ``222px``, so I set it to ``300px`` to allow some extra space. Of course, since it's a max-height, ``400px`` or ``500px`` would also be fine.
ss 2806

sasamuku: Thank you for your explanation👍

sasamuku (approved):
LGTM👍

PR #667, opened 2025-02-04
🔧 Fix E2E test workflow

FunamaYukina:
## Summary

I found that using ``workflow_run`` only fires when PR is merged into main.🙏
 Therefore, I modified it to use ``workflow_call`` so that the e2e test is run when the vercel deploy completes.

## Testing

I made some experimental changes to the ``erd-core`` file in the https://github.com/liam-hq/liam/pull/667/commits/ef70608b6f13e58488945aab9978e6a1a69d249d commit and verified that the e2e test runs and passes after running the vercel workflow as shown below.

test code(to run vercel workflow)
frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/usePopStateListener.ts
```
    // test
    setTimeout(() => {}, 10)
```

✅ https://github.com/liam-hq/liam/actions/runs/13147938960/job/36689953415?pr=667
✅ https://github.com/liam-hq/liam/actions/runs/13147938960/job/36690140634?pr=667


.github/workflows/vercel-deploy-erd.yml
```diff
@@ -86,3 +86,12 @@ jobs:
...
         id: deployment
+  run-e2e:
+    needs: deploy
+    if: needs.deploy.result == 'success'
+    uses: ./.github/workflows/e2e-test.yml
+    with:
+      url: ${{ needs.deploy.outputs.url }}
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}-e2e
+      cancel-in-progress: true
```
FunamaYukina, line 97: Concurrency is added to avoid deadlock.
ref: https://github.com/liam-hq/liam/actions/runs/13135929311/job/36651312580

MH4GF (approved):
I think it's a good idea, but there doesn't seem to be any reason to depend on deploy.yml -> e2e-test.yml -> _run_e2e_test.yml.

deploy.yml → _run_e2e_test.yml sounds good, what do you think?

hoshinotsuyoshi:
@FunamaYukina 

Just to clarify, I have a couple of questions!

- Is this the correct representation of the call graph?
- As of now, is **vercel-deploy-erd.yml** the only file that uses **e2e-test.yml**?

```mermaid
graph TD
    vercel-deploy-erd.yml --> e2e-test.yml
    e2e-test.yml --> _run_e2e_test.yml
```

-----

similar to https://github.com/liam-hq/liam/pull/667#pullrequestreview-2594455163 , this is not blocking the merge!

FunamaYukina:
Thank you for pointing that out!🙏 

```
vercel-deploy-erd.yml -> _run_e2e_test.yml 
````
This flow looks good.🙇‍♀️ I'm going to fix it now.

FunamaYukina:
> As of now, is vercel-deploy-erd.yml the only file that uses e2e-test.yml?

yes!!

FunamaYukina:
📝 https://github.com/liam-hq/liam/pull/667/commits/4f90234b7081d7c25b66bd014efe29aa941f3f0a
test commits to run vercel deploy ( I will modify this commit.)

FunamaYukina:
Thank you for the review. 🙏 I will merge this now!

PR #674, opened 2025-02-05
✅ Add playwright config file

FunamaYukina:
## Summary

I forgot to add ``playwright.config.ts``, so I added it.🙏
Also, I forgot to add ``ts.config.json``, so I added it.🙏🙏

### DIFF
--- a/frontend/packages/e2e/tsconfig.json
+++ b/frontend/packages/e2e/tsconfig.json
@@ -1,0 +1,12 @@
+{
+  "compilerOptions": {
+    "baseUrl": ".",
+    "declaration": true,
+    "esModuleInterop": true,
+    "module": "ESNext",
+    "target": "ES2022",
+    "moduleResolution": "node"
+  },
+  "include": ["tests/**/*"],
+  "exclude": ["node_modules"]
+}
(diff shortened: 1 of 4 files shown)


FunamaYukina:
I'll make a few corrections.🙏

frontend/packages/e2e/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
```
MH4GF, line 1: I want you to use base config as you would any other package.

https://github.com/liam-hq/liam/blob/350fc44d06a264a5359c842fecc40ab3b73b8806/frontend/packages/erd-core/tsconfig.json#L2

This makes the type test more rigorous and convenient.

MH4GF (approved):
LGTM! I wrote a comment.

frontend/packages/e2e/tsconfig.json
FunamaYukina: Sorry for the omission! Corrected.
https://github.com/liam-hq/liam/pull/674/commits/00a1b0a8f5348c4c4b66f9c0c5161bd86653f443

FunamaYukina:
Thank you for the review!

PR #675, opened 2025-02-05
Add `Use fuse.js` ADR

hoshinotsuyoshi:
## Summary

Added `Use fuse.js` ADR


MH4GF (approved):
It's a great ADR that shows that “decision-making is a trade-off.”

sasamuku (approved):
Easy to read and it's convincing.
LGTM👍

Issue #641, opened 2025-02-03
A green dot may appear in the upper left corner


MH4GF:
## 🎯 Background and Objective

- The issue often reproduces when opening from Related Tables in a filtered state.  
- [Reproduction Link](https://liam-erd-sample.vercel.app/?active=follows&hidden=eJx9VOHS3CAIfCHfiSHK5eipWNBkvrfvqLlLcu30TyKrwgKL6L20XAEjo5E5PGwveSM1rCz5RANF6ggo_W5k9bIjCTnDEsW_TjTxql8ukgSaGGSpl4h3q_DljlKcXp5cwGgjxewJaKN8oWAVv6xmZOAjYW4FikT2fAmxo2bOKxQlo_oX_gEMKnYrJM6AflCP0pGcpWVPiXpebdC_Qkrz9B0eVsMISkX0QrSQwln-Nk6WQhjNLehfrZg7yruIvBJqX7X4Ak7Dkcp-A8x5zJLZYwRKyPHdnWtvD973dvtmVRJQkl8MHiutor1yV_xjPThWUnjRzy4avuF3dnfY3KEXjLGzvqtnkr1jD9ykKXeuD8LalMJsy0O6C1DykhLlcM3q33vWSm-5jVQ_Rw49r5S7PCnArUvmOFCuXHsZ3uXlvA1CXN4sI3cfUdZDKdu80HvVk04UGAFrRf-cUujfwWMSzlL5wX7yLKSJD5r3jbeSb-gnhStqTrDVZ9cVmcGq2KPesCovOs-VEj9Xy_okFbCCno7pKhIjbGNS-9JcUdqYdvCoAYrKxqFneoOrUg537Ms6ZfL_Ue-7P_03BD9fjHeDtEUyZ_5JoUUKp0ubvZ4NOVIzqnXM-AhB4fbEmDOuBK1EwWDO5FF31A4E7BGnZ6DA9WOM9-pYz1IdxpH8yeY97kO9FVeYEpzr4_TcI00G8gAj3diTuSppsSqZzLWMG3LEJdIxKeaakcIU5amGAar02vSluZ0WKM2eYG0xr1xmRTr8qcpOS5dDBq80dN8foZ2Wp8jL_gCzgoOG)  
- The issue disappears after a certain period or when moving the cursor.

![Image](https://github.com/user-attachments/assets/0c96f9ad-80a5-4399-ba59-cd035c65aecc)

## 📝 Summary  

[...]

PR #677, opened 2025-02-05
🐛 Removed a green dot on the left top when it rendered

tnyo43:
## Summary

Removed a green dot on the left top when it rendered.
The green dot on the let top is an group of `ellipse` objects who are specified positive values of `begin` attributes.

The `animateMotion` element will position the object at its original coordinates (`cx`, `cy`) before the time specified in the begin attribute has elapsed. If a negative value is specified to `begin` attribute, the object is considered to have started moving before it is rendered.
I created a minimal reproduction how positive/negative `begin` values works. I hope this will help your understanding.
https://codepen.io/tnyo43/pen/vEBqNxw

## Related Issue

resolve: https://github.com/liam-hq/liam/issues/641

## Changes

- specify non-positive number to `begin` attribute of `animateMotion` of `ellipse` object to make them move from the beginning.

## Testing

1. run `pnpm run dev`
2. go to http://localhost:3001/erd/p/raw.githubusercontent.com/liam-hq/liam/b7b9228de8840eb4ac3ca8342ebbda6d611d7ae5/frontend/packages/db-structure/src/parser/schemarb/input/schema1.in.rb?showMode=ALL_FIELDS&active=users
3. check the green dot is not rendered

## Other Information

Unfortunately, it looks that this change won't resolve this issue completely. In my laptop (macOS 15.2, M1 Pro, 32GB), with the today's latest Chrome (version 132.0.6834.160), it shows a green dot in a very small period of time. It doesn't happen in Firefox (version 135.0 (aarch64)) tho.
It's obvious especially in rigt bottom panel.

A green dot in Chrome

https://github.com/user-attachments/assets/bcacabb9-f530-423d-b85b-ff749f9cb17b

No green dot in Firefox

https://github.com/user-attachments/assets/67aa00d2-4e06-498b-bfff-df3338815b17

I believe this is a performance issue because it's rendering a large number of ellipse object (six times more than the number of edges). We can take other actions to solve this problem if you would like. However, this change resolve the core issue and the most meaningful step forward.

---
I really love your idea to implement that particles on edges and make it look be connected with directions. I hope this change will make a bit better decoration with it 😉

### DIFF
--- a/.changeset/fast-pugs-change.md
+++ b/.changeset/fast-pugs-change.md
@@ -1,0 +1,6 @@
+---
+"@liam-hq/erd-core": patch
+"@liam-hq/cli": patch
+---
+
+🐛 Removed a green dot on the left top when it rendered
--- a/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/RelationshipEdge/RelationshipEdge.tsx
+++ b/frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/RelationshipEdge/RelationshipEdge.tsx
@@ -59,7 +59,7 @@ export const RelationshipEdge: FC<Props> = ({
             fill="url(#myGradient)"
           >
             <animateMotion
-              begin={`${i * (ANIMATE_DURATION / PARTICLE_COUNT)}s`}
+              begin={`${-i * (ANIMATE_DURATION / PARTICLE_COUNT)}s`}
               dur={`${ANIMATE_DURATION}s`}
               repeatCount="indefinite"
               rotate="auto"


MH4GF (approved):
Thanks also codepen! That was easy to understand. 
Looks like it will probably be no problem. 👍🏻 

@junkisai Please review too!

junkisai (approved):
I couldn't figure out the cause, so this was very educational!
Thank you very much 🙌 

> I believe this is a performance issue because it's rendering a large number of ellipse object (six times more than the number of edges). We can take other actions to solve this problem if you would like. However, this change resolve the core issue and the most meaningful step forward.

I was adjusting the number of particles, trying to find the maximum number that wouldn't compromise the quality of the edge animations. 
Currently, I have the particle count set to 6, but if the performance issues you pointed out are occurring, I'm willing to consider reducing the number of particles.

I hope we can discuss the number of particles in another issue or pull request! 😄

PR #691, opened 2025-02-06
🔧 Update E2E test workflow to handle multiple app deployments

FunamaYukina:
…rove failure handling


FunamaYukina:
I will try for another PR!🙏

PR #708, opened 2025-02-10
Feat/fit view related table

junkisai:


junkisai:
Differences were detected in the VRT, but it appears that the 50ms delay in the initial load completion in this implementation is the cause, so we decided to update the snapshot.

[✅ update VRT snapshot](https://github.com/liam-hq/liam/pull/708/commits/0a4d6af3fda665232dfd9b92ae6c5d56a7445b6a)

frontend/packages/erd-core/src/components/ERDRenderer/ERDContent/TableNode/TableDetail/RelatedTables/RelatedTables.tsx
MH4GF, line 1: 📝 RelatedTables component became rendering only.

frontend/packages/erd-core/src/features/reactflow/hooks/useCustomReactflow.ts
```diff
@@ -0,0 +1,28 @@
+import { type FitViewOptions, useReactFlow } from '@xyflow/react'
+import { useCallback } from 'react'
+import { MAX_ZOOM, MIN_ZOOM } from '../constants'
+
+export const useCustomReactflow = () => {
```
MH4GF, line 5: IMO: I thought about wrapping useReactFlow completely and using only useCustomReactFlow, but this is trivial so I guess we don't have to do it.

MH4GF (approved):
LGTM 👍🏻 What an amazing!

frontend/packages/erd-core/src/features/reactflow/hooks/useCustomReactflow.ts
junkisai: @MH4GF 
Since we can add a lint rule in Biome to enforce its usage, I thought that the idea of completely wrapping useReactFlow was excellent!
Since it's a good opportunity, I'm considering either creating a "Good First Issue" or having Devin AI handle it.

MH4GF: > Since it's a good opportunity, I'm considering either creating a "Good First Issue" or having Devin AI handle it.

Looks great 👍🏻👍🏻👍🏻

PR #634, opened 2025-01-31
🐛  Fixed forever loading of related tables section for table without relationships

ya2s:
## Summary

The table without the relationShip has a node length of 1.
As a result, tablenodesinitialized was always false
CleanShot 2025-01-31 at 23 50 20@2x

CleanShot 2025-01-31 at 23 50 30@2x

## Related Issue

resolve: #347

PR #736, opened 2025-02-14
🐛 Display Related Tables for Nodes without Relations

junkisai:
There was an issue where, when selecting a table without any relations, the RelatedTables area remained in a loading state and was not displayed. Additionally, pressing the "Open in Main area" button resulted in nothing being shown.

To resolve these issues, we modified the logic that takes into account the parent-child relationships of nodes, so that everything now displays as intended.

#### Related Issue

N/A

#### Testing

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/e1e3aaae-6aaa-4da1-a88f-829e7edf0b3c" /> | <video src="https://github.com/user-attachments/assets/af370965-2ed3-4ddd-b786-4067365ddc29" /> |

### DIFF
--- a/.changeset/gold-birds-occur.md
+++ b/.changeset/gold-birds-occur.md
@@ -1,0 +1,6 @@
+---
+"@liam-hq/erd-core": patch
+"@liam-hq/cli": patch
+---
+
+🐛 When nodes without a relation are present, set the hidden property to false for the non related group node.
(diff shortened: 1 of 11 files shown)


junkisai:
Since I was working on the same issue as [this PR](https://github.com/liam-hq/liam/pull/634), I'll wait for it to be merged before resuming work.

PR #811, opened 2025-03-05
Add fixed width of 80% to text span on hover for better truncation

prakha:
## Issue

- resolve:

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -176,58 +176,6 @@ importers:
         specifier: '5'
         version: 5.6.2
 
-  frontend/apps/migration-web:
-    dependencies:
-      '@langchain/community':
-        specifier: 0.3.33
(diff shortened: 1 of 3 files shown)


pnpm-lock.yaml
MH4GF, line 1: Changes to this file do not seem to be expected. Could you check it?

MH4GF:
@prakha Thanks for always putting out PRs! You are the best! 😄 

---

Looks like it should always be 80%, not just on hover. Right now the tooltip moves slightly because of the size change.

https://github.com/user-attachments/assets/9064b10d-907f-4ffb-9d81-e9601df12cc7

Also, it seems that when the size is changed and truncated, no tooltip appears.

https://github.com/user-attachments/assets/a35b12d6-3bd5-48c9-ae08-a599a32a0559

MH4GF (changes requested):
Thanks!!

Issue #683, opened 2025-02-06
Enhance the Show Mode menu to improve keyboard accessibility


MH4GF:
## Task description
When the Show Mode menu is opened, the button for the current mode should have focus and be able to move up and down.

<img width="461" alt="Image" src="https://github.com/user-attachments/assets/3a2e4d19-25df-4cac-9156-0f7092408640" />

## Background
Currently, when opening the **Show Mode** menu, the focus is not set to the active mode, making it difficult for keyboard users to navigate efficiently. This enhancement will improve accessibility and usability, ensuring a seamless experience for users who rely on keyboard navigation.

## TODO
- [ ] Set focus on the currently selected mode when the menu is opened.
- [ ] Enable navigation between menu items using the up/down arrow keys.
- [ ] Ensure the Enter key selects to change mode.

chetan102:
can i work on this ?

MH4GF:
@chetan102 Of course🚀

PR #842, opened 2025-03-10
✨ Improve DropdownMenu keyboard accessibility

tnyo43:
## Issue

- resolve: https://github.com/liam-hq/liam/issues/683

## Why is this change needed?

The `DropdownMenu` is actually controllable by keyboard but it doesn't show any indicator and hard to use. This change will add styles to make uses be able to control the component by keyboards.

## Testing Verification

1. run `pnpm run dev --filter @liam-hq/erd-web` and visit "localhost:3001"
2. focus Show mode menu button by pressing the Tab key, and press Space to open.
3. press Arrow Up/Down to focus on items and press either Enter or Space to select it.

https://github.com/user-attachments/assets/589acc24-48a4-4f96-b99e-fc79040caa74

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough

## Additional Notes

In the issue, it requires "Set focus on the currently selected mode when the menu is opened." but this change doesn't satisfied it, it focus on the first element always. However, it looks this is the required specification of `menuitemradio` role. 

> When a menu opens, or when a menubar receives focus, keyboard focus is placed on the first item.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/menuitemradio_role#keyboard_interactions

radix-ui, which is the base component of `DropdownMenu`, also follows this specification, so I just used it.

### DIFF
--- a/.changeset/mean-houses-film.md
+++ b/.changeset/mean-houses-film.md
@@ -1,0 +1,5 @@
+---
+"@liam-hq/ui": patch
+---
+
+✨ Improve DropdownMenu keyboard accessibility
--- a/frontend/packages/ui/src/components/DropdownMenu/DropdownMenu.module.css
+++ b/frontend/packages/ui/src/components/DropdownMenu/DropdownMenu.module.css
@@ -57,7 +57,8 @@
   color: var(--primary-accent);
 }
 
-.item:hover {
+.item:hover,
+.item:focus {
   background-color: var(--dropdown-background-hover);
 }
 


MH4GF:
> The DropdownMenu is actually controllable by keyboard but it doesn't show any indicator and hard to use.

Wow!! I didn't know😳

> In the issue, it requires "Set focus on the currently selected mode when the menu is opened." but this change doesn't satisfied it, it focus on the first element always. However, it looks this is the required specification of menuitemradio role.

Umm, I see! That sounds good 👍🏻

PR #866, opened 2025-03-12
✨ implement processGenerateReview function on jobs

FunamaYukina:
## Issue

- resolve:

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough


frontend/apps/app/src/functions/processGenerateReview.ts
```diff
@@ -0,0 +1,50 @@
...
+      temperature: 0.7,
+      model: 'gpt-4o-mini',
+    })
+
+    const chain = prompt.pipe(model)
+    const response = await chain.invoke({
+      schemaChanges: payload.schemaChanges,
+    })
+
+    const review: Review = JSON.parse(response.content.toString())
```
MH4GF, line 44: There must be cases where the LLM response does not return JSON.
Can you implement structured output later?

ref: [Structured outputs | 🦜️🔗 Langchain](https://js.langchain.com/docs/concepts/structured_outputs/)

MH4GF (approved):
LGTM 👍🏻

frontend/apps/app/src/functions/processGenerateReview.ts
FunamaYukina: Oh, yes! I will try to make it chain as follows.
```
chain = prompt | model | output_parser
```

PR #961, opened 2025-03-21
🔧 [PoC] Add schema json generation script

sasamuku:
## Issue

- resolve:

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough


frontend/packages/db/scripts/generate-schema.sh
```diff
@@ -0,0 +1,213 @@
...
+echo "Database: $DB_NAME"
+
+# Get the directory where the script is located
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SQL_FILE="$SCRIPT_DIR/generate-schema.sql"
+
+# Run the SQL query from file
+PGPASSWORD=$DB_PASSWORD psql -h "$DB_HOST" -p "$DB_PORT" -U "$DB_USER" -d "$DB_NAME" -c "
+/* Supabase edition */
+WITH fk_info_supabase AS (
```
sasamuku, line 24: 📝
This is actually ChartDB's SQL.
https://github.com/chartdb/chartdb

sasamuku: Need to fix this for Liam's SQL

MH4GF: I understood. This approach seems to use in cli 👍🏻

PR #1179, opened 2025-04-07
PoC: support implicit many-to-many relations

MH4GF:


frontend/packages/db-structure/src/parser/prisma/parser.ts
```diff
@@ -229,4 +273,143 @@ function normalizeConstraintName(constraint: string): ForeignKeyConstraint {
...
+/**
+ * Creates a join table for a many-to-many relationship
+ */
+function createManyToManyJoinTable(joinTableName: string): Table {
+  return {
+    name: joinTableName,
+    columns: {
+      A: {
+        name: 'A',
+        type: 'serial',
```
MH4GF, line 292: There are some specifications that are not supported by this PoC, including `type`. Please investigate and implement them as well.

https://www.prisma.io/docs/orm/prisma-schema/data-model/relations/many-to-many-relations#rules-for-defining-an-implicit-m-n-relation

PR #1416, opened 2025-04-21
✨(dev): Add MCP server for UI component discovery

MH4GF:
## Issue

- resolve: Add MCP server for UI component discovery

## Why is this change needed?

This PR adds a Model Context Protocol (MCP) server package that enables Cursor IDE to interact with Liam's UI components. The server provides tools for listing all UI components and retrieving component file contents, enhancing developer productivity when working with our component library.

- Security implementation for path handling in component file access
- Error handling approach
- MCP tool interface design

## Testing Verification

Tested by:

- Setting up the MCP server in Cursor IDE
- Verifying component listing works correctly
- Retrieving component files through the MCP tools

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough

## Additional Notes

The `.cursor/mcp.json` file is excluded from Git as it contains machine-specific paths. The README includes instructions for developers to create their own configuration.

### DIFF
--- a/.cursor/mcp.json
+++ b/.cursor/mcp.json
@@ -1,0 +1,11 @@
+{
+  "mcpServers": {
+    "liam-development-mcp-server": {
+      "command": "/Users/mh4gf/.asdf/shims/node",
+      "args": [
+        "--experimental-strip-types",
+        "/Users/mh4gf/ghq/github.com/liam-hq/liam/frontend/internal-packages/mcp-server/src/index.ts"
+      ]
+    }
+  }
+}
--- a/frontend/internal-packages/mcp-server/README.md
+++ b/frontend/internal-packages/mcp-server/README.md
@@ -1,0 +1,57 @@
+# @liam-hq/mcp-server
+
+Model Context Protocol (MCP) server for Liam development tools integration for AI agents.
+
+## Overview
+
+This package provides MCP server functionality that helps Cursor IDE interact with Liam's UI components, offering the following tools:
+
+- `list_components`: Lists all UI components in the Liam UI package
+- `get_component_files`: Gets the contents of a specific UI component's files
+
+## Setup
+
+### Installation
+
+The package is part of the Liam monorepo. Install all dependencies with:
+
+```bash
+pnpm install
+```
+
+## Cursor IDE Integration
+
+To use this MCP server with Cursor IDE, you need to configure a `.cursor/mcp.json` file in the root of your project.
+
+### Setting up mcp.json
+
+1. Create a `.cursor` directory in the project root if it doesn't exist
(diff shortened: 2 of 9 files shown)


frontend/internal-packages/mcp-server/README.md
```diff
@@ -0,0 +1,57 @@
+# @liam-hq/mcp-server
+
+Model Context Protocol (MCP) server for Liam development tools integration for AI agents.
+
+## Overview
+
+This package provides MCP server functionality that helps Cursor IDE interact with Liam's UI components, offering the following tools:
+
+- `list_components`: Lists all UI components in the Liam UI package
+- `get_component_files`: Gets the contents of a specific UI component's files
```
hoshinotsuyoshi, line 10: Just a thought:
I've been using MCP for about a month now, and I was wondering if this explanation would be clear enough for the Cursor editor.

The reason I ask is that when running an MCP server, it's quite common to use it alongside the popular Figma MCP. In that case, if you just say something like “list the components” (which I assume refers to .tsx files), the Cursor editor might not know whether to use Figma or this MCP.

Maybe specifying something like “UI components (.tsx)” could help?
That said, since it seems to be working fine on Miya-san's end, it's probably totally fine! 👍

frontend/internal-packages/mcp-server/src/index.ts
```diff
@@ -0,0 +1,150 @@
...
+
+import { dirname } from 'node:path'
+import { fileURLToPath } from 'node:url'
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = dirname(__filename)
+const uiComponentsPath = path.resolve(
+  __dirname,
+  '../../../packages/ui/src/components',
+)
```
hoshinotsuyoshi, line 16: One aspect that feels uniquely “Liam” about MCP is this part.
If the path here could be passed as a command-line argument, I think it would make the tool more versatile.
That said, for our current use case, this setup is more than enough — so it's totally fine as is! 👍

hoshinotsuyoshi (commented):
LGTM!, one comment, please check .cursor/mcp.json file.

frontend/internal-packages/mcp-server/src/index.ts
MH4GF: It's an internal tool, so it's nice to be able to do this kind of thing 😄 
I would like to improve more and more in the future!

.cursor/mcp.json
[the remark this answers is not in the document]
MH4GF: Oh! It must be deleted. I'll fix!

hoshinotsuyoshi (approved):
👍

PR #1439, opened 2025-04-22
Revert "Remove NOT NULL and UNIQUE constraints from github_repositories migration"

MH4GF:
Reverts liam-hq/liam#1438


NoritakaIkeda (approved):
I've confirmed that the migration was successful! Looks good to me!

PR #1448, opened 2025-04-23
🔧 Refactor migration table to use project_id instead of pull_request_id

MH4GF:
## Issue

- resolve: Refactor migration table to use project_id instead of pull_request_id

## Why is this change needed?
This change improves the database schema design by removing direct dependencies between the migrations table and GitHub-specific tables. By changing from  to , we create a more flexible architecture that follows our schema design patterns and makes the system more maintainable.

- Database schema changes and their impact on existing functionality
- Updated documentation and whether it accurately reflects our best practices
- Implementation of the migration process in the code

## Testing Verification
Tested the migration process to ensure backward compatibility and verified that all affected functionality continues to work correctly.

## What was done
pr_agent:summary

## Detailed Changes
pr_agent:walkthrough

## Additional Notes
This change is part of our ongoing effort to standardize database schema and improve the separation of concerns between GitHub-specific and core application functionality.

### DIFF
--- a/frontend/apps/app/features/migrations/pages/MigrationDetailPage/MigrationDetailPage.tsx
+++ b/frontend/apps/app/features/migrations/pages/MigrationDetailPage/MigrationDetailPage.tsx
@@ -27,18 +27,7 @@ async function getMigrationContents(migrationId: string) {
       id,
       title,
       created_at,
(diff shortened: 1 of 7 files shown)


MH4GF:
-

Issue #557, opened 2025-01-24
Prisma Parser - Support `@@map`


MH4GF:
## Task description

Since the model name and the actual table name can be set separately, it is necessary to support them.
For ERD tools, we want to use properties that match the real table.

```ts
model Comment {
  // Fields

  @@map(“comments”)
}
```

https://www.prisma.io/docs/orm/prisma-schema/data-model/models#mapping-model-names-to-tables-or-collections

I think I can get it with `model.dbName` in DMMF, but I haven't checked.

PR #1590, opened 2025-05-06
Prisma Parser - Support @@map

khiroshi-win:
## Issue
https://github.com/liam-hq/liam/issues/557
- resolve:

### 🤖 Generated by PR Agent at ae61d6e0938ad6ecf3f80c15da3c480144ef533d

- Add sample Prisma schema for local testing
- Demonstrate use of `@@map` and `@map` attributes
- Provide test data for model-to-table mapping support

## Detailed Changes
The test local schema has been updated to fully support the @@Map attribute for proper model-to-table mapping in the Prisma parser, ensuring alignment with database naming conventions while maintaining clean code semantics.
Screenshot_6

schema.prisma

```
datasource db {
  provider = "postgresql"
  url      = env("DATABASE_URL")
}

generator client {
  provider = "prisma-client-js"
}

model User {
  id    Int     @id @default(autoincrement()) @map("_id")
  posts Post[]
  email String  @unique @map("raw_email_address")
  role  Role    @default(USER)

  @@map("users")
}

model Post {
  id     Int   @id @default(autoincrement())
  user   User  @relation(fields: [userId], references: [id])
  userId Int   @map("raw_user_id")

  @@map("posts")
}

enum Role {
  USER
  ADMIN
}
```

## Additional Notes

___

> <details> <summary>  Need help?</summary><li>Type <code>/help how to ...</code> in the comments thread for any questions about Qodo Merge usage.</li><li>Check out the <a href="https://qodo-merge-docs.qodo.ai/usage-guide/">documentation</a> for more information.</li></details>


khiroshi-win:
The test local schema has been updated to fully support the @@map attribute for proper model-to-table mapping in the Prisma parser, ensuring alignment with database naming conventions while maintaining clean code semantics.
Screenshot_6

MH4GF:
@khiroshi-win I can't seem to get the code committed, can you check?

khiroshi-win:
> @khiroshi-win I can't seem to get the code committed, can you check?

I just added the schema.prisma file. I tested it locally and the current parser.ts code works fine, so I don't think there's a need to update it. could you please let me know what have to change?

MH4GF:
@khiroshi-win I see, If so, could you write a test code to describe that it is working?
Sorry if this is already there.

https://github.com/liam-hq/liam/blob/2020c4c71f4822c4a3786aa66140242c5b798490/frontend/packages/db-structure/src/parser/prisma/index.test.ts

khiroshi-win:
> @khiroshi-win I see, If so, could you write a test code to describe that it is working? Sorry if this is already there.
> 
> https://github.com/liam-hq/liam/blob/2020c4c71f4822c4a3786aa66140242c5b798490/frontend/packages/db-structure/src/parser/prisma/index.test.ts

Let me carefully check and I will update. thank you.

schema.prisma
MH4GF, line 1: I think this file is unnecessary.

frontend/packages/db-structure/src/parser/prisma/index.test.ts
```diff
@@ -560,6 +560,140 @@ describe(_processor, () => {
...
+        enum Role {
+          USER
+          ADMIN
+        }
+      `)
+
+      const expectedTables = aSchema({
+        tables: {
+          User: aTable({
+            name: 'User',
```
MH4GF, line 591: The `@@map` specifies the actual database table name, which we want changed.

```suggestion
          users: aTable({
            name: 'users',
```

`@map` is supported because it is a column name mapping, but I thought this one was not.

MH4GF (changes requested):
Please check comments!

frontend/packages/db-structure/src/parser/prisma/index.test.ts
khiroshi-win: I see. let me update the table name.

frontend/packages/db-structure/src/parser/prisma/index.test.ts
```diff
@@ -560,6 +560,140 @@ describe(_processor, () => {
       expect(value).toEqual(expectedTables)
     })
 
+    it('@@map', async () => {
+      const { value } = await processor(`
+        model users {
```
MH4GF, line 565: I want you understand what `@@map` in Prisma does.

The purpose is to keep class and property names handled by Prisma Client readable on the application side, while the DB side uses existing naming conventions (snake_case, plural, etc.) as is.

ref: [Database mapping | Prisma Documentation](https://www.prisma.io/docs/orm/prisma-schema/data-model/database-mapping?utm_source=chatgpt.com)


Therefore, the expected use cases are as follows:

```suggestion
        model User {
```

Perhaps this should be an implementation fix, since `@@map` is not yet supported by Liam.

MH4GF (changes requested):

MH4GF (changes requested):
LGTM 👍🏻 
Please add changeset file?

khiroshi-win:
> LGTM 👍🏻 Please add changeset file?

Ah. I see. let me add right now.

.changeset/breezy-beers-unite.md
```diff
@@ -0,0 +1,5 @@
+---
+"@liam-hq/db-structure": major
```
MH4GF, line 2: According to Semantic Versioning, we only need a **minor** bump here
instead of a **major** one, because:

1. **No breaking change** – the public API surface is untouched.  
2. The change is purely **additive**: the parser *now* understands the
   `@@map` directive, while existing Prisma schemas keep working.  
3. Users integrating `@liam-hq/db-structure` won’t have to modify their
   code after upgrading.

Therefore, could we update the changeset header to:

```suggestion
"@liam-hq/db-structure": minor
```

khiroshi-win: make totally sense.

MH4GF (approved):
LGTM 👍🏻 Thanks!!

PR #1610, opened 2025-05-08
✅ Add tests for organization_members RLS policies from PR #1598

MH4GF:
## Issue

- Adds test coverage for PR #1598 "Fix infinite recursion in organization_members RLS policy"

## Why is this change needed?

PR #1598 fixed issues with the RLS policies for the organization_members table, introducing a new `is_current_user_org_member` function. This PR adds comprehensive test coverage to ensure those changes work as expected and don't regress in the future.

- The test cases cover all expected behaviors of the RLS policies
- A potential security concern is highlighted in test 5 where any authenticated user can add themselves to an organization without invitation
- Is there any other behavior we should test?

## Testing Verification
Executed the test suite for database policies, ensuring all 8 test cases pass.

## What was done

Added comprehensive test suite for organization_members RLS policies and is_current_user_org_member function to validate the fixes implemented in PR #1598. The tests verify proper access control, membership validation, and highlight a potential security concern.

## Detailed Changes

- Added test file `frontend/packages/db/supabase/tests/database/03-organization_members_rls.test.sql` with 8 test cases:
  1. Verifying `is_current_user_org_member` function returns true for org members
  2. Verifying `is_current_user_org_member` function returns false for non-members
  3. Testing RLS policy: Org members can select other members in their org
  4. Testing RLS policy: Non-members cannot select org members
  5. Testing RLS policy: Non-members can add themselves as new members (potential security issue)
  6. Testing RLS policy: Org members can add another user to their org
  7. Testing RLS policy: Non-members cannot add others to an org they don't belong to
  8. Testing RLS policy: Org members can remove another member from their org

## Additional Notes
The tests identify a potential security issue where any authenticated user can add themselves to an organization without invitation. This is noted with a TODO comment in the test file, but should be addressed in a future PR.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/frontend/packages/db/supabase/tests/database/03-organization_members_rls.test.sql
+++ b/frontend/packages/db/supabase/tests/database/03-organization_members_rls.test.sql
@@ -1,0 +1,288 @@
+-- Test file for is_current_user_org_member function and RLS policies
+BEGIN;
+
+-- Load the pgtap extension
+SELECT plan(8);
+
+-- Set role to postgres for preparation
+SET ROLE postgres;
+
+-- Create test users
+SELECT tests.create_supabase_user('org_owner@example.com', 'org_owner');
+SELECT tests.create_supabase_user('non_member@example.com', 'non_member');
+SELECT tests.create_supabase_user('other_user@example.com', 'other_user');
+SELECT tests.create_supabase_user('new_member@example.com', 'new_member');
+SELECT tests.create_supabase_user('temp_user@example.com', 'temp_user');
+
+-- Get user IDs and setup test data
+DO $$
+DECLARE
+    v_owner_id uuid;
+    v_other_user_id uuid;
+BEGIN
+    -- Get user IDs
+    SELECT tests.get_supabase_uid('org_owner@example.com') INTO v_owner_id;
+    SELECT tests.get_supabase_uid('other_user@example.com') INTO v_other_user_id;
+
+    -- Create test organization
+    INSERT INTO organizations (id, name)
+    VALUES ('33333333-3333-3333-3333-333333333333', 'Test Org 1')
+    ON CONFLICT DO NOTHING;
+
+    -- Add org_owner to organization
+    INSERT INTO organization_members (id, user_id, organization_id)
+    VALUES (
+        'dddddddd-dddd-dddd-dddd-dddddddddddd',
+        v_owner_id,
+        '33333333-3333-3333-3333-333333333333'
+    )
+    ON CONFLICT DO NOTHING;
+
+    -- Add other_user to organization for delete test
+    INSERT INTO organization_members (id, user_id, organization_id)
+    VALUES (
+        'eeeeeeee-eeee-eeee-eeee-eeeeeeeeeeee',
+        v_other_user_id,
+        '33333333-3333-3333-3333-333333333333'
+    )
+    ON CONFLICT DO NOTHING;
+END $$;
+
+-- Test 1: is_current_user_org_member function returns true for an org member
+DO $$
+DECLARE
+    v_owner_id uuid;
+BEGIN
+    -- Get user ID
+    SELECT tests.get_supabase_uid('org_owner@example.com') INTO v_owner_id;
+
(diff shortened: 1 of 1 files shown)


frontend/packages/db/supabase/tests/database/03-organization_members_rls.test.sql
```diff
@@ -0,0 +1,288 @@
...
+
+-- Reset authentication context
+RESET ROLE;
+
+-- Test 5: RLS - Non-member can add themselves as a new member
+-- TODO: Security concern - This RLS policy allows any authenticated user to add themselves to any organization without invitation.
+-- This could pose a security risk in a production environment. Consider updating the RLS policy to only allow:
+-- 1. Users who have received an invitation via the invitation system
+-- 2. Or users explicitly approved by existing organization members
+-- The current test confirms the existing behavior but the policy itself should be reviewed.
```
MH4GF, line 142: I noticed this problem while writing test code. This should be fixed because currently anyone can join any organization.
The current behavior is expressed as test code for the time being.

MH4GF: I want to discuss it! @NoritakaIkeda @hoshinotsuyoshi

hoshinotsuyoshi: let me check 👀

NoritakaIkeda: @MH4GF 
I think the current RLS policy was set up because we needed to add the creator to the organization right after creating it.

However, with this setup, any authenticated user can add themselves to any organization, which is a security concern.

A possible solution would be to automatically add the creator as a member using a database trigger or function when the organization is created.

This way, we don't need to loosen the insert RLS policy on the application side, and the issue can be resolved more securely.

MH4GF: > I think the current RLS policy was set up because we needed to add the creator to the organization right after creating it.

I see 👀 
I also thought it made sense to carve out the creation of the organization and member into one special function.

NoritakaIkeda: I'll create a separate PR just for this part and update the tests as well 👍

hoshinotsuyoshi: > I'll create a separate PR just for this part and update the tests as well. 

yes!

hoshinotsuyoshi (approved):

frontend/packages/db/supabase/tests/database/03-organization_members_rls.test.sql
hoshinotsuyoshi: > I'll create a separate PR just for this part and update the tests as well. 👍

thanks! @NoritakaIkeda 

-------

I wrote down one idea!

* Don't allow **any direct inserts** into `organization_members`, even from authenticated users — **simple and safe!**
* Inserts should only happen in these two typical cases:
  * `1:` When `accept_invitation` is called
    * ✅ This one's already done and marked as `SECURITY DEFINER`!
  *  `2:` When an authenticated user runs a **new function to create an organization**
      * That function will make them the first member
      * Basically, we’ll tweak the part in `frontend/apps/app/features/organizations/actions/createOrganizations.ts` where we currently insert into `organization_members` directly
        * Bonus: this makes everything more atomic, which is great too

--------

Any other ideas?

I don't think it's semantically good to require sending an invitation to the person who created the organization (the first member). What do you think?

NoritakaIkeda: @hoshinotsuyoshi 
I agree — it doesn’t make sense to require sending an invitation to the person who just created the organization. It’s redundant and unintuitive. Automatically adding the creator as the first member is much more straightforward and semantically appropriate. 👍

I'll proceed with this approach!

PR #1626, opened 2025-05-09
♻️ relocate migration page components from `features/` to `components/`

junkisai:
## Why is this change needed?

Organized the components under `features/migrations` to match the directory structure introduced in #1623.

### 🤖 Generated by PR Agent at ce8014d0888a55948ad7f4f467ff765a2d28f62a

- Relocate migration page components from `features/` to `components/`
  - Move and refactor all migration-related components under `components/MigrationDetailPage`
  - Add index files for improved component exports
  - Add and update CSS modules for new component structure
- Unify and refactor route and utility libraries
  - Merge `lib/` and `utils/` into unified `libs/` directory
  - Refactor all imports to use new `libs/` paths
  - Add new route generation and schema utilities
- Add and enhance migration feedback and review features
  - Implement comment system for review feedback (add, fetch, modal UI)
  - Add context and hooks for review feedback state management
  - Add copy-to-clipboard functionality for prompts
  - Add radar chart visualization for migration health
  - Add user feedback component with Storybook stories
- Add and update supporting utility and style files
  - Add or update CSS modules for all new/relocated components
  - Add new utility functions for formatting review feedback

## Detailed Changes

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>82 files</summary><table>
<tr>
  <td><strong>MigrationDetailPage.module.css</strong><dd><code>Add and update styles for MigrationDetailPage</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/liam-hq/liam/pull/1626/files#diff-6e2e21384c7cfe252f085b582e6d2c403c6de3467b56e37d8fbda428c6221558">[link]</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>MigrationDetailPage.tsx</strong><dd><code>Refactor and relocate MigrationDetailPage component</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/liam-hq/liam/pull/1626/files#diff-e00ec7902158b7735e0573be6aba53a6cd503bd66d88dbb7d993a6bfa66f33fc">+7/-7</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
[...]

PR #1636, opened 2025-05-09
🔧 Fix import path for langfuseWeb in Storybook configuration

junkisai:
## Why is this change needed?

Fixed the import path for `lib/langfuse` to the correct one, as the Storybook build started failing due to its location being changed in #1626.

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough


junkisai:
same issue: https://github.com/liam-hq/liam/pull/1637

PR #1695, opened 2025-05-16
test: setting ubuntu-24.04-arm

hoshinotsuyoshi:
## Issue

- resolve:

## Why is this change needed?

I want ci result and performance

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough


hoshinotsuyoshi:
note: 
I was able to confirm that it works on arm! 
The performance may have improved, but it wasn't significant. 
I'll close it for now.

PR #1787, opened 2025-05-27
Move figma-to-css-variables package

MH4GF:
## Summary
- move figma-to-css-variables into `internal-packages`
- update workflow and lockfile paths

## Testing
- `pnpm lint:turbo` *(fails: command exited with code 1)*


NoritakaIkeda (approved):
LGTM!

Is my understanding correct that this change was made to exclude the package from the changeset, since it's a development-only library?

hoshinotsuyoshi (approved):
LGTM!

PR #1810, opened 2025-05-29
Add eslint-plugin-css-modules

sakit0:
## Issue

- resolve: Add ESLint rules for CSS Modules

## Why is this change needed?
We need to add `eslint-plugin-css-modules` to detect undefined class names when using CSS Modules. This improves type safety for CSS Modules and helps catch errors early in development.

- Verify that `eslint-plugin-css-modules` dependency is correctly added
- Check if ESLint configuration is properly set up
- Ensure there are no conflicts with existing ESLint rules

## Testing Verification
- Verified by running the project locally and confirming that ESLint errors related to CSS Modules are correctly detected
- Confirmed that the build completes successfully
- Verified that no errors occur in existing CSS Modules usage

## What was done
- Added `eslint-plugin-css-modules` to dependencies
- Added `css-modules` plugin and `no-undef-class` rule to the base ESLint configuration file
- Updated dependency versions in `pnpm-lock.yaml`

## Detailed Changes
1. Added `eslint-plugin-css-modules` version 2.12.0 to `package.json`
2. Added the following configuration to ESLint config file:
   - Enabled `css-modules` plugin
   - Configured `no-undef-class` rule with camelCase option

## Additional Notes
- This change does not affect existing CSS Modules usage
- The new rule is intended to strengthen error detection during development


NoritakaIkeda (approved):
Thank you for the additions! Since the linting passes locally, I believe the changes have been added without any issues. As long as the GitHub Actions CI (especially the lint step) passes, I think it’s good to go and can be merged without problems!

sakit0:
The “request” package used inside eslint-plugin-css-modules contains safety warnings for all versions.
The two packages "eslint-plugin-css-modules" and "request" are not maintained, so we will look for alternatives.

```
pnpm-lock.yaml » request@2.88.2 – Server-Side Request Forgery in Request (moderate severity)
    ↪ https://github.com/advisories/GHSA-p8p7-x288-28g6
  Error: Dependency review detected vulnerable packages.
```

@NoritakaIkeda @MH4GF 
Due to the above reasons, this PR is closed.

NoritakaIkeda:
@sakit0 
Thank you very much for the investigation and the detailed explanation. I understood regarding the vulnerability and the decision to close this PR.👍

PR #1866, opened 2025-06-03
feat: migrate @liam-hq/db package to frontend/internal-packages

MH4GF:
- **feat: migrate @liam-hq/db package to frontend/internal-packages**
- **fix: regenerate pnpm-lock.yaml after merging main branch**
- **fix: update fixture files to reference correct package paths**
- **feat: add symbolic link to root .env in db package - Add symbolic link to allow db package to access environment variables from project root**


MH4GF:
I failed to resolve the conflict and https://github.com/liam-hq/liam/pull/1857 has been closed. Since the content is exactly the same, it will be merged as soon as CI goes through.

PR #1868, opened 2025-06-03
✨(api): allow nullable project_id in design_sessions

MH4GF:
## Issue

- resolve: Allow design sessions to be created without being tied to a specific project

## Why is this change needed?

This change enables the creation of design sessions that are not bound to a specific project, providing greater flexibility for users who want to explore database schemas or create design sessions independently of project context.

- The API endpoint changes in `/api/sessions/route.ts` that handle optional project IDs
- Database schema modifications to support nullable project_id
- Backward compatibility with existing project-based design sessions
- Error handling for cases where project validation is skipped

## Testing Verification

- Tested creating design sessions with and without project IDs
- Verified that existing project-based flows continue to work
- Confirmed proper error handling for invalid project IDs when provided

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough

## Additional Notes

This change maintains full backward compatibility while adding new functionality for project-independent design sessions.


frontend/apps/app/app/api/sessions/route.ts
MH4GF, line 1: `hide whitespace` to make it easier to see 🙏🏻 

Screenshot 2025-06-03 at 19 09 41

frontend/internal-packages/db/package.json
```diff
@@ -14,7 +14,7 @@
     "@biomejs/biome": "1.9.4",
     "@liam-hq/configs": "workspace:*",
     "eslint": "9.27.0",
-    "supabase": "2.23.1",
+    "supabase": "2.24.3",
```
MH4GF, line 17: I faced an unexplained problem with differences between local and CI, but the update solved the problem.

hoshinotsuyoshi (approved):
Seems good!

PR #1937, opened 2025-06-09
✨(db): Refactor project creation to use transactional RPC function

MH4GF:
## Issue

- resolve: https://github.com/liam-hq/liam/issues/XXXX

## Why is this change needed?

The current `addProject.ts` implementation executes project creation, repository creation, and project-repository mapping as separate database operations without transaction management. This creates a risk of data inconsistency if any operation fails partway through the process, as indicated by the FIXME comment in the code.

- The PostgreSQL RPC function implementation for atomic operations
- Error handling and rollback behavior in the RPC function
- TypeScript integration with proper type safety
- Validation logic using valibot for runtime checks

## Testing Verification

- ✅ Local Supabase instance started successfully
- ✅ Migration applied and RPC function created
- ✅ TypeScript types generated and include new RPC function
- ✅ All linting checks pass (Biome, ESLint, TypeScript)
- ✅ Existing functionality preserved (background jobs, redirects)

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough

## Additional Notes

This change resolves the FIXME comment about needing transaction management and follows the existing pattern used by other RPC functions in the codebase like `accept_invitation` and `get_invitation_data`. The implementation ensures data integrity by guaranteeing that either all three database operations succeed or all fail together.


frontend/internal-packages/db/supabase/migrations/20250609120000_add_project_rpc_function.sql
hoshinotsuyoshi, line 1: (I haven’t written many myself either, but...) I think it would be even better if the functions had tests.

hoshinotsuyoshi (approved):
LGTM!

frontend/internal-packages/db/supabase/migrations/20250609120000_add_project_rpc_function.sql
MH4GF: I'll do it with Devin-san!

PR #1961, opened 2025-06-10
feat: integrate LangGraph into streaming workflow with hybrid architecture

FunamaYukina:
## Why is this change needed?

Enable LangGraph usage in streaming workflows while maintaining real-time streaming capabilities through a hybrid architecture approach.

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough


frontend/internal-packages/agent/src/chat/workflow/services/streamingWorkflow.ts
```diff
@@ -0,0 +1,148 @@
+import { END, START, StateGraph } from '@langchain/langgraph'
```
FunamaYukina, line 1: This file has been moved from `frontend/internal-packages/agent/src/chat/workflow/streaming/streamingWorkflow.ts`

PR #1986, opened 2025-06-12
♻️ Refactor User Editing State: Valtio to nuqs + Context

junkisai:
## Issue

- resolve:

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough


frontend/packages/erd-core/src/features/erd/components/ERDContent/hooks/useAutoLayout.ts
```diff
@@ -82,16 +63,23 @@ export const useInitialAutoLayout = ({ nodes, displayArea }: Params) => {
...
   useEffect(() => {
-    initialize()
-  }, [initialize])
+    handleAutoLayout()
+  }, [handleAutoLayout])
+
+  // biome-ignore lint/correctness/useExhaustiveDependencies: To detect changes in query parameters (activeTableName, showMode, hiddenNodeIds) managed by nuqs and trigger re-layout
+  useEffect(() => {
+    handleAutoLayout()
+  }, [activeTableName, showMode, hiddenNodeIds, handleAutoLayout])
```
junkisai, line 84: This code previously used `usePopStateListener` to automatically recalculate the layout on browser back/forward navigation.

However, with the migration to **nuqs**, it's now possible to trigger recalculations by changing the key in `useUserEditing`, so we've integrated that functionality here instead.

frontend/packages/erd-core/src/stores/userEditing/context.ts
junkisai, line 1: I've decided to use a combination of **React Context** and **nuqs**. This is because I have some states that need to be synchronized with query parameters, and others that do not.

frontend/packages/erd-core/src/utils/compressionString.ts
junkisai, line 1: We can't use asynchronous functions with nuqs's custom parser (`parseAsCompressedStringArray`).

So, I've opted to use **lz-string** to implement these features synchronously.

frontend/packages/erd-core/src/features/erd/components/ERDContent/hooks/usePopStateListener.ts
junkisai, line 1: I've made it a hook responsible for changing the state during browser back/forward navigation.

frontend/packages/erd-core/src/features/erd/components/ERDContent/hooks/useQueryParamsChanged.ts
junkisai, line 1: This is a hook that detects changes in query parameters (nuqs state changes) and recalculates the layout to fit the view during browser back/forward operations.

frontend/packages/erd-core/src/features/erd/utils/convertSchemaToNodes.ts
junkisai, line 1: There were cases where the layout result was unintended, even though there were no nodes without relations, due to the addition of a node with the ID `NON_RELATED_TABLE_GROUP_NODE_ID`.

Therefore, I've made it so that if there are no nodes without relations, a node with the ID `NON_RELATED_TABLE_GROUP_NODE_ID` will not be added.

frontend/internal-packages/e2e/tests/e2e/navigation.test.ts
```diff
@@ -96,21 +91,16 @@ test.describe('Navigation and URL Parameters', () => {
...
       await page
-        .getByRole('button', { name: 'Toggle Sidebar Icon Button' })
-        .click()
-      await page
-        .getByRole('button', { name: 'Menu button for accounts', exact: true })
+        .getByTestId('table-name-menu-button-accounts')
         .getByLabel('Hide Table')
         .click()
-      await expect(page).toHaveURL(/.*hidden=eJxLTE7OL80rKQYADrsDYQ/)
+      await expect(page).toHaveURL(/.*hidden=IYYxHsFcDsBcGcg/)
```
junkisai, line 103: I started using lz-string, so the compressed string has changed.

frontend/internal-packages/e2e/tests/e2e/navigation.test.ts
```diff
@@ -4,10 +4,7 @@ const expectUserTableColumnInAccountsTableVisibility = async (
   page: Page,
   visibility: 'visible' | 'hidden',
 ) => {
-  const accountsTable = page.getByRole('button', {
-    name: 'accounts table',
-    exact: true,
-  })
+  const accountsTable = page.getByTestId('rf__node-accounts')
```
junkisai, line 7: Although the rendering was successful, getting the element by `getByRole` was failing for some reason, so I'm changing to get the element by `getByTestId`. 🤔

frontend/internal-packages/e2e/tests/e2e/navigation.test.ts
MH4GF: I would like to know why you switched to lz-string.
I didn't use lz-string to avoid adding unnecessary dependencies, since I can implement this much on my own in the project.

frontend/packages/erd-core/package.json
```diff
@@ -15,9 +15,10 @@
     "clsx": "2.1.1",
     "cmdk": "1.1.1",
     "elkjs": "0.10.0",
+    "lz-string": "1.5.0",
+    "nuqs": "2.4.3",
     "react": "18.3.1",
-    "valibot": "1.1.0",
-    "valtio": "2.1.5"
```
MH4GF, line 20: COOL 😎

MH4GF (approved):
It looks fine. Thanks for the difficult migration!

frontend/internal-packages/e2e/tests/e2e/navigation.test.ts
junkisai: @MH4GF 
I needed to implement a custom parser for nuqs, but I couldn't use my own implementation because I couldn't make `parse` and `serialize` asynchronous functions.

https://github.com/liam-hq/liam/pull/1986#discussion_r2141527413

MH4GF: I see, I didn't see that comment. Understood!

PR #1989, opened 2025-06-12
♻️ refactor: replace type assertion with runtime validation using valibot

FunamaYukina:
## Why is this change needed?

Replaced unsafe type assertion (`as` keyword) with runtime type validation using `valibot` in the AI message service to improve type safety and error handling.🙏

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough

### DIFF
--- a/frontend/apps/app/components/Chat/services/aiMessageService.ts
+++ b/frontend/apps/app/components/Chat/services/aiMessageService.ts
@@ -1,6 +1,7 @@
 'use client'
 
 import type { Schema, TableGroup } from '@liam-hq/db-structure'
+import { boolean, object, optional, parse, string } from 'valibot'
 import { ERROR_MESSAGES } from '../constants/chatConstants'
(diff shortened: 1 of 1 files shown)


frontend/apps/app/components/Chat/services/aiMessageService.ts
```diff
@@ -157,11 +167,8 @@ export const createAndStreamAIMessage = async ({
     )
 
     // Parse JSON response with type safety
-    const data = (await response.json()) as {
-      success?: boolean
-      text?: string
-      error?: string
-    }
+    const rawData = await response.json()
+    const data = parse(ChatAPIResponseSchema, rawData)
```
MH4GF, line 171: If parse is used, an exception is thrown, so error handling should be considered, but since it seems to be thrown in other places, there should be no problem.

MH4GF (approved):
Thanks!

PR #1993, opened 2025-06-13
🔧 implement chat processing to use job-based architecture

FunamaYukina:
## Why is this change needed?

To avoid Vercel function timeout limits, we refactored the LangGraph workflow to use Trigger.dev background jobs, allowing the API to return early while processing continues asynchronously.

Updated the chat API to trigger chat processing as a job using `processChatTask` instead of directly processing messages.

#### Before

```mermaid
sequenceDiagram
    participant U as User
    participant F as Frontend
    participant API as API Route
    participant A as Agent
    participant DB as Database
    
    U->>F: Send Chat Message
    F->>API: POST /api/chat
    API->>API: Create repositories
    API->>A: processChatMessage()
    
    rect rgb(255, 240, 240)
        Note over API,DB: Synchronous Processing
        A->>A: validateInput
        A->>A: generateAnswer (AI)
        A->>DB: Update Schema
        A->>A: formatFinalResponse
    end
    
    A-->>API: Return result
    API-->>F: Return response
    F-->>U: Show result
    
    Note over U,DB: User waits for entire process
```

#### AFTER

```mermaid
sequenceDiagram
    participant U as User
    participant F as Frontend
    participant API as API Route
    participant J as Job Queue
    participant A as Agent
    participant DB as Database
    participant RT as Realtime
    
    U->>F: Send Chat Message
    F->>API: POST /api/chat
    API->>J: Trigger processChatTask
    API-->>F: Return { success: true, jobId }
    F-->>U: Show "Processing..."
    
    Note over U,F: User gets immediate feedback
    
    rect rgb(240, 255, 240)
        Note over J,DB: Asynchronous Job Processing
        J->>J: Create fresh repositories
        J->>A: processChatMessage()
        A->>A: validateInput
        A->>A: generateAnswer (AI)
        A->>DB: Update Schema
        A->>A: formatFinalResponse
        A-->>J: Return result
        J->>DB: Save message to DB
    end
    
    DB->>RT: Notify message update
    RT-->>F: Real-time message
    F-->>U: Show final result
    
    Note over U,DB: Non-blocking user experience

```

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough

### DIFF
--- a/frontend/apps/app/app/api/chat/route.ts
+++ b/frontend/apps/app/app/api/chat/route.ts
@@ -1,6 +1,5 @@
-import { createRepositories } from '@/utils/agentSupabaseHelper'
-import { processChatMessage } from '@liam-hq/agent'
 import { schemaSchema } from '@liam-hq/db-structure'
+import { processChatTask } from '@liam-hq/jobs'
 import { NextResponse } from 'next/server'
 import * as v from 'valibot'
 
@@ -31,21 +30,23 @@ export async function POST(request: Request) {
     )
   }
 
-  const repositories = await createRepositories()
+  try {
+    // Trigger the chat processing job
+    const handle = await processChatTask.trigger({
+      ...validationResult.output,
+    })
 
-  const result = await processChatMessage({
-    ...validationResult.output,
-    repositories,
-  })
-
-  if (!result.success) {
+    return NextResponse.json({
+      success: true,
+      jobId: handle.id,
+      status: 'processing',
+      message: 'Chat processing started',
+    })
+  } catch (error) {
+    console.error('Failed to trigger chat processing job:', error)
     return NextResponse.json(
-      { error: result.error || 'Processing failed' },
+      { error: 'Failed to start processing' },
       { status: 500 },
     )
   }
-
-  return NextResponse.json({
-    success: true,
-  })
 }
--- a/frontend/apps/app/components/Chat/services/aiMessageService.ts
+++ b/frontend/apps/app/components/Chat/services/aiMessageService.ts
@@ -33,6 +33,9 @@ interface SendChatMessageResult {
  */
 const ChatAPIResponseSchema = object({
   success: optional(boolean()),
+  jobId: optional(string()),
+  status: optional(string()),
+  message: optional(string()),
   text: optional(string()),
   error: optional(string()),
 })
--- a/frontend/internal-packages/agent/src/chat/chatProcessor.ts
+++ b/frontend/internal-packages/agent/src/chat/chatProcessor.ts
@@ -3,7 +3,11 @@ import type { Repositories } from '../repositories'
 import { executeChatWorkflow } from './workflow'
 import type { WorkflowState } from './workflow/types'
 
-interface ChatProcessorParams {
+// Export for testing purposes
+export { executeChatWorkflow } from './workflow'
+export type { WorkflowState } from './workflow/types'
+
+export interface ChatProcessorParams {
   message: string
   schemaData: Schema
   history?: [string, string][]
@@ -15,7 +19,7 @@ interface ChatProcessorParams {
   userId: string
 }
 
-type ChatProcessorResult =
+export type ChatProcessorResult =
   | {
       text: string
       success: true
--- a/frontend/internal-packages/jobs/src/trigger/chatJobs.ts
+++ b/frontend/internal-packages/jobs/src/trigger/chatJobs.ts
@@ -1,0 +1,38 @@
+import { createSupabaseRepositories, processChatMessage } from '@liam-hq/agent'
+import type { ChatProcessorParams } from '@liam-hq/agent'
+import { logger, task } from '@trigger.dev/sdk/v3'
+import { createClient } from '../libs/supabase'
+
+// Define type excluding repositories
+type ChatJobPayload = Omit<ChatProcessorParams, 'repositories'>
+
+export const processChatTask = task({
+  id: 'process-chat-message',
+  run: async (payload: ChatJobPayload) => {
+    logger.log('Starting chat processing job:', {
+      buildingSchemaId: payload.buildingSchemaId,
+      messageLength: payload.message.length,
+    })
+
+    try {
+      // Create fresh repositories in job to avoid serialization issues
+      const supabaseClient = createClient()
+      const repositories = createSupabaseRepositories(supabaseClient)
+
+      const result = await processChatMessage({
+        ...payload,
+        repositories,
+      } as ChatProcessorParams)
+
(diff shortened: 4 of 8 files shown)


frontend/internal-packages/jobs/src/trigger/chatJobs.ts
```diff
@@ -0,0 +1,41 @@
...
+    logger.log('Starting chat processing job:', {
+      buildingSchemaId: payload.buildingSchemaId,
+      messageLength: payload.message.length,
+    })
+
+    try {
+      // Create fresh repositories in job to avoid serialization issues
+      // When repositories are passed from API Route to Job, class instances lose their methods
+      // during JSON serialization/deserialization, causing "createMessage is not a function" errors
+      const supabaseClient = createClient()
```
FunamaYukina: I moved repository creation from the API route to the job itself to fix a serialization issue.

When passing repositories from `frontend/apps/app/app/api/chat/route.ts` to the job, class methods were lost during JSON serialization, causing "createMessage is not a function" errors.

Now we create fresh repository instances directly in `frontend/internal-packages/jobs/src/trigger/chatJobs.ts` to ensure all methods are available when the job runs.

frontend/apps/app/app/api/chat/route.ts
```diff
@@ -31,21 +30,23 @@ export async function POST(request: Request) {
...
 
-  const repositories = await createRepositories()
-
-  const result = await processChatMessage({
-    ...validationResult.output,
-    repositories,
-  })
-
-  if (!result.success) {
+  try {
```
MH4GF: trigger() does not seem to throw errors. try-catch may be removed.
ref: https://trigger.dev/docs/triggering#yourtask-trigger

frontend/internal-packages/jobs/src/trigger/chatJobs.ts
MH4GF: Yes, that was the intention. we need to create a repository for jobs, because the suupabase client is slightly different for `app` and `jobs`.

frontend/apps/app/components/Chat/services/aiMessageService.ts
```diff
@@ -33,6 +33,9 @@ interface SendChatMessageResult {
  */
 const ChatAPIResponseSchema = object({
   success: optional(boolean()),
+  jobId: optional(string()),
+  status: optional(string()),
+  message: optional(string()),
```
MH4GF: These values are not used on the front end, correct? If so, I would like to eliminate them. I would like the PR to be issued with as little code as possible.

frontend/internal-packages/agent/src/chat/chatProcessor.ts
```diff
@@ -3,7 +3,11 @@ import type { Repositories } from '../repositories'
 import { executeChatWorkflow } from './workflow'
 import type { WorkflowState } from './workflow/types'
 
-interface ChatProcessorParams {
+// Export for testing purposes
+export { executeChatWorkflow } from './workflow'
+export type { WorkflowState } from './workflow/types'
```
MH4GF: Do I really need to re-export this? Can't we just `import { executeChatWorkflow } from '. /worflow'` at the caller?

frontend/internal-packages/jobs/src/trigger/chatJobs.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export const processChatTask = task({
+  id: 'process-chat-message',
+  run: async (payload: ChatJobPayload) => {
+    logger.log('Starting chat processing job:', {
+      buildingSchemaId: payload.buildingSchemaId,
+      messageLength: payload.message.length,
+    })
+
+    try {
```
MH4GF: Is try-catch really necessary?

MH4GF (changes requested):
I'm including some comments that are a bit harsh, but I'd like to remove any unnecessary codes, so please check back!

frontend/apps/app/app/api/chat/route.ts
FunamaYukina: I'm not very confident, but when I run it without starting trigger.dev, I get the following error. It seems like the error is being caught—do you think so?

ss 3413


Even if I remove the try-catch, the callChatAPI function in /aiMessageService.ts still waits for the response, so a console error like `Error in sendChatMessage: Error: Failed to get response` still appears.

ss 3414

ss 3416


It would be helpful if you could let me know whether I should keep the try-catch block if there's a chance it will be triggered, or if I should remove it anyway in case the error is already being thrown somewhere else.🙏

MH4GF: @FunamaYukina I see! It certainly seems to be threw in that case. Thanks for the confirmation 👍🏻 
However, even in this case, try-catch is not necessary, because Next.js will automatically return a 500 error.

FunamaYukina: I understand! Thank you for your instruction.✨
I found that I do not need to catch the error in this case. I will delete it!

FunamaYukina: I deleted try-catch.Thank you!
https://github.com/liam-hq/liam/pull/1993/commits/c5b21af570592edf8bb2db8b738930d36ea3d5b1

frontend/apps/app/components/Chat/services/aiMessageService.ts
FunamaYukina: As you said, it was an unnecessary response.
I deleted it.
https://github.com/liam-hq/liam/pull/1993/commits/77b6c8ebc2c0991a65c33b91c1d9635b82fbeb77

frontend/internal-packages/agent/src/chat/chatProcessor.ts
FunamaYukina: Thank you very much. I had omitted to check. I will take care.
https://github.com/liam-hq/liam/pull/1993/commits/a4b89d085e6021696e2451336817666ad6d29fa9

frontend/internal-packages/jobs/src/trigger/chatJobs.ts
FunamaYukina: It was an unnecessary try-catch🙏. I checked and it seemed to terminate the error without throwing an error.
https://github.com/liam-hq/liam/pull/1993/commits/840462a95cf9305e476f3888e64d6902b2c69e2c

FunamaYukina:
@MH4GF 
Thanks for the review. I will be careful not to make the same mistake in the future.🙏

MH4GF (approved):
LGTM 👍🏻 Thanks for adjusting!!

PR #2021, opened 2025-06-16
fix: resolve parameter mismatch in `update_building_schema` RPC call

FunamaYukina:
## Why is this change needed?

I implemented two fixes to resolve 404 errors in `update_building_schema` RPC calls:

1. **Client fix**(https://github.com/liam-hq/liam/pull/2021/commits/1b9ca0d8b141558ccf72c91dbfdfc6b063956693): Removed extra `p_design_session_id` parameter from `rpcParams` to match the 6-parameter function signature (**this resolves the 404 error**)
2. **Database cleanup**(https://github.com/liam-hq/liam/pull/2021/commits/6178aced43f98048c7cad6b999b85622dd120e7b ): Added migration to drop the old 5-parameter function version (**not related to this error**, but prevents future conflicts since `supabase db squash` would likely preserve both versions)

Both the parameter mismatch and function overload issues are now resolved.

## Testing Verification

✅ Confirmed update_building_schema RPC calls are now successful.

ss 3426

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough


MH4GF (approved):
I guess I didn't check myself... thank you for your corrective action!

PR #2041, opened 2025-06-17
🔧 feat: implement workflow node tracking for Trigger.dev

FunamaYukina:
## Why is this change needed?

  Add a lightweight logging interface to track LangGraph workflow node execution in Trigger.dev without creating tight coupling between the agent and jobs packages.

| Before  | After |
| ss 3441 | ss 3440  |

 ### Problem

  - LangGraph workflows executed in `Trigger.dev` jobs made it difficult to track which node was currently running
  - The agent package needed to remain independent for standalone execution scenarios
  - Direct use of `Trigger.dev` logger or console.log was not suitable due to architectural constraints

   ### Solution

  Implemented a simple `NodeLogger` interface that:
  - Allows the agent package to emit node execution events without knowing about `Trigger.dev`
  - Enables Trigger.dev jobs to inject their logger implementation
  - Maintains clean separation of concerns between packages

## Testing Verification

We can check it from the preview branch.✅

https://cloud.trigger.dev/orgs/liam-hq-5035/projects/liam-HdAt/env/preview-add-workflow-node-tracking-for-trigger-dev/runs/run_cmc089o873zbd27oow0i4hic7?span=f5df16c86fd522be

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough


frontend/internal-packages/agent/src/chat/workflow/services/workflow.ts
```diff
@@ -55,9 +60,10 @@ const createGraph = () => {
 export const executeWorkflow = async (
   initialState: WorkflowState,
   recursionLimit: number = DEFAULT_RECURSION_LIMIT,
+  log: NodeLogger = () => {},
 ): Promise<WorkflowState> => {
   try {
-    const compiled = createGraph()
+    const compiled = createGraph(log)
```
MH4GF: I think adding a logger function to state would allow us to access the logger from state without major code changes, what do you think?

frontend/internal-packages/agent/src/utils/nodeLogger.ts
```diff
@@ -0,0 +1,2 @@
+// NodeLogger is defined for use with Trigger.dev logging.
+export type NodeLogger = (info: { node: string; state: string }) => void
```
MH4GF: This logger can only assume “that a node has been triggered” and there are other situations that you may want to log.
The following is a type definition of the logger in trigger.dev, but it seems that it could be defined as flexibly as this. What do you think?

```ts
export declare class LoggerAPI implements TaskLogger {
    #private;
    private static _instance?;
    private constructor();
    static getInstance(): LoggerAPI;
    disable(): void;
    setGlobalTaskLogger(taskLogger: TaskLogger): boolean;
    debug(message: string, metadata?: Record<string, unknown>): void;
    log(message: string, metadata?: Record<string, unknown>): void;
    info(message: string, metadata?: Record<string, unknown>): void;
    warn(message: string, metadata?: Record<string, unknown>): void;
    error(message: string, metadata?: Record<string, unknown>): void;
    trace<T>(name: string, fn: (span: Span) => Promise<T>, options?: TraceOptions): Promise<T>;
    startSpan(name: string, options?: TraceOptions): Span;
}
```

MH4GF (commented):
Please check comments!

frontend/internal-packages/agent/src/chat/workflow/services/workflow.ts
FunamaYukina: Oh, you were right, I moved it to state and removed the unnecessary code!
https://github.com/liam-hq/liam/pull/2041/commits/df7bffc5ff8643c817644a8dd3268144d613d967

frontend/internal-packages/agent/src/utils/nodeLogger.ts
FunamaYukina: This is also as you said, and I changed the type of `NodeLogger` as follows!

```ts
export type NodeLogger = {
  debug: (message: string, metadata?: Record<string, unknown>) => void
  log: (message: string, metadata?: Record<string, unknown>) => void
  info: (message: string, metadata?: Record<string, unknown>) => void
  warn: (message: string, metadata?: Record<string, unknown>) => void
  error: (message: string, metadata?: Record<string, unknown>) => void
}
```

FunamaYukina:
I checked the log output in `trigger.dev` again.

ss 3445

frontend/internal-packages/agent/src/chat/chatProcessor.ts
```diff
@@ -30,6 +30,13 @@ export type ChatProcessorResult =
  */
 export const processChatMessage = async (
   params: ChatProcessorParams,
+  log: NodeLogger = {
+    debug: () => {},
+    log: () => {},
+    info: () => {},
+    warn: () => {},
+    error: () => {},
+  },
```
MH4GF: Q: Is there any reason to set the initial value, even though the log is always passed from the caller of processChatMessage?

frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
```diff
@@ -1,15 +1,21 @@
...
+const NODE_NAME = 'analyzeRequirementsNode'
+
 /**
  * Analyze Requirements Node - Requirements Organization
  * Performed by pmAgent
  */
 export async function analyzeRequirementsNode(
   state: WorkflowState,
 ): Promise<WorkflowState> {
+  state.log.log(`[${NODE_NAME}] Started`)
```
MH4GF: IMO: I see, if this is how it is called, it would be easier to understand if it were called `logger`.

```suggestion
  state.logger.log(`[${NODE_NAME}] Started`)
```

MH4GF (approved):
LGTM!
Added minor comments.

frontend/internal-packages/agent/src/chat/chatProcessor.ts
FunamaYukina: Thank you very much. I will check it.

frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
FunamaYukina: I should have used the word logger. Thank you very much!🙏
https://github.com/liam-hq/liam/pull/2041/commits/b320f5ef72a557fe08701412e06872dc6a046c8f

frontend/internal-packages/agent/src/chat/chatProcessor.ts
FunamaYukina: It was an unnecessary initial value. Deleted.🙏
https://github.com/liam-hq/liam/pull/2041/commits/58b9fa21c140e6c521c7be774f226c131319079c

frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
MH4GF: `logger: NodeLogger` The naming and type now match, making it easier to understand!

FunamaYukina:
Thanks for the review. It was very helpful.

PR #2074, opened 2025-06-18
🔧 feat: implement PMAgent workflow    integration with requirements analysis

FunamaYukina:
## Issue

- resolve:

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough

### DIFF
--- a/frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
+++ b/frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
@@ -1,6 +1,18 @@
+import * as v from 'valibot'
+import { PMAgent } from '../../../langchain/agents'
+import type { BasePromptVariables } from '../../../langchain/utils/types'
+import { convertSchemaToText } from '../../../utils/convertSchemaToText'
 import type { WorkflowState } from '../types'
 
 const NODE_NAME = 'analyzeRequirementsNode'
+
+const requirementsAnalysisSchema = v.object({
+  brd: v.string(),
+  functionalRequirements: v.record(v.string(), v.array(v.string())),
+  nonFunctionalRequirements: v.optional(
+    v.record(v.string(), v.array(v.string())),
+  ),
+})
 
 /**
  * Analyze Requirements Node - Requirements Organization
@@ -11,14 +23,50 @@ export async function analyzeRequirementsNode(
 ): Promise<WorkflowState> {
   state.logger.log(`[${NODE_NAME}] Started`)
 
-  // TODO: Implement requirements analysis logic
-  // This node should organize and clarify requirements from user input
+  const pmAgent = new PMAgent()
+  const schemaText = convertSchemaToText(state.schemaData)
+
+  const promptVariables: BasePromptVariables = {
+    schema_text: schemaText,
+    chat_history: state.formattedHistory,
+    user_message: state.userInput,
+  }
+
+  const response = await pmAgent.analyzeRequirements(promptVariables)
+
+  // Parse and validate JSON response
+  let analysisResult: v.InferOutput<typeof requirementsAnalysisSchema>
+  try {
+    const parsed = JSON.parse(response)
+    analysisResult = v.parse(requirementsAnalysisSchema, parsed)
+  } catch {
+    // Fallback: treat response as single requirement
+    analysisResult = {
+      brd: response || 'Failed to parse requirements',
+      functionalRequirements: {},
+      nonFunctionalRequirements: {},
+    }
+  }
+
+  // Log the analysis result for debugging/monitoring purposes
+  // Currently not used elsewhere in the workflow, but useful for observability
+  state.logger.log(`[${NODE_NAME}] Analysis Result:`)
+  state.logger.log(`[${NODE_NAME}] BRD: ${analysisResult.brd}`)
+  state.logger.log(
+    `[${NODE_NAME}] Functional Requirements: ${JSON.stringify(analysisResult.functionalRequirements)}`,
+  )
+  state.logger.log(
+    `[${NODE_NAME}] Non-Functional Requirements: ${JSON.stringify(analysisResult.nonFunctionalRequirements)}`,
+  )
 
   state.logger.log(`[${NODE_NAME}] Completed`)
 
-  // For now, pass through the state unchanged
-  // Future implementation will analyze and organize user requirements
   return {
     ...state,
+    analyzedRequirements: {
+      brd: analysisResult.brd,
+      functionalRequirements: analysisResult.functionalRequirements,
+      nonFunctionalRequirements: analysisResult.nonFunctionalRequirements ?? {},
+    },
   }
 }
--- a/frontend/internal-packages/agent/src/chat/workflow/types.ts
+++ b/frontend/internal-packages/agent/src/chat/workflow/types.ts
@@ -4,6 +4,13 @@ import type { NodeLogger } from '../../utils/nodeLogger'
 
 export type WorkflowState = {
   userInput: string
+  analyzedRequirements?:
+    | {
+        brd: string
+        functionalRequirements: Record<string, string[]>
+        nonFunctionalRequirements: Record<string, string[]>
+      }
+    | undefined
   generatedAnswer?: string | undefined
   finalResponse?: string | undefined
   formattedHistory: string
--- a/frontend/internal-packages/agent/src/langchain/agents/pmAgent/agent.ts
+++ b/frontend/internal-packages/agent/src/langchain/agents/pmAgent/agent.ts
@@ -1,0 +1,46 @@
+import { ChatOpenAI } from '@langchain/openai'
+import { createLangfuseHandler } from '../../utils/telemetry'
+import type { BasePromptVariables, ChatAgent } from '../../utils/types'
+import { PMAgentMode, pmAnalysisPrompt, pmReviewPrompt } from './prompts'
+
+interface PMAgentVariables extends BasePromptVariables {
+  requirements_analysis?: string
+  proposed_changes?: string
+}
+
+export class PMAgent implements ChatAgent {
+  private model: ChatOpenAI
+
+  constructor() {
+    this.model = new ChatOpenAI({
+      model: 'o3',
+      callbacks: [createLangfuseHandler()],
+    })
+  }
+
+  async generate(variables: BasePromptVariables): Promise<string> {
+    // Default to analysis mode for backward compatibility
+    return this.generateWithMode(variables, PMAgentMode.ANALYSIS)
+  }
+
+  async generateWithMode(
+    variables: PMAgentVariables,
+    mode: PMAgentMode,
+  ): Promise<string> {
+    const prompt =
+      mode === PMAgentMode.ANALYSIS ? pmAnalysisPrompt : pmReviewPrompt
+
+    const formattedPrompt = await prompt.format(variables)
+    const response = await this.model.invoke(formattedPrompt)
+    return response.content as string
+  }
+
+  // Convenience methods
+  async analyzeRequirements(variables: BasePromptVariables): Promise<string> {
+    return this.generateWithMode(variables, PMAgentMode.ANALYSIS)
+  }
+
+  async reviewDeliverables(variables: PMAgentVariables): Promise<string> {
+    return this.generateWithMode(variables, PMAgentMode.REVIEW)
+  }
+}
--- a/frontend/internal-packages/agent/src/langchain/agents/pmAgent/prompts.ts
+++ b/frontend/internal-packages/agent/src/langchain/agents/pmAgent/prompts.ts
@@ -1,0 +1,93 @@
+import { ChatPromptTemplate } from '@langchain/core/prompts'
+
+// Requirements Analysis System Prompt
+const pmAnalysisSystemPrompt = `You are PM Agent, a skilled project manager who specializes in analyzing user requirements and extracting structured Business Requirements Documents (BRDs).
+Your role is to:
+1. Analyze user input and conversation history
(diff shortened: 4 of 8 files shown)


frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
```diff
@@ -1,7 +1,19 @@
...
 import type { WorkflowState } from '../types'
 
 const NODE_NAME = 'analyzeRequirementsNode'
 
+const requirementsAnalysisSchema = v.object({
+  brd: v.string(),
+  functionalRequirements: v.record(v.string(), v.array(v.string())),
+  nonFunctionalRequirements: v.optional(
+    v.record(v.string(), v.array(v.string())),
+  ),
```
MH4GF: Making it optional increases the amount of code, since existence checks are required. Make it required.

frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
```diff
@@ -11,14 +23,50 @@ export async function analyzeRequirementsNode(
...
+  try {
+    const parsed = JSON.parse(response)
+    analysisResult = v.parse(requirementsAnalysisSchema, parsed)
+  } catch {
+    // Fallback: treat response as single requirement
+    analysisResult = {
+      brd: response || 'Failed to parse requirements',
+      functionalRequirements: {},
+      nonFunctionalRequirements: {},
+    }
```
MH4GF: Q: I thought it would be better to make an error and retry, what do you think?

frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
```diff
@@ -1,7 +1,19 @@
+import * as v from 'valibot'
+import { PMAgent } from '../../../langchain/agents'
+import type { BasePromptVariables } from '../../../langchain/utils/types'
+import { convertSchemaToText } from '../../../utils/convertSchemaToText'
 import type { WorkflowState } from '../types'
 
 const NODE_NAME = 'analyzeRequirementsNode'
 
+const requirementsAnalysisSchema = v.object({
+  brd: v.string(),
```
MH4GF: It does not seem necessary to make this the only abbreviation. I thought `businessRequirement` would be fine, what do you think?

frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
MH4GF: IMO: Non-functional requirements may be more clearly defined responsibilities for the DBAgent to consider than for the PM Agent to consider. 
But for now, we can leave it as it is and adjust it in the process of improving the accuracy in the future.

MH4GF (commented):
Please check comments!

frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
FunamaYukina: You're right on that. I've pushed the fix!
https://github.com/liam-hq/liam/pull/2074/commits/1e8e5fc9d705b017219d4f6652cb45da6915561c

frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
FunamaYukina: Appreciate that! I've incorporated the fix.
https://github.com/liam-hq/liam/pull/2074/commits/a7b28bc6f7262c62411e1a047c426ab4d05aa863
https://github.com/liam-hq/liam/pull/2074/commits/5073e4716fe5f247280ebab02e2e65d8883cfe29

frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
FunamaYukina: You're absolutely right! The code lacked consistency. I've corrected `brd` to `businessRequirement`.
https://github.com/liam-hq/liam/pull/2074/commits/f297a70e165d381e87d251db36572b693454bd4c

frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
FunamaYukina: You are right, “Non-functional requirements” may not be necessary at this time. It should have been a simple implementation. Thank you for your comment.🌟

frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
```diff
@@ -1,7 +1,55 @@
...
+const requirementsAnalysisSchema = v.object({
+  businessRequirement: v.string(),
+  functionalRequirements: v.record(v.string(), v.array(v.string())),
+  nonFunctionalRequirements: v.record(v.string(), v.array(v.string())),
+})
+
+/**
+ * Execute requirements analysis with retry logic
+ */
+async function executeRequirementsAnalysisWithRetry(
```
MH4GF: It seems to be a code consistency issue to use throw handling only here while other Nodes are handling errors by putting a value in `state.error`. What do you think?
It would be more consistent to use `ConditionalEdge` for retry logic as well, rather than implementing it separately.

frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
```diff
@@ -1,7 +1,55 @@
...
+): Promise<v.InferOutput<typeof requirementsAnalysisSchema>> {
+  let lastError: Error | undefined
+
+  for (let attempt = 1; attempt <= maxRetries; attempt++) {
+    try {
+      const response = await pmAgent.analyzeRequirements(promptVariables)
+
+      // Parse and validate JSON response
+      const parsed = JSON.parse(response)
+      return v.parse(requirementsAnalysisSchema, parsed)
```
MH4GF: I was not comfortable with the response validation being done here. If we want to use PM agents elsewhere, we need to validate at all callers. Perhaps we should do it this way:

```ts
async analyzeRequirements(variables: BasePromptVariables): Promise<v.inferOutput<typeof requirementsAnalysisSchema>>
```

Also, if `analyzeRequirements` returns only JSON, it would be more effective to use Structured Output.

frontend/internal-packages/agent/src/langchain/agents/pmAgent/prompts.ts
```diff
@@ -0,0 +1,93 @@
...
+const pmAnalysisSystemPrompt = `You are PM Agent, a skilled project manager who specializes in analyzing user requirements and extracting structured Business Requirements Documents (BRDs).
+Your role is to:
+1. Analyze user input and conversation history
+2. Extract clear, structured requirements
+3. Convert ambiguous expressions into specific, actionable requirements
+4. Separate multiple use cases into individual requirements
+5. Include specific screens, operations, constraints, and processing details when available
+
+Current Database Schema:
+{schema_text}
```
MH4GF: Is `schema text` really necessary for analyzing requirements? I was wondering about this.
This is information for understanding the product, but perhaps it would be better to have the user upload the documentation in a less direct way.

I know there are many pros and cons to this opinion. However, I took the stance that each agent should focus on the task and that the less unnecessary context, the better the output.
I don't think there is a right answer and I would like to discuss it!

MH4GF (commented):
Please check comment 🙏🏻 
Because agent implementations tend to be complex, I intentionally make our reviews rigorous.

frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
FunamaYukina: Ah, you're absolutely right! The error handling was inconsistent with other nodes. Let me fix that. 
And yes, the retry logic should definitely be handled on the workflow side🙏

frontend/internal-packages/agent/src/langchain/agents/pmAgent/prompts.ts
FunamaYukina: I've reconsidered. You're right, the PM Agent's goal is requirements definition, so schema information would likely become unnecessary context for it. I believe that information should only be known by the DB Agent. I'll plan to remove it!

MH4GF: Thanks for the confirmation. It's a difficult point as there are so many issues to discuss, but I thought it was a good idea to start small. Best regards!

frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
FunamaYukina: I've added a retry mechanism! Each node now keeps track of its retry count in the `WorkflowState`. `workflow.ts` will use a shouldRetry function to decide if a failed node gets another shot, up to 3 times, before moving on.

Also, I've fixed the error handling.

https://github.com/liam-hq/liam/pull/2074/commits/d7ba5c4d76cf786f1ec8629ef5976ac3da16f173

frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
FunamaYukina: You're absolutely right! I realized that I don't need to `JSON.parse` here; I can simply await the promise within the `analyzeRequirements` function. The changes for the retry logic ended up in the same commit, but I've created the commit as follows: https://github.com/liam-hq/liam/pull/2074/commits/d7ba5c4d76cf786f1ec8629ef5976ac3da16f173

frontend/internal-packages/agent/src/langchain/agents/pmAgent/prompts.ts
FunamaYukina: I've addressed this as follows. I considered changing `BasePromptVariables` but decided to leave it as is.
https://github.com/liam-hq/liam/pull/2074/commits/43dc83cf5af4151687b6d6e5b3577f9e41ec2976

frontend/internal-packages/agent/src/chat/workflow/types.ts
```diff
@@ -4,12 +4,20 @@ import type { NodeLogger } from '../../utils/nodeLogger'
...
+        nonFunctionalRequirements: Record<string, string[]>
+      }
+    | undefined
   generatedAnswer?: string | undefined
   finalResponse?: string | undefined
   formattedHistory: string
   schemaData: Schema
   projectId?: string | undefined
   error?: string | undefined
+  retryCount?: Record<string, number> | undefined
```
MH4GF: Please remove undefined as it will require an existence check.

```ts
retryCount: Record<string, number>
```

frontend/internal-packages/agent/src/langchain/agents/pmAgent/agent.ts
```diff
@@ -0,0 +1,54 @@
+import { ChatOpenAI } from '@langchain/openai'
+import * as v from 'valibot'
+import { requirementsAnalysisSchema } from '../../../chat/workflow/nodes/analyzeRequirementsNode'
```
MH4GF: It is better to define `requirementsAnalysisSchema` in this file since it is a circular reference.

MH4GF (approved):
I only commented on the details, LGTM!!
Thanks for adjusting!!

frontend/internal-packages/agent/src/langchain/agents/pmAgent/prompts.ts
MH4GF: Yes, basePromptVariables may need to be changed, but that is fine for another time 👍🏻

frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
MH4GF: Thanks!
I think it's now easier to use on other nodes!

FunamaYukina:
It doesn't work in my local environment, so I'm investigating.🧐

FunamaYukina:
Fixed `PMAgent`.`analyzeRequirements` to parse JSON string response before validation🙏.
https://github.com/liam-hq/liam/pull/2074/commits/4276b856b15cb68574849ebc1c6f8a905e4820e8

I have confirmed✅
https://cloud.trigger.dev/orgs/liam-hq-5035/projects/liam-HdAt/env/preview-add-pm-agent-flow/runs/run_cmc4o5osdnhsx27oiqk7auxcr?span=894e88c3c3e208d0

frontend/internal-packages/agent/src/chat/workflow/types.ts
FunamaYukina: I have fixed it. Thank you!
https://github.com/liam-hq/liam/pull/2074/commits/e6fac3cfb093f86e55912b4c50f1b89ce01b9b42

frontend/internal-packages/agent/src/langchain/agents/pmAgent/agent.ts
FunamaYukina: I have moved the code. Thank you!!
https://github.com/liam-hq/liam/pull/2074/commits/9d05604edeff9b69ffb67e1507274257b6aa42f2

FunamaYukina:
Thank you for the many reviews!

PR #2094, opened 2025-06-19
convert create table ddl 2

MH4GF:
- **feat: enhance operations schema with addTableOperation support**
- **feat: implement postgresqlOperationDeparser for DDL generation**
- **refactor: move constants to operation directory and update imports**


frontend/packages/db-structure/src/operation/constants.ts
MH4GF, line 1: We originally prepared PATH_PATTERNS to implement UI diffs, but decided to move it to a different location because we wanted to use it elsewhere.

frontend/packages/db-structure/src/operation/operationsSchema.ts
```diff
@@ -1,4 +1,28 @@
...
+}
+
+const addTableOperationSchema = v.object({
+  op: v.literal('add'),
+  path: v.custom<`/tables/${string}`>(
+    isTablePath,
+    'Path must match the pattern /tables/{tableName}',
+  ),
+  value: tableSchema,
+})
```
MH4GF, line 17: In the interim, I have decided to strictly define the schema for adding tables, possibly supporting many patterns such as columns, indexes, etc.We'll adjust the strictness once we understand how hard it will be.

frontend/packages/db-structure/src/deparser/postgresql/operationDeparser.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        "CREATE TABLE users (
+          id bigint PRIMARY KEY,
+          email varchar(255) UNIQUE NOT NULL
+        );
+
+        COMMENT ON TABLE users IS 'User table';
+        COMMENT ON COLUMN users.id IS 'User ID';
+        COMMENT ON COLUMN users.email IS 'User email';"
+      `)
+    })
```
MH4GF, line 54: @NoritakaIkeda 
It still only supports `CREATE TABLE`, but using it this way allows us to create DDL from operation!
The operation is the value in `building_schema_versions.patch`.Please let me know if you have any questions!

ref: https://liam-erd-web.vercel.app/app/projects/284f1bca-2633-4a93-9567-7bc98ab298f7/ref/main/schema/frontend/internal-packages/db/schema/schema.sql?showMode=ALL_FIELDS&active=building_schema_versions

junkisai (approved):
That policy seems good!

PR #2114, opened 2025-06-20
✨(agent): implement LangChain structured outputs with valibot integration

MH4GF:
## Issue

- feat: Implement LangChain structured outputs to replace manual JSON parsing

## Why is this change needed?

This change modernizes the database schema build agent to use LangChain's structured outputs feature, providing better type safety and more reliable response parsing. The current implementation relies on manual JSON parsing which can be error-prone and doesn't leverage the latest LangChain capabilities.

- The integration between LangChain's structured outputs and valibot schemas
- The generic ChatAgent interface design that maintains backward compatibility
- The removal of redundant parsing logic while maintaining the same functionality
- The FIXME comment documenting the v.custom() limitation

## Testing Verification

- All lint checks pass (TypeScript, ESLint, Biome)
- Package version constraints are satisfied (syncpack)
- Agent interface remains compatible with existing workflow code
- Error handling is preserved through valibot parsing

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough

## Additional Notes

There is a known limitation with valibot's v.custom() schemas not being convertible to JSON Schema format. This is documented with a FIXME comment and currently handled with errorMode: 'ignore'. Future work should explore alternative validation approaches or replace custom validators with JSON Schema-compatible ones.


frontend/internal-packages/agent/src/chat/workflow/workflow.test.ts
```diff
@@ -242,27 +243,9 @@ describe('Chat Workflow', () => {
...
-      const state = createBaseState({
-        userInput: 'Add a created_at timestamp column to the users table',
-      })
-
-      const result = await executeChatWorkflow(state)
-
-      expect(result.error).toBeUndefined()
-      expect(result.finalResponse).toBe(malformedResponse)
-      expect(mockSchemaRepository.createVersion).not.toHaveBeenCalled()
-    })
```
MH4GF, line 262: Since it is now type-safe, format validation testing is no longer required.

PR #2162, opened 2025-06-23
Feat/erd column diff

junkisai:
## Issue

- resolve:

## Testing Verification

https://github.com/user-attachments/assets/0ac33609-1bf2-4d27-b173-b064027ba0d4

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough


frontend/packages/erd-core/src/features/diff/utils/getChangeStatus.ts
junkisai, line 1: I've moved this file under `TableHeader` because I found that the logic for `getChangeStatus` in `TableHeader` and `TableColumn` is slightly different.

MH4GF (approved):
👍🏻

PR #2172, opened 2025-06-24
feat: add automated Renovate PR review workflow

MH4GF:
## Summary
- Added GitHub Actions workflow to automatically review Renovate PRs using Claude AI
- Workflow triggers on PR creation/update and posts AI-generated code reviews
- Helps streamline the review process for dependency updates

## Test plan
- Workflow will be tested when Renovate creates PRs for dependency updates
- Can be manually tested by triggering the workflow on a test PR

### DIFF
--- a/.github/workflows/renovate-review.yml
+++ b/.github/workflows/renovate-review.yml
@@ -1,0 +1,67 @@
+name: Renovate PR Auto Review
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+
+jobs:
+  renovate-review:
+    # Only run for Renovate PRs
+    if: github.actor == 'renovate[bot]' || github.actor == 'renovate-bot'
+    permissions:
+      contents: read
+      pull-requests: write
+      id-token: write
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/renovate-review.yml
```diff
@@ -0,0 +1,67 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # Need full history for accurate diff analysis
+
+      - name: Auto review Renovate PR
+        uses: anthropics/claude-code-action@main
```
hoshinotsuyoshi: we should use the same pin? 📌

https://github.com/liam-hq/liam/blob/a50970292e6f2ddd1d179e5dbe65b5d681979666/.github/workflows/claude.yml#L36

```suggestion
        uses: anthropics/claude-code-action@beta
```

Or, SHA  ref is more secure?

MH4GF: You are correct that it would be better to use same pin!
You are also right that it is more secure to use SHA.I would like to address this in the next PR.
It would be good to use a linter like https://github.com/suzuki-shunsuke/ghalint 💭

PR #2188, opened 2025-06-25
🔧 Fix duplicate chat message sending caused by component remounting

FunamaYukina:
## Issue

- resolve:

## Why is this change needed?

  Chat messages were being sent twice - once immediately and again  after ~40 seconds. This issue occurred when the Chat component was  remounted due to unstable Supabase authentication state, causing the `autoStartExecuted.current` ref to reset to `false`.

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough

### DIFF
--- a/frontend/apps/app/components/Chat/Chat.tsx
+++ b/frontend/apps/app/components/Chat/Chat.tsx
@@ -35,7 +35,22 @@ export const Chat: FC<Props> = ({ schemaData, designSession }) => {
   )
   const [isLoading, startTransition] = useTransition()
   const messagesEndRef = useRef<HTMLDivElement>(null)
-  const autoStartExecuted = useRef(false)
+
+  // Use sessionStorage to persist autoStartExecuted across component remounts
+  const getAutoStartExecuted = () => {
+    if (typeof window === 'undefined') return false
+    return (
+      sessionStorage.getItem(`autoStartExecuted_${designSession.id}`) === 'true'
+    )
+  }
+
+  const setAutoStartExecuted = (value: boolean) => {
+    if (typeof window === 'undefined') return
+    sessionStorage.setItem(
+      `autoStartExecuted_${designSession.id}`,
+      value.toString(),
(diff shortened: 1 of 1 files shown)


FunamaYukina:
> The variable 'timelineItems' is used here, but it is not clear if it is the same as 'designSession.timelineItems' used earlier. Consider renaming it or ensuring that it is defined, to maintain consistency in naming.

The code uses two different references to timeline items:

  1. designSession.timelineItems - passed as a prop (line 20 in the type definition)
  2. timelineItems - returned from the useRealtimeTimelineItems hook  (line 32)

Maybe we should change the naming of `timelineItems` to `realtimeTimelineItems`💭
I'll do it in another pull request.

MH4GF:
Thanks for noticing and fixing 🙏🏻 
I am wondering if SessionStorage is the best choice for this use case.
@claude I'd love to hear your thoughts as well.

MH4GF:
Session Storage seems to be the best choice. Thank you!
@FunamaYukina If you are exchanging data from outside of React, I think good to use useSyncExternalStore. This will detect state changes and reflect them back to the React component as appropriate. 

ref: https://claude.ai/share/ffb02c1b-11f6-44fb-bca1-0665a8b65820 (ja)

FunamaYukina:
@MH4GF 
Thank you for the research and advice! I'll update it to use `useSyncExternalStore`.

FunamaYukina:
Handled in the following commit!
https://github.com/liam-hq/liam/pull/2188/commits/23f1a68e0e761710ba08d58d3066f78715f755cb

FunamaYukina:
I've checked the preview again and it looks good.
ss 3478
https://cloud.trigger.dev/orgs/liam-hq-5035/projects/liam-HdAt/env/preview-fix-duplicate-workflow-executing/runs

MH4GF (approved):
LGTM 👍🏻 Thanks!!

PR #2191, opened 2025-06-25
✨(app): add idempotency to chat API endpoint

MH4GF:
## Issue

- resolve: N/A

## Why is this change needed?
Currently, the chat API endpoint can process duplicate requests if called multiple times with the same payload. This PR adds idempotency key generation to prevent duplicate task execution in Trigger.dev, ensuring that identical chat requests are only processed once.

- The idempotency key generation logic using SHA-256 hash of the entire payload
- Integration with Trigger.dev's idempotencyKeys SDK
- Removal of timestamp and TTL as requested to handle cases with tens of seconds of delay

## Testing Verification
- Type checks pass
- Linting passes
- Format checks pass
- Manual testing: Need to verify that duplicate chat requests only trigger one task execution

## What was done
pr_agent:summary

## Detailed Changes
pr_agent:walkthrough

## Additional Notes
- Added `@trigger.dev/sdk` dependency to the app package
- Idempotency keys are stored by Trigger.dev for 30 days (default TTL)
- The key is based on design session ID and a hash of the complete validation output

### DIFF
--- a/frontend/apps/app/app/api/chat/route.ts
+++ b/frontend/apps/app/app/api/chat/route.ts
@@ -1,4 +1,6 @@
+import { createHash } from 'node:crypto'
 import { processChatTask } from '@liam-hq/jobs'
+import { idempotencyKeys } from '@trigger.dev/sdk'
 import { NextResponse } from 'next/server'
 import * as v from 'valibot'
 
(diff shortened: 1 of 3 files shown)


FunamaYukina (approved):
Thank you so much! LGTM!🚀

I thought users might resubmit the same content if a job failed, but I realized it's re-sendable because `validationResult` changes (due to the history changing(maybe)).

Issue #2140, opened 2025-06-23
Remove redundant column.unique field from schema


MH4GF:
# Remove redundant `column.unique` field from schema

## Overview
The `column.unique` field in our schema definition is redundant with the existing unique constraints and unique indexes functionality. This creates confusion and maintenance overhead.

## Current State
- `column.unique` (boolean) exists on the Column type
- `UniqueConstraint` exists in the constraints system  
- `index.unique` (boolean) exists for unique indexes
- All three represent the same database concept: uniqueness constraint on a column

## Problem
1. **Redundancy**: The same information is stored in multiple places
2. **Confusion**: Developers need to check multiple places to determine if a column is unique
3. **Maintenance**: Changes need to be synchronized across multiple representations

### Phase 1: Add Compatibility Layer
1. Update parsers to create UNIQUE constraints when encountering unique columns
2. Add utility function to check if a column has a unique constraint
3. Update UI components to use the utility function

### Phase 2: Update Core Components
1. **PostgreSQL Parser** (`frontend/packages/db-structure/src/parser/sql/postgresql/converter.ts`)
   - Remove lines that set `column.unique = true`
   - Ensure UNIQUE constraints are properly created instead

2. **Schema.rb Parser** (`frontend/packages/db-structure/src/parser/schemarb/parser.ts`)
   - Remove extraction of unique option from column definitions
   - Create UNIQUE constraints for columns marked as unique

3. **PostgreSQL Deparser** (`frontend/packages/db-structure/src/deparser/postgresql/utils.ts`)
   - Update `generateColumnDefinition` to check for UNIQUE constraints instead of `column.unique`
   - Generate UNIQUE keyword based on constraints, not column property

4. **Diff System**
   - Remove `buildColumnUniqueDiffItem` function
   - Remove `ColumnUniqueDiffItem` type
   - Remove `columnUniqueSchema` from diff types
   - Update diff tests

### Phase 3: Schema Updates
1. Remove `unique` field from `columnSchema`
2. Remove `columnUniqueSchema` export
3. Update all Column type references
4. Update schema validation

### Phase 4: Testing & Validation
1. Add migration tests to ensure existing schemas work correctly
2. Verify all parsers correctly create UNIQUE constraints
3. Ensure deparser generates correct DDL
4. Validate UI displays uniqueness correctly

## Benefits
- Simpler, more maintainable codebase
- Single source of truth for uniqueness constraints
- Aligns with PostgreSQL's actual implementation
[...]

PR #2224, opened 2025-06-26
feat: remove redundant column.unique field from schema

MH4GF:
## Issue

- resolve: #2140

## Why is this change needed?
This change eliminates redundancy in the schema structure by removing the `column.unique` field. Previously, uniqueness was tracked in two places: the column's `unique` property and as UNIQUE constraints. This led to potential inconsistencies and confusion about the source of truth for uniqueness.

- Verify that all parsers correctly create UNIQUE constraints instead of setting column.unique
- Check that the migration path is clear (existing schemas with column.unique will now use constraints)
- Ensure no functionality is lost in the deparser and diff systems

## Testing Verification
All tests have been updated and are passing:
- ✅ Unit tests for all parsers (PostgreSQL, Schema.rb, Prisma, tbls)
- ✅ Deparser tests updated to verify UNIQUE constraints are not generated inline
- ✅ Diff system tests updated to remove unique field support
- ✅ TypeScript compilation successful
- ✅ Linting and formatting checks pass

### Removed column.unique field
- Removed `columnUniqueSchema` from schema definitions
- Updated `Column` type to exclude the `unique` field
- Updated factory functions to remove `unique: false` defaults

### Updated parsers to use UNIQUE constraints
- **PostgreSQL parser**: Now creates UNIQUE constraints when columns have unique modifiers
- **Schema.rb parser**: Creates UNIQUE constraints for columns with `unique: true` option
- **Prisma parser**: Already used constraints, removed redundant unique field setting
- **tbls parser**: Removed unique column extraction logic

### Updated deparser
- Removed inline UNIQUE constraint generation from column definitions
- Added comment clarifying that UNIQUE should be added as separate constraints

### Removed diff system support
- Deleted `buildColumnUniqueDiffItem` and related tests
- Removed unique field from diff building logic

### Fixed all affected tests
- Updated test expectations to remove unique field
- Added `unique: true` to index definitions where needed
- Fixed compilation errors in dependent packages

## Additional Notes
This is a breaking change for any code that relies on the `column.unique` field. Users should now check for UNIQUE constraints or index uniqueness instead.

Co-Authored-By: Claude <noreply@anthropic.com>


frontend/packages/db-structure/src/deparser/postgresql/operationDeparser.test.ts
```diff
@@ -43,14 +41,14 @@ describe('postgresqlOperationDeparser', () => {
 
       expect(result.errors).toHaveLength(0)
       expect(result.value).toMatchInlineSnapshot(`
-        "CREATE TABLE \"users\" (
-          \"id\" bigint PRIMARY KEY,
-          \"email\" varchar(255) UNIQUE NOT NULL
```
MH4GF, line 48: Since the test case is intended to convert from column.unique, there is no problem to delete it.
There is a separate test to convert from unique constraints to `ALTER TABLE ~~ ADD CONSTRAINT`.

frontend/packages/db-structure/src/parser/prisma/index.test.ts
```diff
@@ -486,20 +479,18 @@ describe(_processor, () => {
...
                 notNull: true,
-                unique: false,
               }),
             },
             indexes: {
               posts_pkey: anIndex({
                 name: 'posts_pkey',
-                unique: true,
                 columns: ['id'],
+                unique: true,
```
MH4GF, line 493: 📝 column.unique has been deleted, but index.unique needs to remain.

NoritakaIkeda (approved):
Thanks! LGTM!
The code looks clean and consistent, and all tests are passing correctly, so I don't see any major risks at this point.

PR #2231, opened 2025-06-26
Feat/chat progress tracking

junkisai:
## Issue

- resolve:

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough


junkisai:
@coderabbitai review

frontend/apps/app/components/Chat/services/timelineItemServiceClient.ts
```diff
@@ -97,7 +124,7 @@ export const setupRealtimeSubscription = (
     .on(
       'postgres_changes',
       {
-        event: 'INSERT',
+        event: '*',
```
hoshinotsuyoshi, line 127: <del>
[nits]

is this change needed?
</del>

hoshinotsuyoshi (approved):
LGTM!

frontend/apps/app/components/Chat/services/timelineItemServiceClient.ts
hoshinotsuyoshi: Sorry, it looks like we’ll need to detect updates to the timeline_items records—so this approach seems like the right one!

hoshinotsuyoshi:
I've approved this, but would appreciate it if others could take a look as well!

FunamaYukina (approved):
Thanks! I think using the existing **Supabase Realtime** feature is a great choice. 
If we encounter performance issues down the line, we might consider streaming the LangGraph nodes directly instead of saving them to the DB.💭

https://langchain-ai.github.io/langgraph/how-tos/streaming/#stream-graph-state

Issue #507, opened 2025-01-20
Cmd+K - Enhanced search experience


MH4GF:

Jal-GG:
hey can i work on this feature ?? please guide me as this is my first open source contribution?

MH4GF:
Hi @Jal-GG, Thanks for comment.
I'm in the process of designing the experience for this issue with the designers on my team.
If you are interested let's discuss at https://github.com/liam-hq/liam/discussions/646.

JeremieAlcaraz:
That would be so cool to have a search bar with CMD + K!

JeremieAlcaraz:
@MH4GF OMGGGGGGGGGGG can"t wait !!!! Looooks AWESOME !!!

MH4GF:
Hey @JeremieAlcaraz @Jal-GG, Actually, in recent versions, typing Cmd+K brings up the Command Palette 😎

https://github.com/user-attachments/assets/f50acc8a-ac5a-498d-82d2-e1d749a85b3e

Currently, search is available by table name, and we are considering various searches and commands to be executed in the future!(change mode, hide table etc...)
Could we get some feedback if you like? cc @tnyo43

Jal-GG:
Awesome feature @MH4GF would love to contribute to this. Sorry was not active much recently

MH4GF:
We have officially released Command Palette with v0.7.0! Please use it, and we welcome your feedback!

https://github.com/liam-hq/liam/releases/tag/%40liam-hq%2Fcli%400.7.0

PR #2294, opened 2025-06-30
💄 style(command palette): refine CommandPalette combobox styles

tnyo43:
## Issue

- resolve:
related to https://github.com/liam-hq/liam/issues/507

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough

### DIFF
--- a/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalette.module.css
+++ b/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalette.module.css
@@ -57,15 +57,17 @@
 .searchFormWithIcon {
   display: flex;
   flex-grow: 1;
-  padding: var(--spacing-2) 0;
-  gap: var(--spacing-2);
+  position: relative;
+  align-items: center;
 }
 
 .searchFormWithIcon [cmdk-input] {
+  padding: var(--spacing-2) 0 var(--spacing-2) var(--spacing-6);
   width: 100%;
 }
 
 .searchIcon {
+  position: absolute;
   width: 16px;
   height: 16px;
   color: var(--overlay-60);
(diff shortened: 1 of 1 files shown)


MH4GF:
I prefer to have padding here, but would like to hear @kumanoayumi's opinion. 

Pasted_Image_2025_07_01__11_53Pasted_Image_2025_07_01__11_53

It's a minor point, so it looks like it should be addressed separately!

kumanoayumi:
@MH4GF @tnyo43 

https://github.com/liam-hq/liam/pull/2294#issuecomment-3021560222

I agree — I feel the same. But when the element isn't focused, adding extra spacing just for the focus ring feels unnecessary... it's tricky.
Maybe we could even consider an approach that doesn't rely on a focus ring at all?
For example, GitHub’s Command Palette doesn’t use one.
Of course, the trade-off is that it relies on the blinking text cursor as a visual cue instead.

https://github.com/user-attachments/assets/c9dfba25-9d17-429a-acec-8257ae0759f5

MH4GF:
@kumanoayumi I see, I suppose there is an option to turn off the focus ring 👍🏻 With the modal display, the user's attention is focused within the command palette, and I think the blinking of the caret alone could indicate the input position.

ref: https://claude.ai/share/3f39c7cd-5fd6-4d67-91f3-0b40d6f9adda

However, I would like to hear @tnyo43's opinion on accessibility!

tnyo43:
@MH4GF @kumanoayumi 
Thanks for your suggestions!

I would like to confirm the behaviour about "auto complete" function int the input element.
If we are going to support the auto complete, I guess we use the Tab key to complete. I feel like the input element actually will always be focused 🤔 
If users understand that the input field doesn't lose focus and that the Tab key is used for completion, then maybe it's reasonable to conclude that an outline isn’t strictly necessary.
Closing the dialog can still be done with the ESC key, which is a very common behavior, and since the dialog indicates that visually, it may not be a big problem if users don’t immediately notice the focus state.

PR #2295, opened 2025-06-30
🐛 fix(command palette): keep the focus within the combobox to improve usability when using both keyboard and mouse

tnyo43:
## Issue

- resolve:

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough

### DIFF
--- a/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalette.test.tsx
+++ b/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalette.test.tsx
@@ -281,6 +281,7 @@ describe('focus on options by single click', () => {
     expect(within(preview).getByText('follows')).toBeInTheDocument()
 
     // focuses on "posts" option and displays a preview of the "posts" table
+    // TODO: It should keep focus on the combobox element when any option is clicked. This test should pass with `user.keyboard('posts')` as well instead of `user.type(searchCombobox, 'posts')`
     await user.type(searchCombobox, 'posts')
     expect(within(preview).getByText('posts')).toBeInTheDocument()
   })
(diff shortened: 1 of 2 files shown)


frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalette.tsx
```diff
@@ -67,10 +69,14 @@ export const CommandPalette: FC = () => {
           <Command.Input
             placeholder="Search"
             onValueChange={() => setFocusedTableName(null)}
+            onBlur={(event) => {
+              if (event.relatedTarget === closeButtonRef.current) return
```
tnyo43, line 73: This line allows user to focus on the close button, especially when using keyboard. This condition should be updated when the dialog has other element to focus, like `event.relatedTarget === closeButtonRef.current || event.relatedTarget === otherRef.current`.

https://github.com/user-attachments/assets/a02a939a-7112-47ab-a934-c93eabce0306

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalette.test.tsx
```diff
@@ -281,6 +281,7 @@ describe('focus on options by single click', () => {
     expect(within(preview).getByText('follows')).toBeInTheDocument()
 
     // focuses on "posts" option and displays a preview of the "posts" table
+    // TODO: It should keep focus on the combobox element when any option is clicked. This test should pass with `user.keyboard('posts')` as well instead of `user.type(searchCombobox, 'posts')`
```
tnyo43, line 284: It looks like the onBlur callback isn't handled properly with happy-dom (or possibly with @testing-library/react).
I tried updating some tests, but they didn’t work as expected. We need to consider a better way to test this behavior.

Updated test https://github.com/liam-hq/liam/commit/538bbbf73296f04ec0534254a76ac34f64780615, which has been reverted.

MH4GF (approved):
I understand the limitations.
This is a pretty usable improvement! Thanks!

PR #2337, opened 2025-07-03
feat: replace AI DDL generation with mechanical deparser

NoritakaIkeda:
## Summary
- Removed AI-powered DDL generation agent (`qaDDLGenerationAgent`)
- Replaced with mechanical DDL generation using existing `postgresqlSchemaDeparser`
- Implemented comprehensive test suite following TDD methodology
- Improved type safety and eliminated LLM dependency for DDL generation

## Test plan
- [x] All existing tests pass
- [x] New test suite covers DDL generation scenarios
- [x] Code passes linting and type checking
- [x] Manual verification of DDL output quality


frontend/packages/db-structure/src/deparser/postgresql/schemaDeparser.test.ts
NoritakaIkeda, line 1: Vitest auto-normalizes snapshots: `\"` becomes `"` for readability, so I removed the backslashes.
ref: https://vitest.dev/guide/migration.html?utm_source=chatgpt.com#snapshots-update-3961

frontend/packages/db-structure/src/deparser/postgresql/utils.ts
```diff
@@ -40,6 +44,49 @@ function formatDefaultValue(value: string | number | boolean): string {
   return value.toString()
 }
 
+/**
+ * Check if a string represents a PostgreSQL function call
+ */
+function isPostgreSQLFunction(value: string): boolean {
```
NoritakaIkeda, line 50: If I enclose CURRENT_TIMESTAMP or gen_random_uuid() in quotes, pglite fails, so I don't enclose them.

frontend/packages/db-structure/src/deparser/postgresql/utils.ts
```diff
@@ -40,6 +44,49 @@ function formatDefaultValue(value: string | number | boolean): string {
...
+    'uuid_generate_v1()',
+    'extract(',
+    'age(',
+    'date_part(',
+    'date_trunc(',
+    'random()',
+    'floor(',
+    'ceil(',
+    'round(',
+  ]
```
hoshinotsuyoshi: @NoritakaIkeda 

Q.
I wasn't quite sure about this part, so I have a question.
Most elements seem to match the pattern `/^[a-zA-Z_][a-zA-Z0-9_]*\s*\(/`.
In that case, what would be the problem with defining it like this?


```
  const commonFunctions = [
    'current_timestamp',
    'current_date',
    'current_time',
  ]
```

hoshinotsuyoshi (commented):
Checking now 🙏

frontend/packages/db-structure/src/deparser/postgresql/ddl-function-syntax.test.ts
MH4GF, line 1: Can you add this test to postgresqlSchemaDeparser.test.ts?

frontend/packages/db-structure/src/deparser/postgresql/utils.ts
```diff
@@ -27,7 +27,11 @@ function generateColumnDefinition(
  */
 function formatDefaultValue(value: string | number | boolean): string {
   if (typeof value === 'string') {
```
hoshinotsuyoshi, line 29: 💭  Just a rough note.

It seems that the logic for handling default values is not very robust, especially for strings.

We might need to revisit what kind of value the input (`schemaSchema`) should be holding.

* For string literals:
  – Should the value include single quotes? (It seems like it currently doesn’t?)

* For expressions:
  – Should the value *not* be wrapped in single quotes?

frontend/packages/db-structure/src/deparser/postgresql/ddl-function-syntax.test.ts
hoshinotsuyoshi, line 1: nits

I understand that this change adds a test to ensure the behavior of `utils.ts`.
That said, I noticed some overlap with `schemaDeparser.test.ts`.
It might be worth considering creating a dedicated test file for `utils.ts`, or splitting up `utils.ts` itself.
But this is a minor point—I'm totally fine with leaving it as is!

frontend/internal-packages/agent/src/chat/workflow/nodes/generateDDLNode.ts
MH4GF, line 1: It seems that generateDDLNode now simply calls `postgresqlSchemaDeparser()`. Isn't this no longer necessary to cut it out as a generateDDLNode, since it can now deparse whenever it has a schema?
How about calling `postgresqlSchemaDeparser()` in the Node where you want to run DDL?

frontend/internal-packages/agent/src/chat/workflow/nodes/designSchemaNode.ts
```diff
@@ -39,8 +40,41 @@ const applySchemaChanges = async (
...
+
+    state.logger.debug(
+      `[${NODE_NAME}] Applying ${schemaChanges.length} patch operations`,
+      {
+        originalTableCount: Object.keys(state.schemaData.tables).length,
+        operations: schemaChanges,
+      },
+    )
+
+    applyPatchOperations(updatedSchemaData, schemaChanges)
```
MH4GF: We are running applyPatchOperations() in createVersion() at the top, why do we need to do this again?

frontend/packages/db-structure/src/deparser/postgresql/ddl-function-syntax.test.ts
NoritakaIkeda: Clarification of input data specifications seems necessary! I will make it an Issue and address it in a separate PR!

frontend/packages/db-structure/src/deparser/postgresql/utils.ts
NoritakaIkeda: Clarification of input data specifications seems necessary! I will make it an Issue and address it in a separate PR!

frontend/internal-packages/agent/src/chat/workflow/nodes/designSchemaNode.ts
NoritakaIkeda: The createVersion() at the top saved it to the DB, but it was not passed to the workflow state and DDL could not be generated. We are therefore running it a second time

Problem Flow: 
1. createVersion() saved to DB ✅ 
2. but workflow state.schemaData is still out of date ❌ 
3. when generating DDL in the next step, the out of date schema data is referenced 
4.As a result, DDL is generated that does not reflect the latest changes.

MH4GF: Why not just return the new schema in the return value of createVersion?

frontend/packages/db-structure/src/deparser/postgresql/utils.ts
NoritakaIkeda: As you pointed out, it was a redundant definition!I have accepted your suggestion!
https://github.com/liam-hq/liam/pull/2337/commits/395a4bd915c73d78d3332593cc37dd4785efc113

frontend/packages/db-structure/src/deparser/postgresql/ddl-function-syntax.test.ts
NoritakaIkeda: Thank you, we have addressed it here 🙏 https://github.com/liam-hq/liam/pull/2337/commits/251de18c79467a9d742be4a8b7eb11d55daaf26a#diff-1203c9741f89ee40cbbb49e3aa2a031464beb89f4046cc4e9ad49d569ad23a01

NoritakaIkeda: Thank you, we have addressed it here 🙏 https://github.com/liam-hq/liam/pull/2337/commits/251de18c79467a9d742be4a8b7eb11d55daaf26a#diff-1203c9741f89ee40cbbb49e3aa2a031464beb89f4046cc4e9ad49d569ad23a01

frontend/internal-packages/agent/src/chat/workflow/nodes/generateDDLNode.ts
NoritakaIkeda: It looked fine to cut out the agent by Node in the unit of use, so I removed generateDDLNode and integrated Node
https://github.com/liam-hq/liam/pull/2337/commits/269ca93c290d1af4784b9978f6ef66451832e7a6

frontend/internal-packages/agent/src/chat/workflow/nodes/designSchemaNode.ts
NoritakaIkeda: Thank you, we have investigated and proceeded as you indicated👍
Our stance is to go in the direction of "include the new schema in the return value of createVersion()".

Until now, we were concerned about the scope of the impact of the change in the createVersion() interface, and were proceeding cautiously, so we had to deal with this by executing applySchemaChanges() on the job side.

However, the process of reapplying the same patch on the job side was duplicated, and we judged that this would be a liability in the medium to long term in terms of readability and maintainability.

Therefore, we have modified it so that state.schemaData in the workflow can be updated directly by including the applied newSchema (newContent) in the return value of createVersion().

[refactor: include newSchema in createVersion return value](https://github.com/liam-hq/liam/pull/2337/commits/7cb089ec6c24899e3fc2adc9935b61194f8fa8ba)

frontend/internal-packages/agent/src/chat/workflow/nodes/designSchemaNode.integration.test.ts
```diff
@@ -0,0 +1,289 @@
...
+    userInput: 'Add a users table with id and name fields',
+    schemaData,
+    logger: mockLogger,
+    formattedHistory: '',
+    retryCount: {},
+    buildingSchemaId: 'test-schema',
+    latestVersionNumber: 1,
+    userId: 'test-user',
+    designSessionId: 'test-session',
+    repositories: mockRepository as never,
```
MH4GF, line 44: It is preferable not to cast by as. Please fix this so it can be done after the merge.

frontend/internal-packages/agent/src/chat/workflow/nodes/executeDdlNode.ts
```diff
@@ -25,17 +26,41 @@ export async function executeDdlNode(
...
+    )
+    state.logger.debug(`[${NODE_NAME}] Generated DDL:`, { ddlStatements })
+  } catch (error) {
+    state.logger.log(`[${NODE_NAME}] DDL generation failed: ${error}`)
+    state.logger.log(`[${NODE_NAME}] Completed`)
+    return {
+      ...state,
+      ddlStatements: 'DDL generation failed due to an unexpected error.',
+    }
+  }
```
MH4GF, line 50: There should be no function in this that throws an error. try-catch can't be turned off?

Issue #2392, opened 2025-07-07
Foreign Key constraints parsing fails for complex PostgreSQL dump files


MH4GF:
## Issue Description

Foreign Key constraints are not being parsed correctly when processing complex PostgreSQL dump files. While the parser handles simple cases correctly, it fails to parse most foreign key relationships in real-world dump files, significantly impacting ERD visualization.

## Root Cause Analysis

Through systematic testing, we've identified that:

1. **Parser logic is correct**: Simple test cases with schema-qualified foreign keys work perfectly
2. **Individual FK syntax is correct**: Even the "failed" foreign keys work when tested individually  
3. **Issue occurs only with complex dump files**: The problem emerges when processing complete PostgreSQL dump files

### Failing Test Case (Real World)
```typescript
it('should parse actual PostgreSQL dump file correctly', async () => {
  // Using real dump file: https://raw.githubusercontent.com/tokuhirom/liam-erd-test/refs/heads/main/src/dump-ec.sql
  const dumpContent = await fs.readFile('dump-ec.sql', 'utf-8')
  const { value } = await processor(dumpContent)
  
  const totalForeignKeys = Object.values(value.tables).reduce((count, table) => {
    return count + Object.values(table.constraints || {}).filter(
      constraint => constraint.type === 'FOREIGN KEY'
    ).length
  }, 0)
  
  console.log(`Foreign keys parsed: ${totalForeignKeys} (expected: 16)`)
  expect(totalForeignKeys).toBe(16) // ❌ FAILS - only 2 FKs parsed
})
```

**Result**: Only 2 out of 16 foreign key constraints are parsed successfully.

### Working Test Cases
All of these test cases **pass successfully**:

```typescript
// ✅ Basic schema-qualified foreign keys (2 FKs) - PASSES
// ✅ Multiple tables and foreign keys (5 FKs) - PASSES  
// ✅ Complex elements (ENUM, functions, comments, sequences) - PASSES
// ✅ Individual testing of "failed" foreign keys (5 FKs) - PASSES
// ✅ Complete synthetic test with all 16 FKs - PASSES
```

### Real Dump File Analysis
- **File**: 1,068 lines, 27,738 characters
- **Expected FKs**: 16 foreign key constraints
- **Parsed FKs**: Only 2 (`fk_page_view_user`, `fk_image_product`)
- **Missing FKs**: 14 constraints lost during parsing
- **Parsing Errors**: 0 (no syntax errors detected)

### Test Results Summary
[...]

PR #2396, opened 2025-07-07
fix: resolve PostgreSQL parser chunked processing context loss for foreign keys

MH4GF:
## Summary

Fixes issue #2392 where only 2 out of 16 foreign key constraints were parsed from real-world PostgreSQL dump files due to chunked processing losing context when ALTER TABLE statements referenced tables from previous chunks.

- ✅ Enhanced convertToSchema() to accept main schema for cross-chunk table lookup
- ✅ Updated all constraint processors to fall back to main schema when tables not found in current chunk
- ✅ Added support for PRIMARY KEY and UNIQUE constraints in ALTER TABLE statements  
- ✅ Added proper error handling for table lookup failures

## Test plan

- [x] All existing PostgreSQL parser tests pass
- [x] Real-world dump file test now correctly parses all 16 foreign key constraints (was 2 before)
- [x] Chunked processing test correctly parses all 8 foreign key constraints with small chunk sizes
- [x] No regression in other constraint types (PRIMARY KEY, UNIQUE, CHECK)


hoshinotsuyoshi (commented):
<del>  As of commit `8115c2e9c30dbb00644cb14e249c066745960fb5`, it appears that there are no differences in the generated `schema.json` when comparing with the `main` branch using the provided `dump-ec.sql`. I’ll double-check it again later. </del>

MH4GF:
@hoshinotsuyoshi Thanks for the confirmation! I've confirmed below that the foreign key is displayed with this fix, if I've missed something, please let me know!

- prod: https://liambx.com/erd/p/raw.githubusercontent.com/tokuhirom/liam-erd-test/refs/heads/main/src/dump-ec.sql?showMode=ALL_FIELDS
- this branch: https://liam-j2pqttexe-liambx.vercel.app/erd/p/raw.githubusercontent.com/tokuhirom/liam-erd-test/refs/heads/main/src/dump-ec.sql?showMode=ALL_FIELDS

hoshinotsuyoshi:
nice urls!

frontend/packages/db-structure/src/parser/sql/postgresql/converter.ts
```diff
@@ -765,9 +774,17 @@ export const convertToSchema = (
...
+
     if (table) {
       table.constraints[foreignKeyConstraint.name] = foreignKeyConstraint
+    } else {
+      // Table not found - this could be the cause of missing foreign keys
+      errors.push(
+        new UnexpectedTokenWarningError(
+          `Table "${foreignTableName}" not found when processing foreign key constraint "${foreignKeyConstraint.name}"`,
+        ),
+      )
```
hoshinotsuyoshi, line 787: @MH4GF 

Is there a possibility that something which previously didn't cause an error might now result in one?
I felt it would be best to prepare a `.changeset` and mark this as at least a minor change.

hoshinotsuyoshi (approved):

frontend/packages/db-structure/src/parser/sql/postgresql/converter.ts
MH4GF: This is a warning, so it just adds a log at runtime. So it is not an error.
However, we need to add a changeset for this behavior!

hoshinotsuyoshi: > This is a warning, so it just adds a log at runtime. So it is not an error.

Ah, you're absolutely right! That makes sense! 👍

MH4GF: Added!: https://github.com/liam-hq/liam/pull/2396/commits/46ee542cf596437c2472e4756810863a0cdbc8e8

MH4GF:
I'm currently checking with tokuhirom-san to see if it is working as expected, but since the test case confirms that this problem exists, we will preemptively merge it.

ref: https://x.com/MH4GF/status/1942106496860565609

hoshinotsuyoshi:
>  As of commit `8115c2e9c30dbb00644cb14e249c066745960fb5`, it appears that there are no differences in the generated `schema.json` when comparing with the `main` branch using the provided `dump-ec.sql`. I’ll double-check it again later.

Sorry, I think it’s probably an issue with my local CLI build environment. I’ll check again later, but I believe this PR is fine.

MH4GF:
@hoshinotsuyoshi 

> Sorry, I think it’s probably an issue with my local CLI build environment. I’ll check again later, but I believe this PR is fine.

I see, thanks for the confirmation!

frontend/packages/db-structure/src/parser/sql/postgresql/converter.ts
MH4GF: @hoshinotsuyoshi Sorry, I pushed one commit after your review. This warning caused the following deparser test to fail. This is an intentional change as we now need to check for referenced tables.
Could you review it just one more time?

https://github.com/liam-hq/liam/pull/2396/commits/e90911e08f75b3fb433ddc98e5fa058de484d2f6

hoshinotsuyoshi: Looks good!
1. test: https://github.com/liam-hq/liam/actions/runs/16133375671/job/45525123835
    - 
2.  fixing commit https://github.com/liam-hq/liam/commit/e90911e08f75b3fb433ddc98e5fa058de484d2f6

hoshinotsuyoshi (approved):

PR #2414, opened 2025-07-08
✨ Add commit-and-push command

MH4GF:
## Issue

- resolve: Add new Claude command for commit and push workflow

## Why is this change needed?
This change adds a new Claude command file that provides a workflow for committing and pushing changes without creating a pull request, complementing the existing commit-and-pr command.

- The content and structure of the new command file
- Whether it follows the same pattern as the existing commit-and-pr command

## Testing Verification
Created the `.claude/commands/commit-and-push.md` file with appropriate content that mirrors the existing commit-and-pr command but focuses on commit and push operations only.

## Additional Notes
This command provides a simpler workflow for cases where users want to commit and push changes without immediately creating a pull request.


junkisai (approved):
👌

PR #2456, opened 2025-07-10
♻️ refact(command palette): separate command palette content

tnyo43:
## Issue

- resolve:

## What was done

pr_agent:summary

## Detailed Changes

pr_agent:walkthrough


frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/__snapshots__/CommandPalette.test.tsx.snap
```diff
@@ -5,13 +5,24 @@ exports[`snapshot test 1`] = `
...
+    id="radix-«r50»"
+  >
+    Command Palette
+  </h2>
+  <p
+    hidden=""
+    id="radix-«r51»"
+  >
+    A search-based interface that allows quick access to various commands and features within the application.
+  </p>
```
tnyo43: The dialog title and description are currently outside of the cmdk-root,
but since they are within the dialog root, that should be fine.

tnyo43 (commented):
Here is a description for snapshot update.
I'm sure this change is fine.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteContent/CommandPaletteContent.tsx
```diff
@@ -36,7 +35,7 @@ export const CommandPaletteContent: FC<Props> = ({ open, closeDialog }) => {
   // Select option by pressing [Enter] key (with/without ⌘ key)
   useEffect(() => {
     const down = (event: KeyboardEvent) => {
-      if (!open || !tableName) return
+      if (!tableName) return
```
tnyo43: In the dialog content, the `open` state is always `true`, so it doesn't have to refer to it.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteContent/CommandPaletteContent.tsx
```diff
@@ -8,7 +8,6 @@ import { TableNode } from '../../../ERDContent/components'
 import styles from '../CommandPalette.module.css'
 
 const getTableLinkHref = (activeTableName: string) => {
-  if (typeof window === 'undefined') return
```
tnyo43: Since the content is rendered only on the client side, it doesn't need to check the window object.

FunamaYukina (approved):
LGTM!!Thanks for the refactoring!♻️✨

PR #2014, opened 2025-06-15
✨ feat(command palette): show table options and table preview

tnyo43:
## Issue

~- resolve:~
relate to https://github.com/liam-hq/liam/issues/507

## Why is this change needed?

This PR introduces table options to the CommandPalette component and displays a preview of the selected table.

|||
|-|-|
|when opened|Screenshot 0007-06-15 at 14 24 00|
|input something in the search combobox|Screenshot 0007-06-15 at 14 24 09|
|input unmatched text|Screenshot 0007-06-15 at 14 24 24|

## What would you like reviewers to focus on?

- If the features are well implemented
  - displays table options
  - input text will filter the options
  - show preview of selected options (via arrow keys, via mouse hover)
- unit tests

## Testing Verification

- In the application
  1. open dialog by ⌘+K
  1. type something in the combobox to filter the options
  1. arrow key or hover select table options and display a preview.

### 🤖 Generated by PR Agent at 8a9046d23aa2e1893e5294c952885c1dd11dd8fc

• Add interactive search functionality with table filtering
• Implement table preview display with ReactFlow integration
• Add comprehensive test coverage for user interactions
• Enhance UI with search input, icons, and styling

## Detailed Changes

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>CommandPalette.module.css</strong><dd><code>Add complete styling for command palette layout</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalette.module.css

• Remove fixed height constraint from dialog content<br> • Add <br>comprehensive styling for search container, input, and icons<br> • <br>Implement main layout with list and preview sections<br> • Add scrollable <br>styling and empty state handling

</details>

  </td>
  <td><a href="https://github.com/liam-hq/liam/pull/2014/files#diff-1b2c933681f2482ea26b144a844c87e82437516eb31e6b5c5c24c78286c49279">+117/-1</a>&nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>CommandPalette.tsx</strong><dd><code>Implement interactive search and table preview functionality</code></dd></summary>
<hr>

[...]

PR #2533, opened 2025-07-12
💄 style(command palette): set border-radius of option

tnyo43:
## Issue

~resolve:~
- fix for https://github.com/liam-hq/liam/pull/2014

## Why is this change needed?

To make it look prettier, I added border-radius.

|before|after|
|--|--|
|Screenshot 0007-07-12 at 10 04 18|Screenshot 0007-07-12 at 10 04 12|


tnyo43:
@junkisai Thanks for reviewing! I fixed the conflict.

junkisai (approved):
thx!

Issue #2411, opened 2025-07-07
[Cloud Spanner] Table relations are not visualized in generated ERD by @liam-hq/cli (>= 0.5.9).


KaoruMuta:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/liam-hq/liam/discussions/categories/q-a).
- [x] I have searched for existing issues [search for existing issues](https://github.com/liam-hq/liam/issues), including closed ones.

### Version Type

CLI Version (npm package)

### Version (only for Self Hosted)

\>= 0.5.9

### Steps to reproduce

1. Set up project with Cloud Spanner (emulator is fine)
2. Run [k1low/tbls](https://github.com/k1LoW/tbls) to output `schema.json` which has schema information with relationships between tables.
  a. example command: `SPANNER_EMULATOR_HOST=localhost:9010 tbls doc`
3. Run latest `liam-hq/cli`
  a. example command: `npx @liam-hq/cli erd build --input schema.json --format tbls`

### Expected Behavior

Table relations are visualized.
As of `liam-hq/cli@0.5.8`, it works like the attached screenshot in [my playground](https://github.com/KaoruMuta/spanner-playground/tree/d2268649e5064189abdfe5f124bf07a4a6eac55d).

<img width="360" alt="Image" src="https://github.com/user-attachments/assets/661a5280-95e4-4272-aa17-8c7fd37f8c70" />

### Actual Behavior

Table relations are not visualized like the attached screenshot  in [my playground](https://github.com/KaoruMuta/spanner-playground/tree/136ace1626a75c7dcb5c44308d36a4cbddea7287).

<img width="360" alt="Image" src="https://github.com/user-attachments/assets/2e02a516-3023-4557-8c56-658758c64611" />

### Additional Context

As far as I investigated, https://github.com/liam-hq/liam/pull/2156 (released as 0.5.9) tried to introduce the new mechanism to visualize relationship without `schema.relationships` by using foreign key relations.
But, Cloud Spanner mostly uses interleave to create parent and children relationships instead of foreign keys. ([ref](https://cloud.google.com/spanner/docs/foreign-keys/overview?hl=ja#fk-and-table-interleaving))

Therefore, I think table relations are not visualized for now.

KaoruMuta:
If I need to provide another information, please let me know.
Anyway, I also try to fix this issue.

MH4GF:
@KaoruMuta 
[...]

PR #2548, opened 2025-07-13
fix: resolve the issue not to visualize Cloud Spanner table relations in generated ERD

KaoruMuta:
## Issue

- resolve: #2411

## Why is this change needed?

As of 0.5.9 or later, table relations in Cloud Spanner are not visualized in generated ERD due to lack information of `relationships` in liam schema.json.
The main cause is that `INTERLEAVE` type in tbls schema cannot be successfully parsed by vailbot.

Therefore, this PR creates custom schema for Cloud Spanner's interleave and fixes processing in tbls, schemarb.
(Prisma does not support Cloud Spanner for now, so there are no changes)

Evidence is just passing unit tests.

## Summary by CodeRabbit

* **New Features**
  * Added support for Google Cloud Spanner interleave constraints, enabling parsing, schema validation, and relationship extraction for interleaved tables.
  * Users can now define and visualize interleave relationships and their ON DELETE actions (CASCADE or NO ACTION) in supported schema tools.

* **Tests**
  * Introduced new test cases to ensure correct parsing and handling of interleave constraints in schema definitions and relationships.

* **Documentation**
  * Updated type exports and schema definitions to include interleave constraint types and options.

### DIFF
--- a/frontend/packages/db-structure/src/parser/schemarb/parser.ts
+++ b/frontend/packages/db-structure/src/parser/schemarb/parser.ts
@@ -184,7 +184,7 @@ function processCallNode(
     return
   }
 
-  // Process Google Cloud Spanner's interleave definition since Rails 7.1.
+  // Process Google Cloud Spanner interleave definition since Rails 7.1.
   // See:https://github.com/googleapis/ruby-spanner-activerecord/blob/a7ca729dc6fe13b1b23fda6698abab9e2c1ace8e/examples/snippets/interleaved-tables/README.md#models-for-interleaved-tables
   if (node.name === 'interleave_in') {
     const argNodes = node.arguments_?.compactChildNodes() ?? []
--- a/frontend/packages/db-structure/src/schema/schema.ts
+++ b/frontend/packages/db-structure/src/schema/schema.ts
@@ -98,10 +98,6 @@ export type InterleaveConstraintReferenceOption = v.InferOutput<
 
 /**
  * Schema for Google Cloud Spanner interleave constraints
- *
- * Interleave constraints define parent-child table relationships where child rows
- * are physically co-located with parent rows for better query performance.
- *
  * @see {@link https://cloud.google.com/spanner/docs/schema-and-data-model#creating-interleaved-tables}
  */
 export const interleaveConstraintSchema = v.object({
--- a/frontend/packages/db-structure/src/utils/constraintsToRelationships.ts
+++ b/frontend/packages/db-structure/src/utils/constraintsToRelationships.ts
@@ -20,7 +20,7 @@ export type Relationship = {
 export type Relationships = Record<string, Relationship>
 
 /**
- * Convert foreign key constraints and Google Cloud Spanner's interleave to relationships for UI display
+ * Convert foreign key constraints and Google Cloud Spanner interleave to relationships for UI display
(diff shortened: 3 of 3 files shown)


frontend/packages/db-structure/src/deparser/postgresql/utils.ts
```diff
@@ -210,6 +210,9 @@ export function generateAddConstraintStatement(
     case 'CHECK':
       return `ALTER TABLE ${tableNameEscaped} ADD CONSTRAINT ${constraintName} CHECK (${constraint.detail});`
 
+    case 'INTERLEAVE':
+      return `ALTER TABLE ${tableNameEscaped} SET INTERLEAVE IN PARENT ${escapeIdentifier(constraint.targetTableName)} ON DELETE ${constraint.deleteConstraint.replace('_', ' ')};`
+
```
KaoruMuta, line 215: [ask]
For now, I added the way to set interleave by alter operation once because I'm not sure about this method responsibility..

I would like to confirm one thing, is this method currently used..?
As far as I confirmed, `postgresqlOperationDeparser` uses this method but it's not used anywhere except unit tests.

frontend/packages/db-structure/src/parser/schemarb/parser.ts
```diff
@@ -581,6 +596,180 @@ function extractCheckConstraintWithTableName(
...
+  constraint: string,
+): 'CASCADE' | 'NO_ACTION' {
+  // For interleave constraints, only CASCADE and NO_ACTION are valid
+  switch (constraint) {
+    case 'cascade':
+      return 'CASCADE'
+    default:
+      return 'NO_ACTION'
+  }
+}
```
KaoruMuta, line 771: [memo]
These are based on foreign key constraint implementation.
The main difference is action type (CASCADE, NO_ACTION only)

frontend/packages/db-structure/src/schema/schema.ts
```diff
@@ -82,6 +82,37 @@ export type ForeignKeyConstraint = v.InferOutput<
...
+  name: constraintNameSchema,
+  columnName: columnNameSchema,
+  targetTableName: tableNameSchema,
+  targetColumnName: columnNameSchema,
+  updateConstraint: v.literal('NO_ACTION'),
+  deleteConstraint: interleaveConstraintReferenceOptionSchema,
+})
+export type InterleaveConstraint = v.InferOutput<
+  typeof interleaveConstraintSchema
+>
```
KaoruMuta, line 114: [memo]
I separated interleave schema from foreign key.

Reason
- There are some differences between foreign key and interleave schema like updateConstraint, type
  - we have to force interleave implementation to logic/type def for foreign key unintentionally
- There is actually little benefit to define common schema for foreign key and interleave in my opinion

frontend/packages/db-structure/src/deparser/postgresql/utils.ts
hoshinotsuyoshi: You're absolutely right — as far as I can tell, that function isn't being called outside of the unit tests yet.
I'm not entirely sure, but I think it's intended to be used by another web app we're currently working on.

MH4GF:
Thanks, I'll review it now. 🏃🏻‍♂️

PR #2520, opened 2025-07-11
♻️ Refactor database schema design workflow to use function agents and messages

MH4GF:
## Issue

- resolve: #2504

## Why is this change needed?
Refactored DesignSchemaNode and Agent to utilize LangGraph messages. messages makes it easier to retry by adding errors to the end of the messages.We are working on addressing this in the next PR.

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Improved schema design workflow with a new, streamlined design agent for database schema operations.

* **Refactor**
  * Replaced the previous class-based schema build agent with a functional design agent approach.
  * Updated prompts and agent naming conventions for clarity and consistency.
  * Simplified agent invocation and message handling for schema design tasks.

* **Bug Fixes**
  * Adjusted agent message sender names to ensure accurate identification in chat history.

* **Tests**
  * Updated and modernized test cases to use the new design agent interface and mocking strategy.

* **Chores**
  * Removed obsolete exports, types, and configuration suppressions related to the old agent implementation.

PR #2555, opened 2025-07-14
fix: correct field name mismatch in design agent response schema

FunamaYukina:
## Issue

- resolve:

## Why is this change needed?

There was a minor oversight in this pull request.
- https://github.com/liam-hq/liam/pull/2520

The design agent was failing with a validation error: **Failed to parse design agent response: ValiError: Invalid key: Expected "operations" but received undefined**

ss 3579

This occurred because there was a mismatch between the prompt template and the response schema:
- The prompt instructed the LLM to return schema changes in a field called "schemaChanges"
- But the parsing code expected the field to be named "operations"

This change updates the prompt template to use "operations" consistently, aligning it with the expected response schema and resolving the parsing failures.

## Testing

I have confirmed that the error has been resolved in my local environment.
（Sometimes it fails with another error.）

ss 3578

### DIFF
--- a/frontend/internal-packages/agent/src/deepModeling.test.ts
+++ b/frontend/internal-packages/agent/src/deepModeling.test.ts
@@ -291,7 +291,7 @@ describe('Chat Workflow', () => {
     it('should handle Build mode with structured JSON response and schema changes', async () => {
       const structuredResponse = {
         message: 'Added created_at column to users table',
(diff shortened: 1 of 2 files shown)


junkisai (approved):
Thank you for the corrections!

MH4GF:
Oh.. Thanks for noticing!!

PR #2560, opened 2025-07-14
fix: resolve message channel update error in design schema workflow

FunamaYukina:
## Issue

- resolve:

## Why is this change needed?

ss 3585

The workflow was failing with "Invalid update for channel 'messages'" errors due to incorrect message state management. The issue occurred because the code was creating a local `messages` array (including a new HumanMessage) and then referencing it when updating the workflow state, causing message duplication and serialization issues.

This change fixes the message update logic by:
- Building the final messages array directly from `state.messages`
- Adding the new HumanMessage and AI response in the correct sequence
- Avoiding references to intermediate message arrays that could cause state conflicts

This ensures proper message channel updates in the LangGraph workflow and prevents the workflow from crashing due to invalid state transitions.

## Testing

https://liam-app-git-fix-message-channel-update-error-liambx.vercel.app/app/design_sessions/5b7985d0-4152-42dd-9ddf-6453da0010d4

ss 3583

### DIFF
--- a/frontend/internal-packages/agent/src/chat/workflow/nodes/designSchemaNode.ts
+++ b/frontend/internal-packages/agent/src/chat/workflow/nodes/designSchemaNode.ts
@@ -198,7 +198,11 @@ export async function designSchemaNode(
   // Clear retry flags after processing
   const finalResult = {
     ...result,
-    messages: [...messages, invokeResult.value.message],
+    messages: [
+      ...state.messages,
+      new HumanMessage(userMessage),
+      invokeResult.value.message,
+    ],
     shouldRetryWithDesignSchema: undefined,
     ddlExecutionFailureReason: undefined,
   }


frontend/internal-packages/agent/src/chat/workflow/nodes/designSchemaNode.ts
```diff
@@ -198,7 +198,11 @@ export async function designSchemaNode(
   // Clear retry flags after processing
   const finalResult = {
     ...result,
-    messages: [...messages, invokeResult.value.message],
+    messages: [
+      ...state.messages,
+      new HumanMessage(userMessage),
+      invokeResult.value.message,
+    ],
```
FunamaYukina, line 205: Corresponds to the same as line 172

https://github.com/liam-hq/liam/blob/26fbb969d3d49689be4b751a621a40f4022c05a0/frontend/internal-packages/agent/src/chat/workflow/nodes/designSchemaNode.ts#L171-L172

PR #2561, opened 2025-07-14
refactor: remove no-throw-error from liam-hq/db-structure

khiroshi-win:
## Issue

- resolve:


hoshinotsuyoshi (commented):
let me check 👀

hoshinotsuyoshi (commented):
I've reviewed commit e060f4cf0a7aea206cbffdac98d558632a76a88b.

Thank you for the PR! Overall, this is a good refactoring effort. I have some observations and suggestions.

##  👍

1. **Non-breaking changes**: I confirmed that public APIs like `parse` function remain unchanged, so there's no impact on users
2. **Code quality improvement**: `eslint-suppressions.json` is now empty, achieving full ESLint compliance 👍
3. **Better error handling**: Internal implementation now uses neverthrow's Result type for more type-safe error handling

## Suggestions (IMHO) 🤔

### About throw statements in operationDeparser.ts

The PR title says "remove no-throw-error", but I noticed the current implementation still throws:

```typescript
// Type definition
export type OperationDeparser = (operation: Operation) => DeparserResult
// DeparserResult = { value: string; errors: DeparserError[] }
```


```typescript
// Actual implementation
if (result.isErr()) {
  throw result.error  // 🤔 Seems inconsistent with the type definition
}
```

## My observation:
- The original throw new Error() was changed to use Result type, but it still throws in the end
- The type definition expects DeparserResult to be returned, but it throws exceptions on error

But I might be missing some context here - perhaps there's a reason for keeping the throws? Would love to
hear your thoughts on this.

## Request 📝

- PR description: Would be helpful to fill in why this change is needed

## Summary

This is a valuable effort to improve error handling. My main question is about the operationDeparser.ts implementation - it seems like we're halfway between throwing and returning errors. If the goal is to fully comply with "no-throw-error", perhaps we could align the implementation with the type definition?

Looking forward to your thoughts on this!

khiroshi-win:
@hoshinotsuyoshi 
Thank you for your review!
type `DeparserResult` has a error field so I think that we don't need to make the `throw result.error`.
how about to implement like that to keep the type?

```
if (result.isErr()) {
   errors.push({ message: result.error.message })
   return { value: '', errors }
}
```

frontend/packages/db-structure/src/deparser/postgresql/schemaDeparser.ts
```diff
@@ -11,14 +11,14 @@ export const postgresqlSchemaDeparser: SchemaDeparser = (schema: Schema) => {
   const errors: { message: string }[] = []
 
   // 1. Generate CREATE TABLE statements for each table
-  for (const table of Object.values(schema.tables) as Table[]) {
+  for (const table of Object.values(schema.tables) satisfies Table[]) {
```
MH4GF, line 14: Using `satisfies` sounds good. Thanks!

frontend/packages/db-structure/src/parser/prisma/index.test.ts
```diff
@@ -314,37 +315,45 @@ describe(_processor, () => {
...
-
-      expect(usersColumnNames).toEqual(['id'])
-      expect(usersColumnNames).not.toContain('posts')
-      expect(postsColumnNames).toEqual(['id', 'user_id'])
-      expect(postsColumnNames).not.toContain('user')
+        return ResultAsync.fromPromise(
+          Promise.resolve(),
+          () => new Error('Unexpected error'),
+        )
+      })
```
MH4GF: It is odd to use ResultAsync in the test code. I guess the original `throw new Error` was for Type Narrowing, but I think it can be tested without using it.
Can you try to write a meaningful test without `throw new Error` or `ResultAsync`?

frontend/packages/db-structure/src/parser/schemarb/parser.ts
```diff
@@ -449,12 +449,17 @@ function extractForeignKeyTableNames(
...
-      return ''
-    },
-  ) as [string, string]
+  const [firstNode, secondNode] = stringNodes.map((node): string => {
+    if (node instanceof StringNode) return node.unescaped.value
+    return ''
+  })
+  if (!firstNode || !secondNode) {
+    return err(
+      new UnexpectedTokenWarningError('Expected exactly 2 string values'),
```
MH4GF: `UnexpectedTokenWarningError` is an error class that allows processing to continue with only an error noted in the log. Can you please determine if this class is really appropriate and if processing should continue or terminate?

frontend/packages/db-structure/src/parser/tbls/parser.ts
```diff
@@ -377,10 +377,53 @@ async function parseTblsSchema(schemaString: string): Promise<ProcessResult> {
...
+            typeof colObj['type'] === 'string' &&
+            typeof colObj['nullable'] === 'boolean'
+          )
+        }
+        return false
+      })
+    }
+
+    return false
+  }
```
MH4GF: If the CompatibleTable type is defined as a valibot and parsed with it, this function definition can be eliminated and simplified.

khiroshi-win:
@MH4GF 
Thank you for your review.
Let me carefully check more.

frontend/packages/db-structure/src/parser/tbls/parser.ts
```diff
@@ -7,7 +8,7 @@ import type {
 } from '../../schema/index.js'
 import { aColumn, anIndex, aTable } from '../../schema/index.js'
 import type { Processor, ProcessResult } from '../types.js'
-import schema from './schema.generated.js'
+import { TablesSchema } from './valibotSchema.js'
```
MH4GF: Sorry, my earlier review was misunderstood.
The `schema.generated.js` uses zod schema for a reason. This is used to create a validator from the JSON Schema provided by tbls.
Therefore, instead of creating a new valibot schema for the CompatibleTable, we should just use the zod schema that originally exists.

MH4GF (approved):
LGTM 👍🏻 Thanks!!

hoshinotsuyoshi:
> @hoshinotsuyoshi Thank you for your review! type `DeparserResult` has a error field so I think that we don't need to make the `throw result.error`. how about to implement like that to keep the type?
> 
> ```
> if (result.isErr()) {
>    errors.push({ message: result.error.message })
>    return { value: '', errors }
> }
> ```

Sorry for the delayed response!
The changes in [this commit](https://github.com/liam-hq/liam/pull/2561/commits/e6641882332b11bde18892e742ce63a203c10133) look great!

PR #2562, opened 2025-07-15
fix: remove non-existent 'primary' property references from database schema prompts

FunamaYukina:
## Issue

- resolve:

## Why is this change needed?

The database schema build agent prompts incorrectly referenced a 'primary' property in column definitions, but this property doesn't exist in the actual column schema.

https://github.com/liam-hq/liam/blob/434b169c169b0fe597b47b8172949bb68fc9cd22/frontend/packages/db-structure/src/schema/schema.ts#L20-L27

Primary keys should only be defined through the constraints object using `primaryKeyConstraintSchema`, not as a column property. This mismatch could lead to confusion and incorrect schema generation attempts by the agent.

This change:
- Removes all references to the non-existent 'primary' column property
- Updates examples to show proper primary key definition via constraints
- Aligns the prompts with the actual schema structure

## Testing

https://liam-app-git-fix-remove-invalid-primary-column-property-liambx.vercel.app/app/design_sessions/88afe70f-8744-4d93-9367-2be2ac9a62fe?showMode=ALL_FIELDS

ss 3584

### DIFF
--- a/frontend/internal-packages/agent/src/langchain/agents/databaseSchemaBuildAgent/prompts.ts
+++ b/frontend/internal-packages/agent/src/langchain/agents/databaseSchemaBuildAgent/prompts.ts
@@ -85,9 +85,9 @@ Example Response:
         "comment": null,
         "indexes": {{}},
         "constraints": {{
-          "users_pkey": {{
+          "pk_users": {{
             "type": "PRIMARY KEY",
-            "name": "users_pkey",
+            "name": "pk_users",
             "columnNames": ["id"]
           }}
         }}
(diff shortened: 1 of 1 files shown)


frontend/internal-packages/agent/src/langchain/agents/databaseSchemaBuildAgent/prompts.ts
```diff
@@ -100,13 +106,18 @@ Example with Foreign Key Constraint:
...
+          "user_id": {{"name": "user_id", "type": "uuid", "notNull": true, "default": null, "comment": "References the user who created the post", "check": null, "unique": false}}
         }},
         "comment": null,
         "indexes": {{}},
         "constraints": {{
+          "posts_pkey": {{
+            "type": "PRIMARY KEY",
+            "name": "posts_pkey",
+            "columnNames": ["id"]
+          }},
```
FunamaYukina:
```suggestion
          "pk_posts": {{
            "type": "PRIMARY KEY",
            "name": "pk_posts",
            "columnNames": ["id"]
          }},
```

junkisai (approved):
Thank you for noticing the cause of the problem!

NoritakaIkeda (approved):
Thank you very much! I think this was a very effective fix!

PR #2564, opened 2025-07-15
✨ Add workflow run status tracking to session detail page

junkisai:
## Issue

- resolve: N/A

## Why is this change needed?

This change integrates workflow run status tracking into the session detail page to provide comprehensive state management for schema generation workflows. Previously, the generating state was not properly tracked after removing the progress role, which led to unclear UX when users were waiting for schema generation to complete.

The implementation adds:
- Server-side workflow run status fetching
- Real-time subscription to workflow run status changes  
- Integration of status into the existing generating state logic

This enables two key improvements:
1. **Loading state control**: Users have clear visibility into when schema generation is in progress (status: 'pending')
2. **Error state control**: The system can now detect and handle workflow failures (status: 'error'), allowing for proper error messaging and recovery options

By tracking workflow run status through the dedicated `workflow_runs` table, the application can provide users with accurate feedback about both successful operations and failures, improving the overall user experience during schema generation workflows.

https://github.com/user-attachments/assets/a860de69-7a76-467d-b3eb-be27c3144fb2

### DIFF
--- a/frontend/internal-packages/db/supabase/migrations/20250714080634_add_workflow_runs_table.sql
+++ b/frontend/internal-packages/db/supabase/migrations/20250714080634_add_workflow_runs_table.sql
@@ -14,7 +14,7 @@ CREATE TABLE IF NOT EXISTS "public"."workflow_runs" (
     "id" "uuid" DEFAULT "gen_random_uuid"() NOT NULL,
     "design_session_id" "uuid" NOT NULL,
     "organization_id" "uuid",
-    "run_id" "uuid" NOT NULL,
+    "workflow_run_id" "uuid" NOT NULL,
     "status" "public"."workflow_run_status" DEFAULT 'pending' NOT NULL,
     "created_at" timestamp(3) with time zone DEFAULT CURRENT_TIMESTAMP NOT NULL,
     "updated_at" timestamp(3) with time zone DEFAULT CURRENT_TIMESTAMP NOT NULL
@@ -33,14 +33,13 @@ ALTER TABLE ONLY "public"."workflow_runs"
 ALTER TABLE ONLY "public"."workflow_runs"
     ADD CONSTRAINT "workflow_runs_organization_id_fkey" FOREIGN KEY ("organization_id") REFERENCES "public"."organizations"("id") ON UPDATE CASCADE ON DELETE RESTRICT;
 
--- Add unique constraint on run_id
+-- Add unique constraint on workflow_run_id
 ALTER TABLE ONLY "public"."workflow_runs"
-    ADD CONSTRAINT "workflow_runs_run_id_key" UNIQUE ("run_id");
+    ADD CONSTRAINT "workflow_runs_workflow_run_id_key" UNIQUE ("workflow_run_id");
 
 -- Add index for better query performance
 CREATE INDEX "workflow_runs_design_session_id_idx" ON "public"."workflow_runs" ("design_session_id");
 CREATE INDEX "workflow_runs_organization_id_idx" ON "public"."workflow_runs" ("organization_id");
-CREATE INDEX "workflow_runs_status_idx" ON "public"."workflow_runs" ("status");
 
 -- Create function to set organization_id from design_session
 CREATE OR REPLACE FUNCTION "public"."set_workflow_runs_organization_id"() RETURNS "trigger"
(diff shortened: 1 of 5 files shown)


frontend/apps/app/components/SessionDetailPage/SessionDetailPageClient.tsx
```diff
@@ -102,10 +109,15 @@ Please suggest a specific solution to resolve this problem.`
...
+    designSessionId,
+    initialWorkflowRunStatus,
+  )
+  // TODO: Include logic to check if the latest version of Schema/DDL exists
   const isGenerating = useMemo(() => {
-    // Since progress role is removed, we no longer track generating state
-    return false
-  }, [timelineItems])
+    return status === 'pending'
+  }, [status, timelineItems])
```
junkisai, line 115: 🔔 The calculation logic for **isGenerating** isn't complete in this PR yet.

After this, we'll adjust the implementation for adding/updating **building_schema_versions** records.

This will allow us to track the creation status of the latest schema versions.
Once this change is in place, the **isGenerating** calculation logic will meet the intended use case, and we'll be able to display loading indicators at the appropriate times.

FunamaYukina (commented):
Thank you!!! just one question 
Do you store one RunId per workflow?

Since each node has its own RunId, I would like to know if you are thinking of saving them.

ss 3590


Testing in preview, one RunId was saved

ss 3591

frontend/internal-packages/db/supabase/migrations/20250714080634_add_workflow_runs_table.sql
```diff
@@ -0,0 +1,184 @@
...
+    ADD CONSTRAINT "workflow_runs_organization_id_fkey" FOREIGN KEY ("organization_id") REFERENCES "public"."organizations"("id") ON UPDATE CASCADE ON DELETE RESTRICT;
+
+-- Add unique constraint on run_id
+ALTER TABLE ONLY "public"."workflow_runs"
+    ADD CONSTRAINT "workflow_runs_run_id_key" UNIQUE ("run_id");
+
+-- Add index for better query performance
+CREATE INDEX "workflow_runs_design_session_id_idx" ON "public"."workflow_runs" ("design_session_id");
+CREATE INDEX "workflow_runs_organization_id_idx" ON "public"."workflow_runs" ("organization_id");
+CREATE INDEX "workflow_runs_status_idx" ON "public"."workflow_runs" ("status");
```
hoshinotsuyoshi: nits

Since `status` is an enum with only a few possible values(three values: pending, process, error), this index might not offer much performance benefit and could increase write overhead. Consider removing it unless you’ve observed clear query performance gains.

hoshinotsuyoshi (approved):
LGTM! I’d appreciate it if you could also address Funama-san’s question [in this comment](https://github.com/liam-hq/liam/pull/2564#pullrequestreview-3022922921) — I’m curious about it as well.

junkisai:
@FunamaYukina 
Yes, I was indeed assuming to manage one **run ID** per workflow!

Currently, I don't plan to store individual **run IDs** for each node. However, there's a risk they might get confused with the workflow-specific **run IDs**, so to be safe, I've decided to rename the column to `workflow_run_id`.

FunamaYukina:
@junkisai 
Understood. Thank you very much!
If we consider storing each node's runId and status as well in the future, we may want to create a separate table💭

junkisai:
@FunamaYukina 

> If we consider storing each node's runId and status as well in the future, we may want to create a separate table💭

I was thinking of handling it by creating a `node_runs` table like the following:

```sql
CREATE TABLE IF NOT EXISTS "node_runs" (
  id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
  workflow_run_id uuid REFERENCES workflow_runs(id) ON DELETE CASCADE,
  node_run_id uuid NOT NULL,
  status node_run_status DEFAULT 'pending' NOT NULL,
  created_at timestamp(3) with time zone DEFAULT CURRENT_TIMESTAMP NOT NULL,
  updated_at timestamp(3) with time zone DEFAULT CURRENT_TIMESTAMP NOT NULL
);
```

frontend/internal-packages/db/supabase/migrations/20250714080634_add_workflow_runs_table.sql
junkisai: I've deleted the unnecessary indexes here. 🙏 

[🔧 Rename run_id to workflow_run_id in workflow_runs table](https://github.com/liam-hq/liam/pull/2564/commits/3cbe20e657a8ccf34725f1a112f4d2f4e036d802)

FunamaYukina:
@junkisai It seems very good!😄
Thank you for sharing your thoughts. This PR looks good to me, so I'll go ahead and approve it.

PR #2565, opened 2025-07-15
🔥(api): Remove unused API route for building schema versions

junkisai:
## Issue

- resolve: N/A

## Why is this change needed?
Remove unused API route for building schema versions and make related database columns nullable.


FunamaYukina (approved):
Thank you for noticing that! Looks good to me.👍

PR #2571, opened 2025-07-16
✨ Separate version creation and update for better tracking

junkisai:
## Issue

- resolve: N/A (Internal improvement)

## Why is this change needed?

This change improves schema version tracking by separating the creation and update processes:

**As-Is**: New schema versions are created with INSERT when the complete schema is ready
- This makes it impossible to track when schema construction started
- Version creation timing is unclear

**To-Be**: Schema versions are created with NULL values before job execution, then updated when construction completes
- Better tracking of schema construction lifecycle
- Clear separation between version creation and schema application
- Improved debugging and monitoring capabilities

## Changes Made

- Added `createEmptyVersion` method to create versions with null patches
- Added `updateVersion` method to apply patches to existing versions  
- Updated `designSchemaNode` to use two-step version creation process
- Removed RPC dependency in favor of direct table operations
- Improved version conflict detection and error handling
- Updated all related tests to support new repository interface

## Testing

https://github.com/user-attachments/assets/980b94c9-cf60-4c63-a31f-6b84f4a52161


frontend/internal-packages/agent/src/repositories/supabase.ts
```diff
@@ -210,41 +200,130 @@ export class SupabaseSchemaRepository implements SchemaRepository {
     return versions.length > 0 ? Math.max(...versions.map((v) => v.number)) : 0
   }
 
-  async createVersion(params: CreateVersionParams): Promise<VersionResult> {
-    const { buildingSchemaId, latestVersionNumber, patch } = params
+  async createEmptyVersion(
+    params: CreateEmptyVersionParams,
```
hoshinotsuyoshi, line 204: I see! how about name `createEmptyPatchVersion` ?

hoshinotsuyoshi (approved):

frontend/internal-packages/agent/src/repositories/supabase.ts
junkisai: You're right, I think the expression you suggested is clearer and better!

PR #2577, opened 2025-07-16
refactor: update no-throw-error from liam-hq/schema-bench

khiroshi-win:
## Issue

- resolve:


frontend/internal-packages/schema-bench/src/workspace/evaluation/evaluation.test.ts
```diff
@@ -290,33 +291,42 @@ describe('evaluateSchema', () => {
       consoleSpy.mockRestore()
     })
 
-    it('should create individual result files with correct naming', async () => {
+    it('should create individual result files with correct naming', (): ResultAsync<
```
MH4GF: Something is wrong with using the Result type in the test code. Is it possible to write test code without throwing Error in the first place?

frontend/internal-packages/schema-bench/src/workspace/evaluation/evaluation.test.ts
```diff
@@ -308,13 +308,12 @@ describe('evaluateSchema', () => {
...
+        expect.fail('No result files found')
       }
       const resultFile = path.join(evaluationDir, firstResultFile)
-      const content = JSON.parse(fs.readFileSync(resultFile, 'utf-8')) as {
-        caseId: string
-        metrics: unknown
-      }
+      const ResultFileContent = fs.readFileSync(resultFile, 'utf-8')
+      const content: { caseId: string; metrics: unknown } =
+        JSON.parse(ResultFileContent)
```
hoshinotsuyoshi: nits


```suggestion
      const resultFileContent = fs.readFileSync(resultFile, 'utf-8')
      const content: { caseId: string; metrics: unknown } =
        JSON.parse(resultFileContent)
```

PR #2585, opened 2025-07-16
feat: implement accordion UI for DDL/DML query results

NoritakaIkeda:
## Issue

- resolve: Implements accordion UI for DDL/DML query results display

## Why is this change needed?
This change implements an accordion UI component to display DDL/DML query execution results in a more organized and user-friendly way. The accordion structure allows users to expand/collapse individual query results, making it easier to navigate through multiple query executions and their outcomes.


NoritakaIkeda:
https://github.com/liam-hq/liam/pull/2592 We can handle it here

PR #2598, opened 2025-07-17
🎨 Refactor timeline items to use unified type system with role-based agents

junkisai:
## Issue

- resolve: Refactor multi-agent message type system for better maintainability

## Why is this change needed?

The current timeline item type system has separate types for each agent (assistant_pm, assistant_db, assistant_qa). This change unifies the type system by using a single `assistant` type with a `role` field to distinguish between agents, improving code maintainability and extensibility.

### Changes Made

- Consolidated `assistant_pm`, `assistant_db`, `assistant_qa` types into a unified `assistant` type
- Added `role` field to identify agent roles ('pm' | 'db' | 'qa')
- Implemented role-based rendering using pattern matching in `TimelineItem` component
- Updated stories and tests to use the new type structure
- Improved error handling with conditional `onRetry` prop passing

### Technical Details

- Defined `AssistantRole` type ('pm' | 'db' | 'qa')
- Removed individual agent types from `TimelineItemEntry` union
- Updated `timelineUtils.ts` to generate unified agent entries
- Ensured type safety with pattern matching using `.exhaustive()`

This refactoring makes it easier to add new agent types and reduces code duplication while maintaining the same functionality.


MH4GF (approved):
Thanks!

PR #2599, opened 2025-07-17
🔧(agent): Simplify deepModeling return type to WorkflowState

MH4GF:
## Issue

- resolve: https://github.com/route06/liam-internal/issues/5062

## Why is this change needed?

Currently, the `generatedAnswer` generated in `designSchemaNode` is stored in state and later saved as timeline_items in `finalizeArtifactsNode`. However, by saving directly to timeline_items at the `designSchemaNode` stage, we can eliminate the need to store `generatedAnswer` in state, enabling the following optimizations:

- **Reduced State Size**: No need to carry `generatedAnswer` through the workflow state
- **Improved Data Flow**: Immediate storage at generation time, clearer flow
- **Performance Improvement**: Reduced overhead from carrying unnecessary data
- **Code Simplification**: Elimination of duplicate save logic

Pasted_Image_2025_07_17__18_59

Additionally, to enable meaningful testing, we changed the `deepModeling` function's return value to the actual workflow state, allowing tests to verify observable behavior instead of meaningless success flags.

### Key Changes:

1. **Direct Timeline Storage**: `designSchemaNode` now saves responses directly to timeline_items when generated
2. **State Cleanup**: Removed unused `generatedAnswer` and `finalResponse` from WorkflowState and LangGraph annotations  
3. **Improved Return Type**: Changed `DeepModelingResult` from `Result<{ text: string }, Error>` to `Result<WorkflowState, Error>`
4. **Enhanced Error Handling**: Timeline item save failures now properly propagate errors via `state.error`
5. **Better Test Assertions**: Tests now verify actual schema changes and state instead of generic success flags

### Benefits:

- **Memory Efficiency**: Reduced serialization overhead in workflow state transitions
- **Better Testability**: Tests verify actual observable behavior (schema changes, error states, etc.)
- **Improved Debugging**: Full workflow state available for inspection
- **YAGNI Compliance**: Eliminated unused fields and dead code
- **Enhanced Error Handling**: Proper error propagation instead of silent console.error logging

### DIFF
--- a/frontend/internal-packages/agent/src/deepModeling.test.ts
+++ b/frontend/internal-packages/agent/src/deepModeling.test.ts
@@ -133,8 +133,8 @@ describe('Chat Workflow', () => {
 
     expect(result.isOk()).toBe(true)
     if (result.isOk()) {
-      expect(result.value.text).toBe('Mocked agent response')
-      expect(typeof result.value.text).toBe('string')
+      expect(result.value.schemaData).toBeDefined()
+      expect(result.value.error).toBeUndefined()
     }
     expect(mockInvokeDesignAgent).toHaveBeenCalledOnce()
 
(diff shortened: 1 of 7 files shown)


frontend/internal-packages/agent/src/deepModeling.test.ts
```diff
@@ -395,7 +428,8 @@ describe('Chat Workflow', () => {
 
       expect(result.isOk()).toBe(true)
       if (result.isOk()) {
-        expect(result.value.text).toBe('Added created_at column to users table')
+        expect(result.value.schemaData).toEqual(expectedSchema)
+        expect(result.value.error).toBeUndefined()
```
MH4GF, line 432: Currently the tests are of little value because we are mocking the repository and if we try to save data, we cannot retrieve that data from the tests.
We believe this can probably be solved by implementing a repository for testing. I intend to do this in the next PR.

FunamaYukina (approved):
LGTM!

If `validateSchemaNode` returns false, the build agent may repeat “Added!” two or more times, but I thought it would be good to have it timed with the generation of the ERD.

Thank you for even updating the README!

PR #2606, opened 2025-07-18
💄 Fix long text wrapping in VersionMessage component

kumanoayumi:
## Issue

- resolve:

## Why is this change needed?

When VersionMessage component displays long paths, the text doesn't wrap and overflows horizontally, breaking the UI layout.
This fix ensures that long paths wrap properly within the container. The arrow icons (→) are also handled correctly with the wrapping, improving  readability.

| Before | After |
| ------ | ----- |
| スクリーンショット 2025-07-17 18 25 20 | スクリーンショット 2025-07-18 11 27 35 |


kumanoayumi:
@FunamaYukina Thank you for your review 🚀

PR #2636, opened 2025-07-22
[apps/app] Enable TypeScript unused variable detection and clean up existing

FunamaYukina:
## Issue

- resolve:

## Why is this change needed?

I discovered that `timelineUtils.ts` contained several unused functions.
e.g (`generateAgentContent`, `addPMAgentProgress`, `addDBAgentProgress`)

（The file has already been deleted.）
https://github.com/liam-hq/liam/blob/18c0601042e61395a48ec638cb90990cff1da5c1/frontend/apps/app/components/SessionDetailPage/components/Chat/utils/timelineUtils.ts

so, I added unused variable detection to`frontend/apps/app/tsconfig.json`


frontend/apps/app/tsconfig.json
```diff
@@ -13,6 +13,8 @@
     "isolatedModules": true,
     "jsx": "preserve",
     "incremental": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
```
MH4GF, line 17: Question: does this mean the other packages are activated, or does it make sense to just add app?

FunamaYukina: I believe it is a good idea to activate these settings in other packages as well. I hope to address the remaining packages in the next pull request!

MH4GF: Thanks!

PR #2649, opened 2025-07-22
💄 Fix box-shadow clipping in SessionForm presenters

kumanoayumi:
## Issue

- resolve:

## Why is this change needed?

The box-shadow effects (blur) were being clipped due to overflow: hidden on SessionFormPresenter's `.formContainer`. Additionally, URLSessionFormPresenter was missing the loading/error state box-shadow effects that other presenters had, causing UI inconsistency.
This change fixes the clipping issue and ensures all form presenters have consistent visual feedback for loading and error states.


kumanoayumi:
Thank you 🚀

PR #2657, opened 2025-07-23
refactor(agent): separate system prompt and context in PMAnalysisAgent

MH4GF:
## Issue

- resolve: https://github.com/route06/liam-internal/issues/5106

## Why is this change needed?

PMAnalysisAgent was using `ChatPromptTemplate.fromMessages()` with chat history embedded in the system prompt, which mixed system instructions with user context. This change separates system prompt from context by using `BaseMessage[]` directly, similar to other agents like `databaseSchemaBuildAgent`.

## Summary

- Updated PMAnalysisAgent to accept `BaseMessage[]` directly instead of `BasePromptVariables`
- Removed string-based `chat_history` formatting in favor of native message handling  
- Eliminated redundant `analyzeRequirements()` method, keeping only `generate()`
- Updated `analyzeRequirementsNode` to pass `state.messages` directly
- Removed `ChatPromptTemplate` usage and manual prompt formatting
- Aligned with `databaseSchemaBuildAgent` pattern for consistency
- Made `requirementsAnalysisSchema` internal to fix lint warning

|before|after|
| --- | --- |
| image | image |

## Test plan

- [x] TypeScript type checking passes
- [x] All linting rules pass
- [x] Code formatting is consistent
- [x] Agent interface remains compatible with workflow

### DIFF
--- a/frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
+++ b/frontend/internal-packages/agent/src/chat/workflow/nodes/analyzeRequirementsNode.ts
@@ -3,10 +3,8 @@ import type { RunnableConfig } from '@langchain/core/runnables'
 import type { Database } from '@liam-hq/db'
 import { ResultAsync } from 'neverthrow'
 import { PMAnalysisAgent } from '../../../langchain/agents'
-import type { BasePromptVariables } from '../../../langchain/utils/types'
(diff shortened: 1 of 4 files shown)


NoritakaIkeda (approved):
I thought that separating prompts could improve maintainability!
Thanks! LGTM!

PR #2658, opened 2025-07-23
✨(workflow): Link DML queries and execution results with usecases

NoritakaIkeda:
## Issue

- resolve: N/A

## Why is this change needed?
This PR enhances the DML generation workflow to better track the relationship between usecases, their generated DML queries, and execution results. Previously, DML operations were not clearly associated with their originating usecases, making it difficult to trace which queries belonged to which business requirements.

### Key improvements:
- Each usecase can now have multiple DML operations with individual execution tracking
- WorkflowState stores DML execution results per operation
- Enhanced artifact output includes execution results alongside queries
- Comprehensive test coverage for all workflow transformations


NoritakaIkeda:
PoC-like confirmed, so close.

PR #2683, opened 2025-07-25
✨ feat: support shortcut commands

tnyo43:
## Issue

- resolve:


frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalette.tsx
```diff
@@ -12,8 +12,11 @@ import { type FC, useEffect } from 'react'
 import styles from './CommandPalette.module.css'
 import { CommandPaletteContent } from './CommandPaletteContent'
 import { useCommandPalette } from './CommandPaletteProvider'
+import { useSubscribeCommands } from './hooks'
 
 export const CommandPalette: FC = () => {
+  useSubscribeCommands()
```
tnyo43, line 18: It doesn't have to be subscribed here but I do it anyway.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/hooks/useSubscribeCommands/useSubscribeCopyLinkCommand.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect } from 'react'
+
+export const useSubscribeCopyLinkCommand = () => {
+  // Copy page link when ⌘C is pressed
+  useEffect(() => {
+    const down = (event: KeyboardEvent) => {
+      if (event.key === 'c' && (event.metaKey || event.ctrlKey)) {
+        // Copy the selected text instead if there is any
+        const selection = window.getSelection()
+        if (selection?.toString()) return
```
tnyo43, line 10: Stop copying on clipboard when anything is selected.
Actually, the default copy is executed after copying link to clipboard, so this lines doesn't affect the behaviour at the moment. However, this is necessary if we implement something for after copying clipboard in the future, such as showing toast message. Otherwise it shows the toast even thought link is not copied.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/hooks/useSubscribeCommands/useSubscribeSwitchShowMode.ts
```diff
@@ -0,0 +1,24 @@
+import { useEffect } from 'react'
+import { useUserEditingOrThrow } from '@/stores'
+
+export const useSubscribeShowModeCommand = () => {
+  const { setShowMode } = useUserEditingOrThrow()
+
+  // Switch show mode when ⇧2, ⇧3 or ⇧4 is pressed
+  useEffect(() => {
+    const down = (event: KeyboardEvent) => {
+      if (event.code === 'Digit2' && event.shiftKey) {
```
tnyo43, line 10: The `event.key` will be "@" with US keyboard when shift is pressed.
[`event.code`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) returns the physical keys, so the values are always same.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/hooks/useSubscribeCommands/useSubscribeTidyUpCommand.ts
```diff
@@ -0,0 +1,23 @@
...
+  const { getNodes, getEdges, setNodes } = useReactFlow()
+  const { fitView } = useCustomReactflow()
+
+  // Tidy up ERD when ⇧T is pressed
+  useEffect(() => {
+    const down = async (event: KeyboardEvent) => {
+      if (event.code === 'KeyT' && event.shiftKey) {
+        const { nodes } = await computeAutoLayout(getNodes(), getEdges())
+        setNodes(nodes)
+        fitView()
```
tnyo43, line 16: The steps are copied from here:
https://github.com/liam-hq/liam/blob/7df9f3615c2c5cd150d5b6a4e7a082e481084ccb/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/Toolbar/TidyUpButton/TidyUpButton.tsx#L40-L42

We could consider extracting the logic into a custom hook that these components/functions can refer to.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/hooks/useSubscribeCommands/useSubscribeSwitchShowMode.ts
```diff
@@ -0,0 +1,24 @@
...
+        setShowMode('ALL_FIELDS')
+      }
+      if (event.code === 'Digit3' && event.shiftKey) {
+        setShowMode('TABLE_NAME')
+      }
```
MH4GF, line 15: https://github.com/user-attachments/assets/a2eb7936-b0bd-4820-b6e1-1860f893631c

Shift+3 doesn't seem to be working. Could you please check?

MH4GF (commented):
LGTM, but check a comment!

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/hooks/useSubscribeCommands/useSubscribeSwitchShowMode.ts
tnyo43: Thanks for your report.
It works on my PC, so the issue may be caused by differences in the environment between us 🤔 
I’d like to clarify whether the problem is caused by ⇧3 or by `setShowMode('TABLE_NAME')`.

I added [some changes for investigation](https://github.com/liam-hq/liam/pull/2683/commits/9c4e145886cc85e26929b03541f8c3897db27f8c) and created another preview -> https://liam-rj464pdmm-liambx.vercel.app

The changes I added are:
- enable ⇧5 to set show mode as "TABLE_NAME", which is the same behaviour with ⇧3
- add console.logs when execute ⇧3 and ⇧5

Could you please try it with new preview and tell me how it works?

https://github.com/user-attachments/assets/0a40001e-73d7-442c-a0ba-2013186d79eb

MH4GF: @tnyo43 

Thanks. I noticed during the test that it worked in Incognito mode, so it was probably due to an extension conflict in my environment 🙏🏻 There is probably nothing wrong with the implementation. Thanks!

In case you are interested, I am attaching the video I took, Shift + 5 was working.


https://github.com/user-attachments/assets/ece5fa95-dfa8-4aa0-b3d2-ad100f958806

PR #2690, opened 2025-07-25
♻️ Extract WebSearchAgent from webSearchNode to agent package

FunamaYukina:
## Issue

- resolve:

## Why is this change needed?

This is a pure refactoring to align `webSearchNode` with our established agent pattern. No functionality has been changed - the web search logic has simply been moved from the node file into a dedicated `WebSearchAgent` class for better code organization.

### DIFF
--- a/frontend/internal-packages/agent/src/chat/workflow/nodes/webSearchNode.ts
+++ b/frontend/internal-packages/agent/src/chat/workflow/nodes/webSearchNode.ts
@@ -1,8 +1,8 @@
-import { AIMessage, SystemMessage } from '@langchain/core/messages'
+import { AIMessage } from '@langchain/core/messages'
 import type { RunnableConfig } from '@langchain/core/runnables'
-import { ChatOpenAI } from '@langchain/openai'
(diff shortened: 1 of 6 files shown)


MH4GF:
@FunamaYukina 
Frankly, I do not agree with this modification. I can't think of any benefit to splitting it into classes, and it just seems to increase the number of lines of code by a factor of 2.5 and complicate things. What do you think?

FunamaYukina:
@MH4GF Oh, I  just thought it would be a good idea to unify the agent information in the agent package. But you are right that the amount of code will increase.It seems unlikely that this agent will be reused on another node.

I will quit this change and try to imagine this agent as a single tool!

MH4GF:
@FunamaYukina Thanks. It might be nice to cut it out when I want to reuse it from another node, but the prompts will probably be different. There's not much reason to make it a class either, so I want to avoid that.

FunamaYukina:
As you say, the prompts are going to be different even when reused🤔
Thank you for telling me about it!!

PR #2730, opened 2025-07-28
✨ Add reasoning summary logging in analyze requirements node

junkisai:
## Issue

- resolve: N/A

## Why is this change needed?
This change adds logging for AI reasoning summaries in the analyze requirements node. When the PM Analysis Agent generates reasoning summaries, they are now logged as assistant messages before the main analysis result, providing better transparency into the AI's decision-making process.

スクリーンショット 2025-07-28 15 18 18

### DIFF
--- a/frontend/internal-packages/agent/src/langchain/agents/pmAnalysisAgent/agent.ts
+++ b/frontend/internal-packages/agent/src/langchain/agents/pmAnalysisAgent/agent.ts
@@ -1,38 +1,116 @@
 import { type BaseMessage, SystemMessage } from '@langchain/core/messages'
+import type { Runnable } from '@langchain/core/runnables'
 import { ChatOpenAI } from '@langchain/openai'
-import { toJsonSchema } from '@valibot/to-json-schema'
 import * as v from 'valibot'
 import { PM_ANALYSIS_SYSTEM_MESSAGE } from './prompts'
 
-const requirementsAnalysisSchema = v.object({
+// Direct JsonSchema definition instead of using toJsonSchema
+// because the generated schema has subtle incompatibilities with withStructuredOutput
+// (specifically the properties:{}, required:[] structure).
+// TODO: Migrate from valibot to zod, which is officially supported by langchain
+const REQUIREMENTS_ANALYSIS_SCHEMA = {
+  title: 'RequirementsAnalysis',
+  type: 'object',
+  properties: {
+    businessRequirement: { type: 'string' },
+    functionalRequirements: {
+      type: 'object',
+      properties: {},
+      required: [],
+      additionalProperties: {
+        type: 'array',
+        items: { type: 'string' },
+      },
+    },
+    nonFunctionalRequirements: {
+      type: 'object',
+      properties: {},
+      required: [],
+      additionalProperties: {
+        type: 'array',
+        items: { type: 'string' },
+      },
+    },
+  },
+  required: [
+    'businessRequirement',
+    'functionalRequirements',
+    'nonFunctionalRequirements',
+  ],
+  additionalProperties: false,
+}
+
+const requirementsAnalysisSchema = v.strictObject({
   businessRequirement: v.string(),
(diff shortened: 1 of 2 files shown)


frontend/internal-packages/agent/src/langchain/agents/pmAnalysisAgent/agent.ts
```diff
@@ -1,38 +1,116 @@
...
-    // Convert valibot schema to JSON Schema and bind to model
-    const analysisJsonSchema = toJsonSchema(requirementsAnalysisSchema)
-
-    this.analysisModel = baseModel.withStructuredOutput(analysisJsonSchema)
+    this.analysisModel = baseModel.withStructuredOutput<RunOutput>(
+      REQUIREMENTS_ANALYSIS_SCHEMA,
+      {
+        includeRaw: true,
+      },
+    )
```
NoritakaIkeda, line 91: Q: Do you understand this as a localized fix specific to this part of the code? I believe we’re still using toJsonSchema in the QA agent, so I’m considering whether we should adopt this as a project-wide standard instead.

> because the generated schema has subtle incompatibilities with withStructuredOutput

If this is indeed the root cause, and the current approach tends to produce vague or incomplete properties in the schema, then I’d like to treat the updated implementation as the correct one and standardize it across the project.

frontend/internal-packages/agent/src/langchain/agents/pmAnalysisAgent/agent.ts
```diff
@@ -1,38 +1,116 @@
...
+    {
+      raw: BaseMessage
+      parsed: RunOutput
+    }
+  >
 
   constructor() {
     const baseModel = new ChatOpenAI({
       model: 'o4-mini',
+      reasoning: { effort: 'high', summary: 'detailed' },
```
MH4GF, line 82: Effort should be tuned depending on the application, but I thought that should be done only after the bench markers are ready 👍🏻

frontend/internal-packages/agent/src/langchain/agents/pmAnalysisAgent/agent.ts
```diff
@@ -1,38 +1,116 @@
...
+import type { Runnable } from '@langchain/core/runnables'
 import { ChatOpenAI } from '@langchain/openai'
-import { toJsonSchema } from '@valibot/to-json-schema'
 import * as v from 'valibot'
 import { PM_ANALYSIS_SYSTEM_MESSAGE } from './prompts'
 
-const requirementsAnalysisSchema = v.object({
+// Direct JsonSchema definition instead of using toJsonSchema
+// because the generated schema has subtle incompatibilities with withStructuredOutput
+// (specifically the properties:{}, required:[] structure).
```
MH4GF, line 9: I see... If you can afford it, you can make an issue on valibot!

MH4GF (approved):
LGTM 👍🏻 Cool!!

frontend/internal-packages/agent/src/langchain/agents/pmAnalysisAgent/agent.ts
junkisai: @NoritakaIkeda 
In my personal opinion, I think we could switch to Zod for the agent package, as it's officially supported by LangChain. However, I've been hesitant to make the switch without considering compatibility with other packages.

It seems we'll need some time to investigate and plan, especially regarding the details of the issues that have arisen, our migration strategy, and commonization. 
I'd like to create an issue to discuss these points there. 🙏

frontend/internal-packages/agent/src/langchain/agents/pmAnalysisAgent/agent.ts
junkisai: I'm not that familiar with the JSON Schema specification, so I've decided to open a discussion to get some opinions for now!
https://github.com/fabian-hiller/valibot/discussions/1263

PR #2735, opened 2025-07-28
🎨 Update agent avatars to use neutral color schemes with Jack character

kumanoayumi:
## Issue

- resolve:

## Why is this change needed?

The current agent avatars use single colors (red for Error, green for Success, yellow for Warning) that unintentionally convey status meanings rather than agent roles. This change introduces Jack character illustrations with composite color schemes and circular backgrounds similar to  UserAvatar, creating a more neutral and cohesive visual design that doesn't imply system states.

スクリーンショット 2025-07-28 18 45 48

スクリーンショット 2025-07-28 18 54 51

スクリーンショット 2025-07-28 18 54 56


kumanoayumi:
Thanks!! 🚀

PR #2756, opened 2025-07-29
♻️ Refactor SendButton to use IconButton component

junkisai:
## Issue

- resolve: N/A

## Why is this change needed?
The previous SendButton implementation had several issues:
1. **Shape inconsistency**: The button could become rectangular in certain conditions due to lack of fixed dimensions
2. **Code duplication**: Custom tooltip implementation when IconButton already provides this functionality
3. **Maintainability**: Multiple style states managed manually instead of leveraging existing UI components

This refactoring simplifies the SendButton component by leveraging the existing IconButton component from the UI library. This ensures consistent circular button shape and reduces code duplication.

スクリーンショット 2025-07-29 18 41 26

### Changes:
- Replaced custom Button implementation with IconButton component (which has fixed 1.5rem × 1.5rem dimensions)
- Simplified CSS by removing redundant styles and states
- Enhanced IconButton to properly handle disabled state with tooltips
- Fixed cursor display for disabled buttons

## Test

I've confirmed that the UI behavior remains unchanged before and after the modification (with the rare exception of it becoming rectangular).

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/13c82270-79c3-4189-8c57-4bf121cfc15d" /> | <video src="https://github.com/user-attachments/assets/5ef4e29f-5f65-4502-8f71-e9937ed6f119" /> |

### DIFF
--- a/frontend/apps/app/components/SessionDetailPage/components/Chat/components/ChatInput/components/SendButton/SendButton.module.css
+++ b/frontend/apps/app/components/SessionDetailPage/components/Chat/components/ChatInput/components/SendButton/SendButton.module.css
@@ -7,80 +7,34 @@
   justify-content: center;
   border: var(--border-width-base) solid transparent;
   border-radius: var(--border-radius-base);
+  background-color: var(--primary-overlay-40);
   transition: background-color 0.3s var(--default-timing-function);
 }
 
-.default {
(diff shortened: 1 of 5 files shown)


MH4GF (approved):
👍🏻

PR #2768, opened 2025-07-30
🐛 Fix hardcoded schema filename display in GitHubSessionFormPresenter

junkisai:
## Issue

- resolve: N/A

## Why is this change needed?
The schema filename was hardcoded as "database.sql" in the GitHubSessionFormPresenter component, regardless of the actual schema file in the repository. This change dynamically fetches and displays the correct schema filename from the database.

## Summary
- Added `getSchemaFilePath` action to fetch the actual schema file path from the database
- Updated SessionForm components to pass the schema file path through props
- Modified GitHubSessionFormPresenter to display the actual filename instead of hardcoded "database.sql"
- Updated all Storybook stories to include the new `schemaFilePath` prop
- Fixed import formatting in processRepositoryAnalysis.ts

## Check

https://github.com/user-attachments/assets/126f902e-ed56-448d-a3b0-4fdd36229114

## Summary by CodeRabbit

* **New Features**
  * Added support for displaying dynamic schema file names in session forms, based on the selected project.
  * Session forms now fetch and display the schema file path relevant to the chosen project.

* **Bug Fixes**
  * Storybook stories updated to support and test the new schema file path feature.

* **Style**
  * Improved formatting and import statements for better code clarity and maintainability.

### DIFF
--- a/frontend/apps/app/features/sessions/components/SessionForm/GitHubSessionFormPresenter/GitHubSessionFormPresenter.tsx
+++ b/frontend/apps/app/features/sessions/components/SessionForm/GitHubSessionFormPresenter/GitHubSessionFormPresenter.tsx
@@ -24,6 +24,7 @@ type Props = {
   onProjectChange: (projectId: string) => void
   formAction: (formData: FormData) => void
   isTransitioning?: boolean
+  schemaFilePath: string | null
 }
 
 export const GitHubSessionFormPresenter: FC<Props> = ({
@@ -37,6 +38,7 @@ export const GitHubSessionFormPresenter: FC<Props> = ({
   onProjectChange,
   formAction,
   isTransitioning = false,
+  schemaFilePath,
 }) => {
   const textareaRef = useRef<HTMLTextAreaElement>(null)
   const formRef = useRef<HTMLFormElement>(null)
@@ -196,7 +198,13 @@ export const GitHubSessionFormPresenter: FC<Props> = ({
                     disabled={isPending}
                     isLoading={isBranchesLoading}
                   />
-                  <SchemaDisplay schemaName="database.sql" />
+                  <SchemaDisplay
+                    schemaName={
+                      schemaFilePath
+                        ? schemaFilePath.split('/').pop() || 'database.sql'
+                        : 'database.sql'
+                    }
+                  />
                 </>
               )}
             </div>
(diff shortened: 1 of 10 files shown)


frontend/apps/app/features/sessions/components/SessionForm/actions/getSchemaFilePath/getSchemaFilePath.ts
```diff
@@ -0,0 +1,28 @@
+'use server'
+
+import { fetchSchemaFilePath } from './fetchSchemaFilePath'
+
+type SchemaFilePathState = {
+  path: string | null
+  loading: boolean
```
MH4GF: Always false, do we need to return this?

frontend/apps/app/features/sessions/components/SessionForm/GitHubSessionFormPresenter/GitHubSessionFormPresenter.tsx
```diff
@@ -196,7 +198,13 @@ export const GitHubSessionFormPresenter: FC<Props> = ({
                     disabled={isPending}
                     isLoading={isBranchesLoading}
                   />
-                  <SchemaDisplay schemaName="database.sql" />
+                  <SchemaDisplay
+                    schemaName={
+                      schemaFilePath
+                        ? schemaFilePath.split('/').pop() || 'database.sql'
+                        : 'database.sql'
+                    }
```
MH4GF, line 206: I would like to sort out the behavior when the schemaFilePath does not exist. In this case, the project is linked but there is no schema, so the design should start empty.
In that case, it would be better to induce the schema file path setting. What do you think?

frontend/apps/app/features/sessions/components/SessionForm/actions/getSchemaFilePath/getSchemaFilePath.ts
junkisai: As the comment stated, **loading** was always returning **false**, so it was unnecessary.

[🔥 Remove unused loading field from getSchemaFilePath](https://github.com/liam-hq/liam/pull/2768/commits/df69841339a2156c42ef573d6ce01d9e8e8564ec)

frontend/apps/app/features/sessions/components/SessionForm/GitHubSessionFormPresenter/GitHubSessionFormPresenter.tsx
junkisai: As you pointed out, when a user wants to link a project, they likely intend to perform database design on an existing schema. 💭 

I want a link to guide settings to be displayed when the project is linked and `schemaFilePath` is null.
I'll create a separate issue for this and handle it there!

MH4GF: Yes, I think that is a good idea!

MH4GF (approved):
LGTM 👍🏻

PR #2793, opened 2025-07-31
Feat/toast command palette mode

tnyo43:
## Issue

- resolve:


tnyo43:
We complete the trial. This change is supervised by https://github.com/liam-hq/liam/pull/2942

PR #2798, opened 2025-07-31
🐛(agent): Remove reasoning field from AIMessages instead of filtering

NoritakaIkeda:
## Issue

- resolve: AIMessage reasoning field causing API errors when invoking design agent
- Alternative approach to message filtering that preserves context

## Why is this change needed?

This change addresses a validation issue with the OpenAI API when processing messages that include a malformed reasoning field inside AIMessage.additional_kwargs.

In our previous implementation, some AIMessage objects retained a reasoning field containing improperly structured data. When such messages were passed to the OpenAI API (e.g., via LangChain agents), they triggered validation errors — specifically, errors like "reasoning without required following item". This was due to the OpenAI API's strict schema expectations for system-level metadata, which reasoning accidentally violated.

To prevent this, we now filter out the reasoning field from all AIMessage instances before invoking the agent. This ensures all messages conform to the API's expected structure, avoiding runtime failures. Notably, HumanMessage objects are unaffected and safely passed through without modification.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue that could cause API errors by ensuring the `reasoning` field is removed from AI-generated messages before processing.
  * Improved error handling by displaying error feedback as a user message, preventing related API errors.

### DIFF
--- a/frontend/internal-packages/agent/src/chat/workflow/nodes/designSchemaNode.ts
+++ b/frontend/internal-packages/agent/src/chat/workflow/nodes/designSchemaNode.ts
@@ -68,7 +68,7 @@ export async function designSchemaNode(
     // Return state with error feedback as HumanMessage for self-recovery
     return {
       ...state,
-      messages: [...state.messages, errorFeedbackMessage],
+      messages: [errorFeedbackMessage],
       error: invokeResult.error,
     }
   }
@@ -98,7 +98,7 @@ export async function designSchemaNode(
 
   return {
     ...state,
-    messages: [...state.messages, syncedMessage],
+    messages: [syncedMessage],
     latestVersionNumber: state.latestVersionNumber + 1,
   }
 }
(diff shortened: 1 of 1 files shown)


frontend/internal-packages/agent/src/chat/workflow/nodes/designSchemaNode.ts
```diff
@@ -38,14 +59,16 @@ export async function designSchemaNode(
...
+    // Create a human message for error feedback to avoid reasoning API issues
+    // Using HumanMessage prevents the "reasoning without required following item" error
+    const errorFeedbackMessage = new HumanMessage({
+      content: `The previous attempt failed with the following error: ${invokeResult.error.message}. Please try a different approach to resolve the issue.`,
+    })
+
+    // Return state with error feedback as HumanMessage for self-recovery
     return {
       ...state,
+      messages: [errorFeedbackMessage],
```
MH4GF, line 71: I was wondering why this is able to retry this one, did you find this one?

NoritakaIkeda: As we previously discussed, this may need to be replaced with a command. We'll investigate and prioritize this in a separate issue. 🙏

PR #2808, opened 2025-07-31
🐛(agent): Fix DDL generation in design workflow - DDL statements not being passed to validateSchemaNode

NoritakaIkeda:
## Issue

- resolve: https://github.com/route06/liam-internal/issues/5197

## Why is this change needed?

The design workflow was failing with "No database structure to build" error because DDL statements were generated but never stored in the workflow state. The `validateSchemaNode` expects `state.ddlStatements` to be populated, but it was always undefined.

This fix ensures that DDL statements are generated from the updated schema and stored in the workflow state after successful schema design operations, allowing the validation node to properly execute and validate the database structure.

## Summary by CodeRabbit

* **New Features**
  * After using the schema design tool, users can now view the generated DDL statements for the updated schema.

### DIFF
--- a/frontend/internal-packages/agent/src/db-agent/nodes/invokeSchemaDesignToolNode.ts
+++ b/frontend/internal-packages/agent/src/db-agent/nodes/invokeSchemaDesignToolNode.ts
@@ -2,7 +2,7 @@ import type { BaseMessage } from '@langchain/core/messages'
 import { ToolMessage } from '@langchain/core/messages'
 import type { RunnableConfig } from '@langchain/core/runnables'
 import { ToolNode } from '@langchain/langgraph/prebuilt'
-import type { Schema } from '@liam-hq/db-structure'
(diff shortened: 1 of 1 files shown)


pnpm-lock.yaml
NoritakaIkeda, line 1: **Reason for the changes:**

1. A lint error occurred during the initial commit.
2. Error details:
   `components/SessionDetailPage/components/Output/components/Artifact/Artifact.tsx(15,23): error TS2307: Cannot find module 'rehype-raw' or its corresponding type declarations.`
3. The TypeScript error indicated that the `rehype-raw` package could not be found.
4. I ran the command `pnpm add rehype-raw` to add the missing dependency.
5. When `pnpm` adds a dependency, it automatically updates `pnpm-lock.yaml`.

NoritakaIkeda: Might not work right, I'll fix it.

hoshinotsuyoshi:
I’m still reviewing this, but feedback from anyone would be greatly appreciated!

PR #2813, opened 2025-08-01
🤖 Remove GitHub Actions updates from dependabot configuration

hoshinotsuyoshi:
## Issue

- resolve: Remove monthly GitHub Actions dependency updates

## Why is this change needed?
This change removes the GitHub Actions package ecosystem from dependabot.yml to stop automatic monthly dependency updates for GitHub Actions. This helps reduce maintenance overhead and prevents potential issues from automatic GitHub Actions updates.

## Summary by CodeRabbit

* **Chores**
  * Removed automatic update checks for GitHub Actions dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,9 +2,3 @@
 
 version: 2
 updates:
-  - package-ecosystem: "github-actions"
-    directory: "/"
-    schedule:
-      interval: "monthly"
-      time: "10:00"
-      timezone: "Asia/Tokyo"


MH4GF (approved):
This is fine, but what do you think about deleting the entire file?

hoshinotsuyoshi:
I was a bit unsure, but I felt keeping the file would make it easier to track the history.

MH4GF:
> keeping the file would make it easier to track the history.

That's right. Thank you! 😄

hoshinotsuyoshi:
oh, sorry, it seems that `updates:` does not allow `null` value.  https://github.com/liam-hq/liam/runs/47169811568  🙏 

I'll fix it

PR #2823, opened 2025-08-01
🐛(agent): Fix tool message ordering to prevent sync errors

NoritakaIkeda:
## Issue

- resolve: https://github.com/route06/liam-internal/issues/5201

スクリーンショット 2025-08-01 18 04 12

## Why is this change needed?

The issue was caused by accidentally removing necessary fields when attempting to
  remove the reasoning field.

  The original code only extracted three properties (content, additional_kwargs, 
  response_metadata) from the AIMessage to create a new instance, which
  unintentionally discarded important fields like `tool_calls`.

  This fix ensures that only the reasoning field is removed while preserving all
  other fields including `tool_calls`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of AI-generated messages to ensure all relevant metadata is preserved during processing.

### DIFF
--- a/frontend/internal-packages/agent/src/chat/workflow/nodes/designSchemaNode.ts
+++ b/frontend/internal-packages/agent/src/chat/workflow/nodes/designSchemaNode.ts
@@ -96,7 +96,7 @@ export async function designSchemaNode(
     // Return state with error feedback as HumanMessage for self-recovery
     return {
       ...state,
-      messages: [...messages, errorFeedbackMessage],
+      messages: [errorFeedbackMessage],
       error: invokeResult.error,
     }
   }
@@ -126,7 +126,7 @@ export async function designSchemaNode(
 
   return {
     ...state,
-    messages: [...messages, syncedMessage],
+    messages: [syncedMessage],
     latestVersionNumber: state.latestVersionNumber + 1,
   }
 }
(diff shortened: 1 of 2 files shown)


frontend/internal-packages/agent/src/chat/workflow/nodes/designSchemaNode.ts
```diff
@@ -34,19 +34,47 @@ export async function designSchemaNode(
...
+        aiMessageFields.tool_calls = tool_calls
+      }
+      if (invalid_tool_calls !== undefined) {
+        aiMessageFields.invalid_tool_calls = invalid_tool_calls
+      }
+      if (usage_metadata !== undefined) {
+        aiMessageFields.usage_metadata = usage_metadata
+      }
+
+      return new AIMessage(aiMessageFields)
```
NoritakaIkeda, line 77: The issue was caused by accidentally removing necessary fields when attempting to
  remove the reasoning field.

  The original code only extracted three properties (content, additional_kwargs, 
  response_metadata) from the AIMessage to create a new instance, which
  unintentionally discarded important fields like `tool_calls`.

  This fix ensures that only the reasoning field is removed while preserving all
  other fields including `tool_calls`.

MH4GF (approved):
It could be written a little better, but for once it looked good. Thank you very much!

PR #2835, opened 2025-08-01
fix(db-structure): handle Prisma views with @id directives

hoshinotsuyoshi:
## Issue

- resolve: #2835 (Prisma schema parsing error for views with @id directives)

## Why is this change needed?

Prisma views cannot have any index constraints (@id, @unique, @@unique), but some schemas like langfuse's include them. This was causing "Views cannot have primary keys" and "Views cannot have indexes" validation errors when parsing these schemas, preventing ER diagram generation.

Additionally, the original regex implementation had a ReDoS vulnerability identified by GitHub security scanning.


hoshinotsuyoshi:
https://github.com/liam-hq/liam/pull/2857 resolves that issue

PR #2859, opened 2025-08-04
Feat/add css modules kit

sakit0:
## Issue

- resolve: Add ESLint rules for CSS Modules

## Why is this change needed?
We need to add `@css-modules-kit/eslint-plugin` to detect undefined class names when using CSS Modules. This improves type safety for CSS Modules and helps catch errors early in development.

## What was done
・add `@css-modules-kit/eslint-plugin` & `@eslint/css`
・There are many errors, so we use eslint-suppressions to avoid them. First of all, we will prioritize the introduction of Lint.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -29,6 +29,7 @@
       "source.fixAll.stylelint": "always"
     }
   },
+  "eslint.validate": ["css"],
   "prettier.requireConfig": false,
   "eslint.workingDirectories": [
     {
--- a/frontend/internal-packages/configs/eslint/base.js
+++ b/frontend/internal-packages/configs/eslint/base.js
@@ -6,6 +6,8 @@ import { noNonEnglishPlugin } from './no-non-english-plugin.js'
 import { noThrowErrorPlugin } from './no-throw-error-plugin.js'
 import { preferClsxPlugin } from './prefer-clsx-plugin.js'
 import { requireUseServerPlugin } from './require-use-server-plugin.js'
+import css from '@eslint/css';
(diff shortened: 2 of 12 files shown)


.vscode/settings.json
```diff
@@ -29,6 +29,7 @@
       "source.fixAll.stylelint": "always"
     }
   },
+  "eslint.validate": ["css"],
```
MH4GF, line 32: 👍🏻

MH4GF (approved):
LGTM 👍🏻 
Please resolve the conflicts and merge.

sakit0:
@MH4GF 
I only enabled the [css-modules-kit/no-unused-class-names](https://github.com/mizdra/css-modules-kit/blob/main/packages/eslint-plugin/docs/rules/no-unused-class-names.md) setting because [css-modules-kit/no-missing-component-file ](https://github.com/mizdra/css-modules-kit/blob/main/packages/eslint-plugin/docs/rules/no-missing-component-file.md)only caused errors in the CI. I believe this configuration still achieves the goal of the issue. I'll address the css-modules-kit/no-missing-component-file issue separately.

frontend/internal-packages/configs/eslint/base.js
MH4GF, line 1: It seems that the format by biome defined in the project is not working, but it is a trivial problem. It would be better to merge them first.

MH4GF (approved):
Thanks!! 🎉

MH4GF:
@sakit0 Ok that's the way to proceed, no problem. Thanks!

PR #2874, opened 2025-08-04
⚡(erd-core): Filter unchanged diff items in SchemaProvider

junkisai:
## Issue

- resolve: N/A

## Why is this change needed?
This change improves performance by filtering out unchanged items from the diffItems array in SchemaProvider. Since unchanged items don't affect the UI rendering, excluding them reduces unnecessary data processing and improves rendering performance, especially for schemas with many tables and columns.

## Summary by CodeRabbit

* **New Features**
  * Schema diff view now hides unchanged items, making it easier to focus on relevant changes.

### DIFF
--- a/frontend/packages/erd-core/src/stores/schema/SchemaProvider.tsx
+++ b/frontend/packages/erd-core/src/stores/schema/SchemaProvider.tsx
@@ -23,13 +23,16 @@ export const SchemaProvider: FC<Props> = ({ children, current, previous }) => {
       tables: {},
     }
     const diffItems = buildSchemaDiff(previous ?? emptySchema, current)
+    const filteredDiffItems = diffItems.filter(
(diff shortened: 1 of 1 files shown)


frontend/packages/erd-core/src/stores/schema/SchemaProvider.tsx
```diff
@@ -23,13 +23,16 @@ export const SchemaProvider: FC<Props> = ({ children, current, previous }) => {
       tables: {},
     }
     const diffItems = buildSchemaDiff(previous ?? emptySchema, current)
+    const filteredDiffItems = diffItems.filter(
+      (item) => item.status !== 'unchanged',
+    )
```
MH4GF, line 28: I could always filter it out in buildSchemaDiff if it's not needed, but once it's done, I guess I'll just leave it as is.

PR #2878, opened 2025-08-04
✨(erd-core): Add diff highlighting for table column properties

junkisai:
## Issue

- resolve: #1337

## Why is this change needed?

This PR adds diff highlighting support for table column properties in the ERD visualization to improve the visibility of schema changes. Users can now easily see what properties have been added, removed, or modified for each column.

### Changes included:
- Extract column property components (Type, Default, NotNull, PrimaryKey, Comment) into separate components
- Add diff highlighting logic for each property type with proper change detection
- Export diff types from db-structure package for use in erd-core
- Implement visual indicators for added/removed/modified properties using consistent styling

### Visual improvements:
- Added properties show green background
- Removed properties show red background with strikethrough
- Modified properties show yellow background
- Consistent diff styling across all column properties

### DIFF
--- a/frontend/packages/db-structure/src/diff/index.ts
+++ b/frontend/packages/db-structure/src/diff/index.ts
@@ -8,7 +8,9 @@ export type {
   TableRelatedDiffItem,
 } from './types.js'
 export {
+  columnDiffItemSchema,
   columnRelatedDiffItemSchema,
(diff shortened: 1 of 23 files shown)


MH4GF (approved):
👍🏻

PR #2881, opened 2025-08-04
🐛 Fix OpenAI reasoning token error in qaGenerateUsecaseAgent

MH4GF:
## Issue

- resolve: https://github.com/route06/liam-internal/issues/5217

## Why is this change needed?

The qaGenerateUsecaseAgent was experiencing OpenAI API errors when processing messages containing reasoning tokens from o4-mini model responses. The error "reasoning without required following item" prevented the usecase generation workflow from completing successfully.

This fix implements a messageCleanup utility that safely removes the reasoning field from AIMessages while preserving essential properties like tool_calls, ensuring compatibility with subsequent OpenAI API calls.

### DIFF
--- a/frontend/internal-packages/agent/src/langchain/agents/qaGenerateUsecaseAgent/agent.ts
+++ b/frontend/internal-packages/agent/src/langchain/agents/qaGenerateUsecaseAgent/agent.ts
@@ -45,7 +45,16 @@ const USECASE_GENERATION_SCHEMA = {
   additionalProperties: false,
 }
 
-// Single usecase schema
+// Schema for usecase from OpenAI response (without id and dmlOperations)
+const usecaseFromApiSchema = v.object({
+  requirementType: v.picklist(['functional', 'non_functional']), // Type of requirement
+  requirementCategory: v.string(), // Category of the requirement
+  requirement: v.string(), // Content/text of the specific requirement
+  title: v.string(),
+  description: v.string(),
+})
+
+// Complete usecase schema with id and dmlOperations (for final output)
 const usecaseSchema = v.object({
   id: v.pipe(v.string(), v.uuid()), // UUID
   requirementType: v.picklist(['functional', 'non_functional']), // Type of requirement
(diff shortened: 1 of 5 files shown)


frontend/internal-packages/agent/src/langchain/agents/qaGenerateUsecaseAgent/agent.ts
```diff
@@ -45,7 +45,16 @@ const USECASE_GENERATION_SCHEMA = {
...
 
-// Single usecase schema
+// Schema for usecase from OpenAI response (without id and dmlOperations)
+const usecaseFromApiSchema = v.object({
+  requirementType: v.picklist(['functional', 'non_functional']), // Type of requirement
+  requirementCategory: v.string(), // Category of the requirement
+  requirement: v.string(), // Content/text of the specific requirement
+  title: v.string(),
+  description: v.string(),
+})
```
MH4GF, line 55: The id and dmlOperations are not being generated right now. when I solved the reasoning problem, I got an error in schema validation and could not solve it. Apparently it was hiding in the reasoning issue. Perhaps it should be implemented in https://github.com/liam-hq/liam/issues/2709, but I don't know the status right now. Will discuss with @NoritakaIkeda tomorrow.

frontend/internal-packages/agent/src/repositories/InMemoryRepository.ts
```diff
@@ -57,6 +57,8 @@ type InMemoryRepositoryOptions = {
 
 export class InMemoryRepository implements SchemaRepository {
   private state: InMemoryRepositoryState
+  // Used by generateId method
+  // biome-ignore lint/correctness/noUnusedPrivateClassMembers: used by generateId method
```
hoshinotsuyoshi, line 61: 👍

MH4GF:
I can't share the URL because I couldn't login in the preview environment, but I confirmed that all workflows can be passed locally 🎉.

Pasted_Image_2025_08_04__18_04

hoshinotsuyoshi (approved):
👍

PR #2901, opened 2025-08-05
feat: Update QA Agent to generate id and dmlOperations fields

NoritakaIkeda:
## Issue

- resolve: #2881

## Why is this change needed?

 The applyPatchOperations function is immutable - it doesn't modify the original object but returns
   a new one with the patches applied. The previous code was calling this function but ignoring its
  return value, which meant the currentSchema object remained empty ({ tables: {}, enums: {} }).

  This caused a cascade of failures:
  1. The schema remained empty after applying patches
  2. The DDL generator (postgresqlSchemaDeparser) received an empty schema with 0 tables
  3. No DDL statements were generated (ddlStatements became undefined)
  4. The DML node couldn't proceed without DDL statements

  By capturing the return value and updating currentSchema with Object.assign, we ensure the schema
  is properly built with all the applied patches, allowing DDL generation and the rest of the
  workflow to function correctly

## Summary by CodeRabbit

- New Features
  - DML generation now includes an empty dml_execution_logs field on every operation for consistent output.

- Bug Fixes
  - More accurate defaulting preserves existing non-nullish values for dml_execution_logs.
  - Safer schema building applies patches only when successful, reducing risk of partial or invalid updates.

- Documentation
  - Prompts and guidelines updated to mandate dml_execution_logs in DML operation outputs.

- Tests
  - Added unit test to validate DML operation structure, including dml_execution_logs.

- Chores
  - Minor logging wording and comment updates; removed a redundant warning.


NoritakaIkeda:
@FunamaYukina 
We're currently working on the issue where the DML is not appearing in the artifacts. If this requires urgent attention, I'd appreciate it if you could take it over. Otherwise, if there are any parts that can be prioritized, I’d be happy to handle them on Tuesday.

frontend/internal-packages/agent/src/repositories/supabase.ts
```diff
@@ -189,7 +189,16 @@ export class SupabaseSchemaRepository implements SchemaRepository {
...
-        applyPatchOperations(currentSchema, patchParsed.output)
+        const patchResult = applyPatchOperations(
+          currentSchema,
+          patchParsed.output,
+        )
+        if (patchResult.isOk()) {
+          // Update currentSchema with the patched result
+          Object.assign(currentSchema, patchResult.value)
+        } else {
+          // Failed to apply patch for this version, continue with next
```
NoritakaIkeda, line 200: An immutable function does not modify the original object but returns a new one, so you must always capture and use its return value.

ref: 
https://github.com/liam-hq/liam/blob/ab507ace76cc79ccd6f6abccc06c41e19db33a8b/frontend/internal-packages/agent/src/repositories/supabase.ts#L111

frontend/internal-packages/agent/src/db-agent/nodes/invokeSchemaDesignToolNode.ts
```diff
@@ -104,8 +104,14 @@ export const invokeSchemaDesignToolNode = async (
     if (schemaResult.isOk()) {
       // Generate DDL statements from the updated schema
       const ddlResult = postgresqlSchemaDeparser(schemaResult.value.schema)
+
+      // Check for DDL generation errors without logging
+      // Errors are handled by returning undefined ddlStatements
+
       const ddlStatements =
-        ddlResult.errors.length > 0 ? undefined : ddlResult.value
+        ddlResult.errors && ddlResult.errors.length > 0
```
MH4GF: Does `ddlResult.errors` ever return a falsy value? Was the following not good enough?

```suggestion
        ddlResult.errors.length > 0
```

frontend/internal-packages/artifact/src/schemas/artifact.ts
```diff
@@ -13,7 +13,7 @@ export const dmlOperationSchema = v.object({
   operation_type: v.picklist(['INSERT', 'UPDATE', 'DELETE', 'SELECT']),
   sql: v.string(),
   description: v.optional(v.string()),
-  dml_execution_logs: v.array(dmlExecutionLogSchema),
+  dml_execution_logs: v.optional(v.array(dmlExecutionLogSchema)),
```
MH4GF: Can't it be an empty array? I want to reduce as much as possible the optional.

frontend/internal-packages/agent/src/db-agent/nodes/invokeSchemaDesignToolNode.ts
NoritakaIkeda: Thank you. It turned out to be unnecessary, so I reverted it on my end.
[refactor: Address PR review comments for dml_execution_logs field](https://github.com/liam-hq/liam/pull/2901/commits/4e6a8066aaee46340dbe310e671fad34e877c7a6)

frontend/internal-packages/artifact/src/schemas/artifact.ts
NoritakaIkeda: At the time the SQL is written to the artifact, there’s nothing in the execution results, so I had made it optional. However, I’ve now changed it so that when the SQL (DML) is generated, this field is passed as an empty array 🙏
I’ve confirmed that it works with this change.

[refactor: Address PR review comments for dml_execution_logs field](https://github.com/liam-hq/liam/pull/2901/commits/4e6a8066aaee46340dbe310e671fad34e877c7a6)
[fix: Add dml_execution_logs to DMLGenerationAgent output](https://github.com/liam-hq/liam/pull/2901/commits/4af63b0d17c7eb1e4f3bb2bded085c66978e7ce3)

frontend/internal-packages/agent/src/langchain/agents/dmlGenerationAgent/agent.test.ts
```diff
@@ -29,4 +29,32 @@ describe('DMLGenerationAgent', () => {
...
+      sql: expect.any(String),
+      description: expect.any(String),
+      dml_execution_logs: expect.arrayContaining([]),
+    }
+
+    // Verify our mock structure matches what we expect
+    const firstOperation = mockDmlOperations[0]
+    expect(firstOperation).toMatchObject(expectedStructure)
+    expect(firstOperation?.dml_execution_logs).toEqual([])
+  })
```
MH4GF: This test does not seem to verify anything. Could you please verify this?

NoritakaIkeda: This part of the code is mocking the agent 🙏 I’ll check it and remove it.

NoritakaIkeda: [test: Remove unnecessary dml_execution_logs test case](https://github.com/liam-hq/liam/pull/2901/commits/7e7d74e5f8506ebc03432afa78bb4a879783e139)
This one has been deleted 🙏.

PR #2941, opened 2025-08-07
♻️ Migrate from trigger.dev to Vercel Functions for workflow execution

MH4GF:
## Issue

- resolve: https://github.com/route06/liam-internal/issues/5330

## Why is this change needed?

This PR migrates the LangGraph workflow execution from trigger.dev to direct Vercel Functions execution. This change enables:

1. **Simpler architecture** - Removes the dependency on trigger.dev background job system
2. **Direct execution** - Workflows run directly in Vercel Functions with 900-second timeout
3. **Foundation for streaming** - Prepares for future streaming response implementation
4. **Auto-trigger on page load** - Automatically starts workflow when user navigates to session page with initial message

### Key Changes:
- Removed trigger.dev task files and dependencies
- Updated chat API to execute workflows directly (deepModeling/invokeDbAgent)
- Pass `isDeepModelingEnabled` flag via query parameters
- Auto-trigger workflow on session page load for unanswered initial messages
- Handle initial state where no building_schema_versions exist yet

### Technical Details:
- Uses `useRef` to prevent duplicate execution in React Strict Mode
- Properly handles the case where no versions exist (initial state)
- Maintains checkpoint system (SupabaseCheckpointSaver) for workflow state persistence


frontend/apps/app/app/api/chat/route.ts
```diff
@@ -16,6 +15,9 @@ const chatRequestSchema = v.object({
   isDeepModelingEnabled: v.optional(v.boolean(), true),
 })
 
+// https://vercel.com/docs/functions/configuring-functions/duration#maximum-duration-for-different-runtimes
+export const maxDuration = 300
```
MH4GF, line 19: I would really like to go to 800 seconds, but that would require activating Fluid Compute. I would like to proceed separately.

junkisai (approved):
It seems to be working well after checking its operation!

image

PR #2943, opened 2025-08-07
Feat/add toast for command palette

tnyo43:
## Issue

- resolve:


frontend/packages/ui/src/components/Toast/Toast.tsx
```diff
@@ -46,9 +46,39 @@ export const Toast: FC<Props> = ({
...
+}: PropsWithChildren) => {
+  return (
+    <RadixToast.Provider>
+      {children}
+      <RadixToast.Viewport
+        className={clsx(styles.viewport, styles.commandPalette)}
+      />
+    </RadixToast.Provider>
+  )
+}
```
tnyo43, line 79: These viewport providers are exported for Storybook.

frontend/packages/ui/src/styles/variables.css
```diff
@@ -13,10 +13,10 @@
   --z-index-dropdown-menu: 1100;
   --z-index-popover: 1500;
   --z-index-tooltip: 2000;
-  --z-index-toast: 3000;
   --z-index-overlay: 3100;
   --z-index-toolbar-opened: 3200;
   --z-index-command-palette: 3300;
+  --z-index-toast: 3400;
```
tnyo43, line 19: The Toast z-index needs to be greater then CommandPalette one, otherwise the Toast will be hidden by the dialog overlay, like the following image.

Screenshot 0007-08-08 at 9 06 50

frontend/packages/ui/src/components/Toast/types.ts
```diff
@@ -9,3 +9,5 @@ export type ToastOptions = {
 export type ToastItem = ToastOptions & { isOpen: boolean }
 
 export type ToastFn = (options: ToastOptions) => void
+
+export type ToastPosition = 'header' | 'command-palette'
```
junkisai, line 13: I've noticed that the implementation in the `ui` package has code related to the specifications of the `erd-core` and `app` packages, and it's a bit concerning.

How about we use a union type like `'bottom-center' | 'top-right'`? 👀 
ref. https://sonner.emilkowal.ski/

junkisai (commented):
I've left one comment. 🙏

frontend/packages/ui/src/components/Toast/types.ts
tnyo43: Thanks for the comment ❤️ 

That library you mentioned is designed to be used in many different products, thus it needs to provide general options.
In our case, I guess `'top-right'` is not suitable since the position is not "top" really. We have some weird gap from the top of the window, which is for the header. I think it's more natural we directly specify the target components, as like "header", because the `Toast` component style practically depends on their styles.

But I understand your feeling. Currently the CommandPalette component is actually in the `erd-core` package, so that it will be a implicit opposite dependency from `ui` to `erd-core`. [Actually there is a similar problem on z-index custom property which CommandPalette component depends on](https://github.com/liam-hq/liam/blob/a7677f1a7329394d137f24f9a9501ca88a775714/frontend/packages/ui/src/styles/variables.css#L19) 🙈 

I think it would be a good solution to define `Header` and `CommandPalette` styles in the `ui` package. The structure will be something like the follows. If we define the positions and sizes of these component in the `ui` package, the implicit dependency from the `ui` to other packages will be removed.

```mermaid
graph TD
    subgraph ui
        uHeader[Header]
        uCommandPalette[CommandPalette]
        Toast
    end
    subgraph erd-core
        eCommandPalette[CommandPalette]
    end
    subgraph app
        aHeader[Header]
    end
    aHeader --> uHeader
    eCommandPalette --> uCommandPalette
    Toast --implicit dependency--> uHeader
    Toast --implicit dependency--> uCommandPalette
```

If you like this approach, I will create issues to apply these changes.

junkisai: @tnyo43 
You're right. To be precise, it's more accurate to say that it's displayed directly below the header component, rather than just in the top-right corner of the screen.

I agree with your proposal to manage the `Header` and `CommandPalette` within the `ui` package!
Please create the issues to apply these changes. 🙏

junkisai (approved):
It seemed like it wouldn't be a problem if the Header and CommandPalette were to be moved to the UI packages later. 👍

frontend/packages/ui/src/components/Toast/types.ts
tnyo43: Thank you, I created issues 😄 
- https://github.com/liam-hq/liam/issues/3042
- https://github.com/liam-hq/liam/issues/3043

PR #2965, opened 2025-08-10
✨ feat(command palette): Improve suggestion display in search input

tnyo43:
## Issue

- resolve:


tnyo43:
I checked the change is fine, but it has conflicts between the main branch.
I'll re-create the PR later.

PR #2969, opened 2025-08-12
🐛 Fix path module error in session creation

hoshinotsuyoshi:
## Issue

- resolve: https://github.com/liam-hq/liam/issues/5345

## Why is this change needed?

After PR #2941 migrated from trigger.dev to Vercel Functions, the session creation process started failing with `{"error":"The \"path\" argument must be of type string or an instance of Buffer or URL. ..."}` errors. This was caused by the Node.js `path` module being used in edge runtime environment where it's not available.

The fix removes the dependency on Node.js `path.resolve()` and replaces it with a simple string template literal to maintain compatibility with Vercel Functions edge runtime.


hoshinotsuyoshi:
was a wrong approach 🙏

PR #2993, opened 2025-08-13
📝 Add cross-repository reference guidance to create-internal-issue command

hoshinotsuyoshi:
## Issue

- resolve: https://github.com/route06/liam-internal/issues/5369

## Why is this change needed?

When creating issues in the `route06/liam-internal` repository using the `/create-internal-issue` command, references to PRs/issues from other repositories (e.g., `liam-hq/liam`) using the `#number` format are incorrectly interpreted as references within the `liam-internal` repository.

This change adds a best practice to the command documentation to ensure proper cross-repository references using the full `repository#number` format (e.g., `liam-hq/liam#2991`), which creates correct GitHub links.


MH4GF (approved):
👍🏻

PR #3006, opened 2025-08-14
✨ Add streaming API endpoint for AI chat workflow

junkisai:
※ The main focus is on the overall design, and the detailed implementation has not yet begun. Some parts may not work as expected in their current state, but this will not be merged into the main branch, so it will have no impact on the production environment.

## Issue

- resolve: N/A

## Why is this change needed?

This PR implements streaming support for the AI chat workflow to enable real-time response delivery. Instead of waiting for the entire workflow to complete before returning a response, the new streaming endpoint allows users to see AI responses as they are generated, significantly improving the user experience for complex database modeling tasks.

### Key improvements:
- Real-time feedback during long-running AI operations
- Better user experience with progressive response rendering
- Proper abort/cancellation support for interrupted requests
- Foundation for future streaming enhancements in the chat interface


frontend/apps/app/app/api/chat/stream/route.ts
```diff
@@ -0,0 +1,125 @@
...
+    ? await deepModelingStream(params, config)
+    : await invokeDbAgentStream(params, config)
+
+  return new Response(stream, {
+    headers: {
+      'Content-Type': 'text/event-stream',
+      'Cache-Control': 'no-cache, no-transform',
+      Connection: 'keep-alive',
+    },
+  })
```
junkisai, line 124: The code is the same as in `/api/chat/route.ts` except for this part.

frontend/internal-packages/agent/src/deepModelingStream.ts
```diff
@@ -0,0 +1,58 @@
...
+    //   if (ev.event === 'on_chat_model_stream' && ev.data?.chunk) {
+
+    //     const chunkId = ev.data.chunk.id
+    //     const content = ev.data.chunk.content
+    //     yield { event: 'assistant', data: { chunkId, content } }
+    //   } else {
+    //     ....
+    //   }
+    // }
+   */
```
junkisai, line 55: I believe an implementation like this part of open-swe is necessary.

ref. https://github.com/langchain-ai/open-swe/blob/210c848997c94f29905ae50fbcb22048675a18bf/apps/cli/src/utils.ts#L63-L89

frontend/apps/app/components/SessionDetailPage/hooks/useStream/useStream.ts
```diff
@@ -0,0 +1,92 @@
...
+
+type StreamError = {
+  type: 'network' | 'abort' | 'unknown'
+  message: string
+  status?: number
+}
+
+// TODO: Consider using useStream from '@langchain/langgraph-sdk/react'
+export const useStream = () => {
+  // NOTE: It might be better to manage with messages/values to align with the design of useStream from '@langchain/langgraph-sdk/react'
```
junkisai, line 20: I'm currently considering whether to manage `messages/values` instead of `timelineItems`, similar to how `useStream` is provided by `@langchain/langgraph-sdk/react`.

While aligning with the official implementation might have its advantages, I'm also thinking that since this is a custom implementation in the `apps/app` package, it might be more beneficial to make it more specific and tailored to our needs. 💭

frontend/apps/app/app/api/chat/stream/route.ts
```diff
@@ -0,0 +1,125 @@
...
+  createSupabaseRepositories,
+  deepModelingStream,
+  invokeDbAgentStream,
+} from '@liam-hq/agent'
+import { NextResponse } from 'next/server'
+import * as v from 'valibot'
+import { createClient } from '@/libs/db/server'
+
+// https://vercel.com/docs/functions/configuring-functions/duration#maximum-duration-for-different-runtimes
+export const maxDuration = 300
```
MH4GF: Let's change it to 800 in production implementation.

```suggestion
export const maxDuration = 800
```

frontend/apps/app/components/SessionDetailPage/components/Chat/Chat.tsx
```diff
@@ -48,7 +49,7 @@ export const Chat: FC<Props> = ({
     }
     onMessageSend(optimisticMessage)
 
-    await sendChatMessage({
+    await start({
```
MH4GF, line 52: On a slightly different note, there should be an additional process to retrieve past messages, just as timeline_items was retrieved in the Server Component. However, I believe this can be done as follows.

https://langchain-ai.github.io/langgraphjs/concepts/persistence/#get-state

MH4GF (approved):
LGTM 👍🏻

PR #3024, opened 2025-08-15
🐛 Fix duplicate React key error in artifact tab

hoshinotsuyoshi:
## Issue

- resolve: https://github.com/route06/liam-internal/issues/5391

## Why is this change needed?

Fixed an issue in the Artifacts tab where React’s duplicate key warning occurred. When use cases with the same number existed under different functional requirements, ReactMarkdown would generate identical keys, resulting in console errors.

## Summary by CodeRabbit

- **Style**
  - Updated use case headers in artifact output to use hierarchical numbering (e.g., “2.3 Title”), reflecting the parent functional requirement and its child use case.
  - Improves readability and scanning of artifacts with multiple requirements and use cases.
  - Applies to the Session Detail page’s Output > Artifact rendering.
  - No changes to content or structure—only header formatting is affected.

## screenshots

for example, `1. ...` -> `7.1 ....` . 

This change wasn’t required to fix the error, but I applied it anyway.

| before | after |
|--------|--------|
| スクリーンショット 2025-08-15 14 43 21 | スクリーンショット 2025-08-15 14 42 53 |

### DIFF
--- a/frontend/apps/app/components/SessionDetailPage/components/Output/components/Artifact/utils/formatArtifactToMarkdown.ts
+++ b/frontend/apps/app/components/SessionDetailPage/components/Output/components/Artifact/utils/formatArtifactToMarkdown.ts
@@ -43,10 +43,14 @@ function formatDmlOperation(operation: DmlOperation): string {
   return sections.join('\n')
 }
 
-function formatUseCase(useCase: UseCase, index: number): string {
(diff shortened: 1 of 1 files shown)


hoshinotsuyoshi:
@coderabbitai review

junkisai (approved):
thx!

PR #3045, opened 2025-08-17
♻️ refact(command palette): separate `SearchInput` component and introduce input-mode

tnyo43:
## Issue

- resolve:

### DIFF
--- a/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteContent/CommandPaletteContent.tsx
+++ b/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteContent/CommandPaletteContent.tsx
@@ -1,10 +1,11 @@
-import { Button, Search, Table2 } from '@liam-hq/ui'
+import { Button, Table2 } from '@liam-hq/ui'
 import { DialogClose } from '@radix-ui/react-dialog'
 import { Command } from 'cmdk'
 import { type FC, useCallback, useEffect, useState } from 'react'
 import { useTableSelection } from '@/features/erd/hooks'
 import { useSchemaOrThrow } from '@/stores'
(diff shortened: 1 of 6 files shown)


frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteContent/CommandPaletteContent.tsx
```diff
@@ -72,27 +77,29 @@ export const CommandPaletteContent: FC<Props> = ({ closeDialog }) => {
...
-            {Object.values(schema.current.tables).map((table) => (
-              <Command.Item key={table.name} value={table.name} asChild>
-                <a
-                  href={getTableLinkHref(table.name)}
-                  onClick={(event) => {
-                    // Do not call preventDefault to allow the default link behavior when ⌘ key is pressed
-                    if (event.ctrlKey || event.metaKey) {
-                      return
-                    }
+          {inputMode.type === 'default' && (
```
tnyo43, line 80: We are plan to show/hide option with input mode as like this. The table options will be displayed with `inputMode.type === "default"`, and the command options will be with `inputMode.type === "default" || inputMode.type === ''command'`.

This change won't affect the production since `inputMode.type` will be "default" permanently with the current implementation.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/types.ts
```diff
@@ -0,0 +1,4 @@
+export type CommandPaletteInputMode = { type: 'default' }
+// upcoming input mode
+// | { type: 'command' }
+// | { type: 'column'; tableName: string }
```
tnyo43, line 4: They will be introduced in the future changes 😎

PR #3078, opened 2025-08-20
🔧 Fix Server Actions CORS error with custom domain in production

FunamaYukina:
## Issue

- resolve: https://github.com/route06/liam-internal/issues/5427

## Why is this change needed?

In production, Server Actions are failing with the error:
x-forwarded-host header with value liam-erd-web.vercel.app does not match origin header with value liambx.com from a forwarded Server Actions request. Aborting the action.

This occurs because:
1. Users access the site via the custom domain `liambx.com`
2. Vercel's infrastructure internally processes requests through `liam-erd-web.vercel.app`
3. Next.js Server Actions perform security validation by comparing the `x-forwarded-host` and `origin` headers
4. The mismatch between these headers causes the validation to fail

This change adds `allowedOrigins` configuration for production environment only, allowing Server Actions to accept requests from both domains while keeping development environment unchanged.

## Testing

Since it cannot be verified in the preview environment, I would like to check it in the production environment after merging.

## Summary by CodeRabbit

* **Chores**
  * Restricted server-side actions to trusted production domains, enforcing cross-origin access rules while leaving development environments unrestricted.
  * Impact: Requests from unrecognized domains may be blocked in production, improving security and reliability for supported domains and preventing unintended cross-site interactions.

### DIFF
--- a/frontend/apps/app/next.config.ts
+++ b/frontend/apps/app/next.config.ts
@@ -22,7 +22,7 @@ const nextConfig: NextConfig = {
     serverActions: {
       allowedOrigins:
         process.env.VERCEL_ENV === 'production'
-          ? ['https://liambx.com', 'https://liam-erd-web.vercel.app']
+          ? ['liambx.com', 'liam-erd-web.vercel.app']
           : undefined,
     },
   },


frontend/apps/app/next.config.ts
```diff
@@ -18,6 +18,14 @@ const nextConfig: NextConfig = {
...
     ignoreDuringBuilds: true,
   },
+  experimental: {
+    serverActions: {
+      allowedOrigins:
+        process.env.NODE_ENV === 'production'
+          ? ['https://liambx.com', 'https://liam-erd-web.vercel.app']
+          : undefined,
+    },
+  },
```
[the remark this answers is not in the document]
FunamaYukina: I have made the corrections below. 
https://github.com/liam-hq/liam/pull/3078/commits/2079f9de77b523cba2a66c71f1761a829647dd96

frontend/apps/app/next.config.ts
```diff
@@ -18,6 +18,14 @@ const nextConfig: NextConfig = {
   eslint: {
     ignoreDuringBuilds: true,
   },
+  experimental: {
+    serverActions: {
+      allowedOrigins:
+        process.env.VERCEL_ENV === 'production'
+          ? ['https://liambx.com', 'https://liam-erd-web.vercel.app']
+          : undefined,
```
NoritakaIkeda: 📝
https://nextjs.org/docs/app/api-reference/config/next-config-js/serverActions#allowedorigins
It seems worth trying to see if this works.

> If not provided, only the same origin is allowed.

hoshinotsuyoshi (approved):

frontend/apps/app/next.config.ts
```diff
@@ -18,6 +18,14 @@ const nextConfig: NextConfig = {
   eslint: {
     ignoreDuringBuilds: true,
   },
+  experimental: {
+    serverActions: {
+      allowedOrigins:
+        process.env.VERCEL_ENV === 'production'
+          ? ['https://liambx.com', 'https://liam-erd-web.vercel.app']
```
NoritakaIkeda: Looking at the Next.js docs, it seems that only domain names should be specified, without the protocol. So I think the correct values are liambx.com and liam-erd-web.vercel.app.

```suggestion
          ? ['liambx.com', 'liam-erd-web.vercel.app']
```

https://nextjs.org/docs/app/api-reference/config/next-config-js/serverActions#allowedorigins

> /** @type {import('next').NextConfig} */
>  
> module.exports = {
>   experimental: {
>     serverActions: {
>       allowedOrigins: ['my-proxy.com', '*.my-proxy.com'],
>     },
>   },
> }

FunamaYukina: Thank you for pointing that out! You're absolutely right!

PR #3086, opened 2025-08-21
♻️ refact(command palette): add suggestion type to make the option categories clearer

tnyo43:
## Issue

~- resolve:~

## Why is this change needed?

It need to identify categories of options for a future update about preview, for example "users" is table option and "copy link" is command option.
To do so, I updated the option `value`s to include category inforrmation.

I defined `CommandPaletteSuggestion` type and converters `getSuggestionText`, which converts `CommandPaletteSuggestion` to suggestion text, and `textToSuggestion`, which does the reverse.

Basically, suggestion texts are used in the `Command.Option` of `cmdk`. The reason why I don't use `CommandPaletteSuggestion` values directly to the `Command.Option` is its value accepts only string values. We may be able to make the logic simpler by pushing some patch to the library in the future 😉

## Summary by CodeRabbit

* **New Features**
  * Unified text input with smart suggestions for tables and commands and mode switching support.

* **Refactor**
  * Keyboard/Enter now acts on the top suggestion; command items and filtering render consistent suggestion-based values for more predictable navigation.

* **API**
  * Suggestion utilities re-exported for reuse.

* **Tests**
  * Added unit and integration tests for suggestion serialization/parsing.

### DIFF
--- a/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/types.ts
+++ b/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/types.ts
@@ -1,3 +1,7 @@
 export type CommandPaletteInputMode = { type: 'default' } | { type: 'command' }
 // upcoming input mode
 // | { type: 'column'; tableName: string }
+
+export type CommandPaletteSuggestion =
+  | { type: 'table'; name: string }
+  | { type: 'command'; name: string }
--- a/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/utils/suggestion.ts
+++ b/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/utils/suggestion.ts
@@ -1,12 +1,14 @@
 import type { CommandPaletteSuggestion } from '../types'
 
+const SEPARATOR = '|'
+
 export const getSuggestionText = (suggestion: CommandPaletteSuggestion) =>
-  `${suggestion.type}|${suggestion.name}`
+  `${suggestion.type}${SEPARATOR}${suggestion.name}`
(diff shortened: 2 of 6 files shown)


frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/types.ts
```diff
@@ -1,3 +1,7 @@
 export type CommandPaletteInputMode = { type: 'default' } | { type: 'command' }
 // upcoming input mode
 // | { type: 'column'; tableName: string }
+
+export type CommandPaletteSuggestion =
+  | { type: 'table'; name: string }
+  | { type: 'command'; name: string }
```
tnyo43, line 7: Maybe we will add another suggestion type in the future 😄 
as like `{ type: "column", tableName: string, name: string }`.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/utils/suggestion.ts
```diff
@@ -0,0 +1,20 @@
...
+  const words = text.split(SEPARATOR)
+
+  const [suggestionType, name] = words
+  if (!suggestionType || !name) return null
+
+  if (suggestionType === 'table' || suggestionType === 'command')
+    return { type: suggestionType, name }
+
+  return null
+}
```
tnyo43, line 20: I defined two converters.
The combination of `textToSuggestion(getSuggestionText(x))` should be an identity function, which returns the same value with input.
I implemented some tests to check that characteristics in suggestion.test.ts.

MH4GF (approved):
LGTM 👍🏻

PR #3096, opened 2025-08-21
✨(agent): Add QA agent execution script for local debugging

MH4GF:
## Issue

- resolve: https://github.com/route06/liam-internal/issues/5452

## Why is this change needed?

This PR adds a standalone execution script for the QA agent workflow to enable faster debugging and local testing. Previously, debugging the QA agent required running the entire system workflow, which was time-consuming and made it difficult to isolate issues in specific nodes.

The new script allows developers to:
- Execute the QA agent workflow independently in ~50 seconds
- Test individual workflow nodes (generateUsecase, prepareDML, validateSchema)
- Resume from previous sessions using session IDs
- Run with debug logging for detailed inspection

This significantly improves the development experience when working on QA agent functionality.

### DIFF
--- a/frontend/internal-packages/agent/package.json
+++ b/frontend/internal-packages/agent/package.json
@@ -40,6 +40,8 @@
     "execute-design-process": "tsx scripts/executeDesignProcess.ts",
     "execute-design-process:debug": "tsx scripts/executeDesignProcess.ts --log-level=DEBUG",
     "execute-design-process:warn": "tsx scripts/executeDesignProcess.ts --log-level=WARN",
+    "execute-qa-agent": "tsx scripts/executeQaAgent.ts",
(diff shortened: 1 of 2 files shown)


NoritakaIkeda (approved):
It looks convenient and promising!
It might be a good idea to add a guard to prevent it from running in production, but since this is a debugging tool, I just did a rough check of the overall flow and confirmed there are no regressions or critical risks.
I think it looks good.👍

NoritakaIkeda:
> Resume from previous sessions using session IDs

This part only has the interface left, but the actual implementation doesn’t exist. I’ll remove it for now (and add it back if it becomes necessary).

MH4GF:
> This part only has the interface left, but the actual implementation doesn’t exist. I’ll remove it for now (and add it back if it becomes necessary).

Oh, I'll check it! thanks!!

PR #3159, opened 2025-08-25
✨ Convert chat interface to timeline-based message display

junkisai:
## Issue

- resolve: N/A

## Why is this change needed?

This PR converts the existing chat interface to use a timeline-based message display system that properly integrates with LangChain v2 streaming messages. The new architecture provides better separation of concerns, improved message handling, and a foundation for displaying AI agent interactions with their distinct avatars.

### Key improvements:
- **Unified message handling**: All messages now flow through a single Messages component that handles different message types
- **Agent avatars**: Each AI agent (PM, DB, QA) now has its own distinct avatar for better visual identification
- **Streaming integration**: Direct integration with LangChain SDK Message types for native streaming support
- **Component reusability**: Extracted shared components like CopyButton for use across multiple features

### Changes included:
1. New Messages component infrastructure with type guards and utility functions
2. Agent avatar components with Storybook stories
3. Integration with LangChain message streaming
4. Refactored Chat component to use new message display system
5. Shared CopyButton component extracted from Output components

## Known Issues / Next PR

- Only the **1st** and **3rd** of the expected four messages are being rendered.  
- Messages that should be sent by the **PM Agent** are currently attributed to the **DB Agent**.

https://github.com/user-attachments/assets/97fa0b2a-a669-4915-a7f8-1df31f28ab0e


MH4GF:
@junkisai There are a lot of comment outs, does that mean it is a work in progress? Can I review it?

junkisai:
@MH4GF 
Sorry, I neglected to add the supplementary comment.🙇 
I had commented it out to ensure nothing would be missed, as the migration from TimelineItem is not yet fully complete.

MH4GF:
@junkisai I see, For example, is it my understanding that this area will be reimplemented once the migration is complete?

https://github.com/liam-hq/liam/pull/3159/files#diff-69089906bb3e4cc3962e216aa760bd5e7f6d398c1a0ad19789bfb2c6264a18bcR66-R68

It would be helpful if you could add TODO comments for clarity!

junkisai:
@MH4GF 
That was the plan!
I’ve added TODO comments for clarity. 🙏 

[💬 Add TODO comments for Messages component integration](https://github.com/liam-hq/liam/pull/3159/commits/be0f394956c2745e6313763f7ebc5451b6790cd8)

MH4GF:
@junkisai Thanks!! I will review in detail tomorrow. Conflicts are occurring, please correct!

junkisai:
The conflict has been resolved.🙏

frontend/apps/app/components/SessionDetailPage/components/Chat/components/Messages/utils/messageTypeGuards.ts
```diff
@@ -0,0 +1,5 @@
+import type { AIMessage, Message } from '@langchain/langgraph-sdk'
+
+export function isAiMessage(m: Message): m is AIMessage {
```
MH4GF, line 3: This would be better to use the one provided by LangChain. Another PR is OK, of course.

https://v03.api.js.langchain.com/functions/_langchain_core.messages.isAIMessage.html

frontend/apps/app/components/SessionDetailPage/components/Chat/factories.ts
MH4GF, line 1: 👍🏻

frontend/apps/app/components/SessionDetailPage/components/Chat/components/Messages/utils/messageTypeGuards.ts
junkisai: I wanted to use this function, but I needed to determine the message type from a `Message` object from the `langgraph-sdk` instead of a `BaseMessage` object from Langchain. Therefore, I decided to implement it myself.

MH4GF: @junkisai I see, so these are different classes.
Sorry, what is the reason for using `@langchain/langgraph-sdk` AIMessage? What is the difference with `@langchain/core/messages`?

junkisai: @MH4GF 
The reason I'm using the `Message` type instead of `BaseMessage` is primarily because I was tracing the implementation of `@langchain/langgraph-sdk` while creating my own `useStream`, and that implementation referenced the `Message` type.

ref. https://github.com/langchain-ai/langgraphjs/blob/3320793bffffa02682227644aefbee95dee330a2/libs/sdk/src/react/stream.tsx

MH4GF: @junkisai 
First, there is no problem to create our own `useStream` 👍🏻 
I understand, However, it looks like langgraph-sdk is currently only used to read Message types. Type compatibility seems to be possible, but langgraph-sdk looks like a sdk for the langgraph platform, and I thought it would be better to use `@langchain/core/messages` unless there is a particularly strong reason to use it. what do you think?

junkisai: @MH4GF 
You're right, not using the Message type seems like a good idea as it would eliminate unnecessary conversion processes. 
I'll try to adjust it!

PR #3178, opened 2025-08-26
✨ Add tool calls extraction for AI message display

junkisai:
## Issue

- resolve: N/A

## Why is this change needed?

This PR extracts tool call information from AI messages and improves the display of tool calls in the chat interface. The changes separate schema definitions into a shared module and create a dedicated utility for extracting tool calls, making the codebase more maintainable and type-safe.

### Changes made:
1. **Schema extraction**: Moved message parsing schemas to a shared `schema.ts` file
2. **Tool calls utility**: Created `extractToolCallsFromMessage` utility for parsing tool calls from messages
3. **Type safety**: Defined `ToolCalls` type for better type safety across components
4. **UI improvements**: Simplified the ToolCalls component with cleaner styling and layout
5. **Conditional rendering**: Response messages now only render when content exists

### Technical details:
- Uses Valibot for runtime type validation of tool calls
- Follows existing patterns from `extractReasoningFromMessage`
- Maintains consistency with the existing message architecture

## Test

https://github.com/user-attachments/assets/32cdfe82-ad05-45c4-9b56-10417e3f83e3

### DIFF
--- a/frontend/apps/app/components/SessionDetailPage/components/Chat/components/Messages/AiMessage/ToolCalls/ToolCalls.tsx
+++ b/frontend/apps/app/components/SessionDetailPage/components/Chat/components/Messages/AiMessage/ToolCalls/ToolCalls.tsx
@@ -1,66 +1,40 @@
-import type { AIMessage } from '@langchain/langgraph-sdk'
+import { Wrench } from 'lucide-react'
 import type { FC } from 'react'
+import type { ToolCalls as ToolCallsType } from '../../schema'
 import styles from './ToolCalls.module.css'
 
-function isComplexValue(value: unknown): boolean {
-  return Array.isArray(value) || (typeof value === 'object' && value !== null)
-}
-
 type Props = {
-  toolCalls: AIMessage['tool_calls']
+  toolCalls: ToolCallsType
 }
 
 /**
- * TODO: Design Request
+ * TODO: Design Improvement
  *
- * ## Component Overview
- * Tool Execution History Cards - Visualizes the external tools/functions called by AI during task execution
+ * ## Current Implementation
+ * Displays tool calls made by AI during conversation with basic styling
  *
- * ## Display Elements
- * 1. Tool Name - Name of the executed function (e.g., search_database, analyze_data)
- * 2. Execution ID - Unique identifier for traceability (optional)
- * 3. Parameter List - Arguments passed to the tool in key-value format
- *    - Simple values: Display as plain text
- *    - Complex values (arrays/objects): Display as formatted JSON
+ * ## Areas for Enhancement
+ * 1. Visual Design - Improve card appearance and spacing
+ * 2. Icon System - Add specific icons for different tool types
+ * 3. Argument Display - Better formatting for JSON arguments
  */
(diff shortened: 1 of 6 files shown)


frontend/apps/app/components/SessionDetailPage/components/Chat/components/Messages/AiMessage/ToolCalls/ToolCalls.tsx
```diff
@@ -1,66 +1,40 @@
...
 import type { FC } from 'react'
+import type { ToolCalls as ToolCallsType } from '../../schema'
 import styles from './ToolCalls.module.css'
 
-function isComplexValue(value: unknown): boolean {
-  return Array.isArray(value) || (typeof value === 'object' && value !== null)
-}
-
 type Props = {
-  toolCalls: AIMessage['tool_calls']
```
junkisai, line 10: I was displaying the contents of `data.tool_calls`, but I changed it because it was more appropriate to display the contents of `data.additional_kwargs.tool_calls`.

Image

frontend/apps/app/components/SessionDetailPage/components/Chat/components/Messages/AiMessage/ToolCalls/ToolCalls.tsx
```diff
@@ -1,66 +1,40 @@
...
-                {tc.id && <code className={styles.id}>{tc.id}</code>}
-              </h3>
+          <div key={tc.id} className={styles.toolCall}>
+            <div className={styles.toolCallTitle}>
+              <Wrench className={styles.icon} />
+              <p className={styles.functionName}>{tc.function.name}</p>
+            </div>
+            <div className={styles.args}>
+              <span className={styles.argTitle}>ARGUMENTS</span>
+              <code className={styles.argCode}>{tc.function.arguments}</code>
```
MH4GF, line 36: IMO: In the future, I would like to switch the display for each tool. The most obvious example is that VersionMessage created timeline_items when the DB agent created a new schema version, but in LangChain terms, LLM will create [schemaDesignTool](https://github.com/liam-hq/liam/blob/c903acd8c9309a16e4555e457b607a3738e51eed/frontend/internal-packages/agent/src/db-agent/tools/schemaDesignTool.ts#L167) can be represented by the result of a call to.

MH4GF, line 36: Also, as a user, Lead Agent routing should not be of interest. In the case of the `routeToAgent` tool, we should consider hiding it or making it more user-friendly. Let's think about this later!

MH4GF (approved):
LGTM 👍🏻

frontend/apps/app/components/SessionDetailPage/components/Chat/components/Messages/AiMessage/ToolCalls/ToolCalls.tsx
junkisai: Right, I understand that in the future, we'll need to implement logic that controls what is displayed—for instance, showing or hiding elements, or determining which component to render—based on the function that is called.👍

PR #3179, opened 2025-08-26
🐛 Fix agent name parsing and Lead Agent support in AiMessage

junkisai:
## Issue

- resolve: N/A

## Why is this change needed?

This PR fixes two issues in the AiMessage component:

1. **Agent name parsing bug**: The `getAgentInfo` function was incorrectly returning DB Agent for all agent types due to a logic error
2. **Lead Agent support**: Added support for the new Lead Agent role in the streaming messages

### Changes made:
- Fixed message state preservation during streaming updates to maintain agent name and other properties
- Extended `agentRoleSchema` to include 'lead' role
- Added temporary Lead Agent display (currently using QA Agent avatar with TODO for dedicated component)
- Cleaned up unused Database type import

### Technical details:
The streaming update was replacing the entire message object instead of merging, which caused the agent name to be lost and default to DB Agent. This has been fixed by properly merging the incoming updates with existing state.

## Test

| Before | After |
|--------|--------|
| スクリーンショット 2025-08-26 14 04 07 | スクリーンショット 2025-08-26 13 59 17 |

### DIFF
--- a/frontend/apps/app/components/SessionDetailPage/components/Chat/components/Messages/AiMessage/AiMessage.tsx
+++ b/frontend/apps/app/components/SessionDetailPage/components/Chat/components/Messages/AiMessage/AiMessage.tsx
@@ -1,5 +1,4 @@
 import type { AIMessage } from '@langchain/langgraph-sdk'
-import type { Database } from '@liam-hq/db'
 import type { FC } from 'react'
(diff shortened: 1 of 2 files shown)


MH4GF (approved):
👍🏻

PR #3181, opened 2025-08-26
♻️(chat): Refactor message type system to use discriminated unions

junkisai:
## Issue

- resolve: Addresses feedback from #3159 (https://github.com/liam-hq/liam/pull/3159#discussion_r2299747853)

## Why is this change needed?

Initially planned to use LangChain's built-in message types as suggested in PR #3159, but after analysis, we found that:
- Our message structure doesn't fully align with LangChain's message schema
- Using discriminated unions with a `role` field provides better type safety for our use case
- This approach eliminates unnecessary dependencies and simplifies the codebase

This refactoring improves type safety and simplifies the message type system by:
- Replacing string literal types with discriminated union pattern using 'role' field
- Removing the need for separate type guard utilities as type checking is now built into the type system
- Simplifying type checks throughout components using direct role comparison
- Reducing dependencies by removing langchain/core

### Type System Refactoring
- Introduced discriminated union pattern with `role` field for message types
- Removed `messageTypeGuards.ts` utility as it's no longer needed
- Updated all message type checks to use direct role comparison

### Code Simplification
- Simplified message extraction utilities (`extractReasoningFromMessage`, `extractResponseFromMessage`, `extractToolCallsFromMessage`)
- Cleaned up component code that was using complex type guards
- Removed unused CSS classes from ReasoningMessage component

### Dependency Updates
- Removed `@langchain/core` dependency as it's no longer needed
- Cleaned up ESLint suppressions for deleted TimelineItem components

### Testing

I've confirmed that the behavior is the same as before the change.

https://github.com/user-attachments/assets/ff495e9c-2584-43da-85c3-6ab5fe76d1cf

### DIFF
--- a/frontend/apps/app/components/SessionDetailPage/components/Chat/Chat.stories.tsx
+++ b/frontend/apps/app/components/SessionDetailPage/components/Chat/Chat.stories.tsx
@@ -1,4 +1,4 @@
-import type { Message } from '@langchain/langgraph-sdk'
+import type { BaseMessage } from '@langchain/core/messages'
 import { aBuildingSchemaVersion } from '@liam-hq/db'
 import { aSchema } from '@liam-hq/schema'
 import type { Meta, StoryObj } from '@storybook/nextjs'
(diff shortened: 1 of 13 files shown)


frontend/apps/app/components/SessionDetailPage/components/Chat/components/Messages/AiMessage/AgentAvatar/AgentAvatars.stories.tsx
```diff
@@ -1,4 +1,4 @@
-import type { Meta, StoryObj } from '@storybook/react'
+import type { Meta, StoryObj } from '@storybook/nextjs'
```
MH4GF, line 1: 👍🏻

PR #3182, opened 2025-08-26
✨ Add streaming support for schema design agent

junkisai:
## Issue

- resolve: https://github.com/route06/liam-internal/issues/5439

## Why is this change needed?

This PR adds streaming support to the schema design agent, enabling real-time response display during schema generation. Users can now see the design process as it happens, improving the interactive experience especially for large schema designs.

### Changes:
- Replace `invoke` with `stream` method in `invokeDesignAgent` to enable streaming responses
- Add chunk accumulation and dispatch custom events for real-time updates
- Convert ToolNode `invoke` to `stream` in `invokeSchemaDesignToolNode`
- Unify chunk IDs using UUID for consistent handling across different providers (OpenAI and LangGraph)
- Maintain reasoning extraction after stream completion

## Test

https://github.com/user-attachments/assets/7d11fb8b-845a-400b-b51f-b0e1243ac1f2

### DIFF
--- a/frontend/internal-packages/agent/src/db-agent/invokeDesignAgent.ts
+++ b/frontend/internal-packages/agent/src/db-agent/invokeDesignAgent.ts
@@ -1,10 +1,12 @@
+import { dispatchCustomEvent } from '@langchain/core/callbacks/dispatch'
 import {
-  type AIMessage,
+  AIMessage,
+  AIMessageChunk,
   type BaseMessage,
   SystemMessage,
 } from '@langchain/core/messages'
 import { ChatOpenAI } from '@langchain/openai'
-import { ok, ResultAsync } from 'neverthrow'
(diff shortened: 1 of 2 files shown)


frontend/internal-packages/agent/src/db-agent/invokeDesignAgent.ts
```diff
@@ -33,22 +35,54 @@ export const invokeDesignAgent = (
...
+        )
+        const reasoning = parsed.success ? parsed.output : null
+
+        return {
+          response,
+          reasoning,
+        }
+      })(),
+      (error) => (error instanceof Error ? error : new Error(String(error))),
+    )
```
MH4GF, line 86: Once it starts to work, cut it out into common functions to make it easier to use!

MH4GF (approved):
Thanks!

PR #3190, opened 2025-08-27
🐛(pglite): Add test case and improve error messages for PGlite transaction abort issue

MH4GF:
## Issue

- resolve: #3176

## Why is this change needed?

When PGlite encounters an error within an explicit BEGIN/COMMIT transaction block, the transaction gets aborted but subsequent operations still return success responses even though they are ignored. This causes misleading error reporting.

**Important Note:** This PR does NOT fix the underlying PGlite transaction handling issue. The issue appears to be a limitation in PGlite itself that requires deeper investigation. Instead, this PR:

1. **Documents the problem** with a regression test that demonstrates the transaction abort behavior
2. **Improves error visibility** by enhancing how validation errors are displayed to users

### Changes made:

1. **Added regression test** (`PGliteInstanceManager.test.ts`)
   - Demonstrates the transaction abort problem with explicit BEGIN
   - Shows that subsequent commands after an error return success but are actually ignored
   - Serves as documentation of the issue for future debugging

2. **Improved error message formatting** (`formatValidationErrors.ts`)
   - Extracted error formatting logic into a testable function
   - Shows SQL statements alongside their error messages for better context
   - Makes it clearer which operations failed and why

3. **Added comprehensive tests** for the error formatting function
   - Ensures error messages are properly formatted
   - Tests edge cases like long SQL statements and special characters

### What this PR does NOT do:

- Does not fix the root cause of transaction abort handling in PGlite
- Does not prevent the misleading "success" responses after transaction abort
- The actual fix would require changes to how PGlite handles transactions or how we interact with PGlite

### Next steps:

The root cause fix would likely require one of:
- Upstream fix in PGlite library
- Wrapping PGlite transactions with additional state management
- Converting explicit BEGIN/COMMIT to implicit transactions

For now, the improved error messages will at least make it clearer to users when transaction issues occur.

### DIFF
--- a/frontend/internal-packages/agent/src/qa-agent/validateSchema/formatValidationErrors.test.ts
+++ b/frontend/internal-packages/agent/src/qa-agent/validateSchema/formatValidationErrors.test.ts
@@ -1,0 +1,309 @@
+import { describe, expect, it } from 'vitest'
+import { formatValidationErrors } from './formatValidationErrors'
+import type { TestcaseDmlExecutionResult } from './types'
+
+describe('formatValidationErrors', () => {
+  it('should format single error with single operation', () => {
+    const results: TestcaseDmlExecutionResult[] = [
+      {
+        testCaseId: 'test-1',
+        testCaseTitle: 'Test Insert Operation',
+        success: false,
+        executedOperations: 1,
+        failedOperations: [
+          {
+            sql: "INSERT INTO users (id, name) VALUES (1, 'John')",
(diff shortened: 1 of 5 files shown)


FunamaYukina (approved):
LGTM!
I understand this pull request is linked to the following issue:
- https://github.com/route06/liam-internal/issues/5492

I hope we can resolve this conflict.

MH4GF:
@FunamaYukina Thanks! I'll resolve conflict.

PR #3199, opened 2025-08-27
💄 style(command palette): fix the `CommandPalettePreview` styles for the `CommandPreview` videos

tnyo43:
## Issue

- resolve:

## Why is this change needed?

To introduce the CommandPreview component with videos, this PR will fix the preview container size with the video size.
It will also arrange the padding size of the elements to fit with the design data.

[Here is one of preview videos](https://github.com/liam-hq/liam/blob/e859f8258e452d44456af89f78e32fdfeadeaa08/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/assets/zoom-to-fit.mp4). The size is  386:271 (height : width).

|before|after|
| --- | --- |
|Screenshot 0007-08-31 at 10 31 37|Screenshot 0007-08-31 at 10 31 44|
|Screenshot 0007-08-31 at 10 30 44|Screenshot 0007-08-31 at 10 30 38|

## Summary by CodeRabbit

* Bug Fixes
  * Command Palette preview now renders at a consistent size with stable scrollbars, preventing overflow and layout shifts.
  * List panel adapts to available space for better responsiveness.
  * Adjusted padding to improve readability and alignment in the preview.

* Style
  * Refined spacing and rounded corners for a cleaner, more consistent look.

* Chores
  * Added changeset for a patch release documenting the Command Palette table preview size fix.

### DIFF
--- a/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/CommandPalettePreview.module.css
+++ b/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/CommandPalettePreview.module.css
@@ -6,9 +6,10 @@
 }
 
 .tableNodeWrapper {
-  padding: var(--spacing-10) var(--spacing-12);
+  padding: var(--spacing-10) var(--spacing-9);
   height: 100%;
   width: 100%;
   overflow-x: hidden;
   overflow-y: scroll;
+  scrollbar-gutter: stable both-edges;
 }
--- a/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/CommandPalettePreview.module.css
+++ b/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/CommandPalettePreview.module.css
@@ -1,7 +1,12 @@
 .container {
+<<<<<<< HEAD
   padding: var(--spacing-10) var(--spacing-12);
   background-color: var(--global-background);
-  height: 100%;
+  height: 361px;
+=======
+  height: 386px;
+>>>>>>> 59f96a61c (fix)
+  width: 271px;
   overflow-x: hidden;
   overflow-y: scroll;
   border-radius: var(--border-radius-lg-plus);
--- a/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/CommandPalettePreview.module.css
+++ b/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/CommandPalettePreview.module.css
@@ -1,13 +1,14 @@
 .container {
-<<<<<<< HEAD
-  padding: var(--spacing-10) var(--spacing-12);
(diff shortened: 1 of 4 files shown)


frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/CommandPalettePreview.module.css
```diff
@@ -1,8 +1,15 @@
...
+
+.tableNodeWrapper {
+  padding: var(--spacing-10) var(--spacing-9);
   height: 100%;
+  width: 100%;
   overflow-x: hidden;
   overflow-y: scroll;
-  border-radius: var(--border-radius-lg-plus);
+  scrollbar-gutter: stable both-edges;
 }
```
tnyo43, line 15: The paddings and `overflow-y: scroll;` are the feature only the `TablePreview` component uses, so that I extracted them from the common container styles.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/CommandPalettePreview.module.css
```diff
@@ -1,8 +1,15 @@
...
+}
+
+.tableNodeWrapper {
+  padding: var(--spacing-10) var(--spacing-9);
   height: 100%;
+  width: 100%;
   overflow-x: hidden;
   overflow-y: scroll;
-  border-radius: var(--border-radius-lg-plus);
+  scrollbar-gutter: stable both-edges;
```
tnyo43, line 14: This is to fix the unexpected gap for scrollbar area.
https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter

You can see the left padding is bigger than the right if it doesn't have `scrollbar-gutter: stable both-edges` declaration. It's because the right padding include the scrollbar width but the left doesn't. With the declaration, the left padding includes the scrollbar width as well, so the table node element comes to the center of the area.

|with `scrollbar-gutter: stable both-edges;`|without|
| --- | --- |
|Image|Image|

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/CommandPalettePreview.module.css
```diff
@@ -1,8 +1,15 @@
 .container {
-  padding: var(--spacing-10) var(--spacing-12);
+  height: 386px;
+  width: 271px;
```
tnyo43, line 3: Here is the most important change of this PR. The preview container fixes it size by setting specific number rather than just inheriting the parent size.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/CommandPalettePreview.module.css
junkisai: I learned something new from this, thank you!

junkisai (approved):
nice work!!

PR #3203, opened 2025-08-28
✨ Add streaming support for lead agent summary workflow

junkisai:
## Issue

- resolve: N/A

## Why is this change needed?

This PR adds streaming support to the lead agent's summary workflow, enabling real-time summary generation. Previously, users had to wait for the entire summary to be generated before seeing any output. With this change, users can see the summary being generated progressively, providing better user experience and immediate feedback.

### Key Changes:
- Modified `summarizeWorkflow` to stream responses chunk-by-chunk
- Implemented proper chunk accumulation and event dispatching
- Updated `invokeSaveArtifactToolNode` to use streaming for consistency
- Ensures consistent UUID assignment for all message chunks


MH4GF (approved):
👍🏻

PR #3209, opened 2025-08-28
♻️ Improve SSE streaming implementation with centralized constants and proper end event handling

junkisai:
## Issue

- resolve: N/A (refactoring improvement)

## Why is this change needed?

This PR improves the Server-Sent Events (SSE) streaming implementation to make it more maintainable and robust:

1. **Centralized SSE Constants**: Extracted hardcoded event strings ('messages', 'end') into a shared constants module to ensure consistency across the codebase and prevent typos
2. **Proper Stream Completion**: Added handling for the 'end' event in the useStream hook to correctly set streaming state when the stream completes
3. **Better Code Organization**: Created a dedicated client module in the agent package for SSE-related utilities

These changes prevent issues with incomplete stream handling and make the codebase easier to understand and modify.

## Summary by CodeRabbit

- New Features
  - Explicit end-of-stream handling for chat/session streams for clean termination.
- Improvements
  - Standardized streaming event names across the app for consistent live updates.
- Bug Fixes
  - Prevented lingering "streaming" state and reduced duplicate end-of-stream updates.
- Performance
  - Faster disconnect when a stream finishes, improving responsiveness.
- Chores
  - Added a browser-safe client entry in the agent package to stabilize imports.


frontend/internal-packages/agent/src/client.ts
```diff
@@ -0,0 +1,4 @@
+// Client-safe exports that can be used in browser environments
```
MH4GF, line 1: We may also use the `client-only` package.

- ref: 
    - https://www.npmjs.com/package/client-only
    - https://nextjs.org/docs/app/getting-started/server-and-client-components#preventing-environment-poisoning

MH4GF (approved):
👍🏻

PR #3212, opened 2025-08-28
🐛 Fix UserDropdown button nesting issue and avatar alignment

junkisai:
## Issue

- resolve: HTML validation error and avatar misalignment in AppBar

## Why is this change needed?

This PR fixes two issues in the UserDropdown component:

1. **HTML Validation Error**: The previous implementation had a button element nested inside another button (via DropdownMenuTrigger), which violates HTML specifications and causes the error "In HTML, <button> cannot be a descendant of <button>."

2. **Avatar Alignment**: The avatar was incorrectly positioned at the left side of the AppBar instead of the right side where user controls should be located.

| 1 | 2 |
|--------|--------|
| image | image | 

### Changes Made:
- Removed the unnecessary button wrapper around Avatar components
- Deleted the associated CSS module that was only styling the wrapper button
- Added `margin-left: auto` to BaseAppBar's actions container to properly align user controls to the right
- Avatar components now serve directly as dropdown triggers while maintaining accessibility with aria-label attributes

This simplifies the component structure while fixing both the HTML validation issue and the layout problem.

## Summary by CodeRabbit

* **Style**
  * App bar layout updated: right-side actions now align to the far right.
  * User trigger visuals simplified by removing custom button styling; the avatar now serves directly as the trigger.

* **Refactor**
  * Reduced markup for the user dropdown by removing an extra wrapper while preserving existing behavior and flows.

### DIFF
--- a/frontend/apps/app/components/CommonLayout/AppBar/UserDropdown.module.css
+++ b/frontend/apps/app/components/CommonLayout/AppBar/UserDropdown.module.css
@@ -1,8 +1,0 @@
-.triggerButton {
-  background: none;
-  border: none;
-  padding: 0;
-  margin: 0;
-  cursor: pointer;
-  display: inline-block;
-}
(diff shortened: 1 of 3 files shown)


hoshinotsuyoshi:
thx! ❤️

PR #3216, opened 2025-08-28
✨(agent): Add message retrieval from checkpointer and fix streaming name preservation

junkisai:
## Issue

- Related: https://github.com/route06/liam-internal/issues/5451

## Why is this change needed?

This PR addresses two key issues:
1. Enable retrieval of conversation history from LangGraph checkpointer storage
2. Fix a bug where agent names were not preserved during message streaming

### 1. Message Retrieval from Checkpointer
- Added `getMessages()` function to fetch thread messages from checkpoint storage
- Implemented proper serialization/deserialization for Server-to-Client Component boundary
- Used `ts-pattern` for type-safe message revival

### 2. AIMessageChunk Name Field Preservation
- Fixed issue where `AIMessageChunk.concat()` doesn't preserve the `name` field
- Added `AGENT_NAME` constants for each agent (lead, pm, db)
- Explicitly set name field when converting chunks to final AIMessage

### Message Serialization
- Server Components can only pass plain objects to Client Components
- BaseMessage class instances must be serialized to StoredMessage format
- Messages are revived in Client Components using pattern matching

### Agent Name Constants
- Centralized agent names as constants to avoid string duplication
- Ensures consistency across all agent implementations

## Testing

https://github.com/user-attachments/assets/d95593e0-a750-4605-a8db-ced050c82dfe

## Notes
- PublicSessionDetailPage message fetching is blocked by organizationId requirement (added TODO)
- AIMessageChunk.concat() limitation is now documented with workaround implemented

## Summary by CodeRabbit

- New Features
  - Session detail pages now preload and restore prior chat history for smoother continuity.
  - Initial messages are provided to the chat stream for faster, more consistent rendering.
  - Public session pages prepared to support loading messages in a future release (no visible change yet).

- Bug Fixes
  - Streamed agent replies now consistently display correct agent labels (DB, Lead, PM), preventing mislabeling on long responses.


frontend/internal-packages/agent/src/getMessages.ts
```diff
@@ -0,0 +1,36 @@
...
+    return []
+  }
+
+  const messagesData = channelValues['messages']
+  if (!Array.isArray(messagesData)) {
+    return []
+  }
+
+  return messagesData.filter(isBaseMessage)
+}
```
MH4GF, line 36: Of course we can leave it as it is this time since we can think about it later, but what I was thinking about after the migration from timeline_items to messages is to restore the state from checkpointer and use it as a UI. I am thinking that perhaps schema and artifact do not need to be prepared as their own database tables.

MH4GF (approved):
Nice, thanks!!

frontend/internal-packages/agent/src/getMessages.ts
junkisai: Oh, I see. I was wondering how to display the schema and query results, but it looks like I can manage it by pulling and using the state from the checkpointer, not just the message list.

PR #3226, opened 2025-08-29
fix: change Avatar import from @/components to @liam-hq/ui in HumanMessage

MH4GF:
## Issue

- resolve: #3222

## Why is this change needed?

The HumanMessage component was importing Avatar from `@/components`, which caused Storybook build failures.

### Root Cause

- Next.js tsconfig.json has multiple path mappings for `@/*`:
  - `./components` (within app)  
  - `../../packages/ui/src/components` (within ui package)
- Storybook's webpack config only resolves `@` to `apps/app`, missing the additional path mappings
- When `@/components` is used in Next.js app, it automatically falls back to `ui/src/components` since there's no `components/index.ts` in app
- But Storybook can't resolve this, causing the build error

### Solution

Changed the import from `@/components` to `@liam-hq/ui` for consistency and to fix the Storybook build.

### Test plan
- [x] Storybook build succeeds (`pnpm --filter @liam-hq/storybook build`)
- [x] HumanMessage component renders correctly in Storybook

## Summary by CodeRabbit

* **Style**
  * Chat avatars in human messages now use the unified design system, resulting in minor visual refinements and improved consistency across the app.

* **Refactor**
  * Switched to a shared UI library for the avatar component in chat messages. No functional changes; behavior remains the same.

### DIFF
--- a/frontend/apps/app/components/SessionDetailPage/components/Chat/components/Messages/HumanMessage/HumanMessage.tsx
+++ b/frontend/apps/app/components/SessionDetailPage/components/Chat/components/Messages/HumanMessage/HumanMessage.tsx
@@ -1,7 +1,7 @@
 import type { HumanMessage as HumanMessageType } from '@langchain/core/messages'
+import { Avatar } from '@liam-hq/ui'
 import type { FC } from 'react'
 import * as v from 'valibot'
-import { Avatar } from '@/components'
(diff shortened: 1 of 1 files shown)


FunamaYukina (approved):
Thank you!
I confirmed that ``pnpm --filter @liam-hq/storybook build`` succeeds in my local environment.

PR #3228, opened 2025-08-29
feat(qa-agent): implement map-reduce pattern for parallel test case generation

MH4GF:
## Summary
- Implements map-reduce pattern using LangGraph.js Send API for parallel test case generation
- Fixes issue #5517 by improving test case generation reliability and performance
- Replaces single-node implementation with subgraph architecture for proper state isolation

## Why is this change needed?
The previous QA agent implementation had several issues:
1. **Poor Performance**: Single LLM call processing all requirements sequentially
2. **Missing Test Cases**: Some requirements would be skipped in generation
3. **Retry Logic Issues**: Message contamination across parallel executions

This PR solves these issues by:
- Processing each requirement in parallel using Send API
- Isolating state with subgraphs to prevent message contamination
- Implementing proper retry logic within each subgraph

## Technical Details
- **Send API**: Distributes requirements to parallel subgraphs
- **Subgraph Architecture**: Each requirement gets isolated message state
- **Command API**: Returns results from subgraphs to parent graph
- **Reducers**: Automatically aggregate test cases from parallel executions

## Test plan
- [ ] Run `pnpm --filter @liam-hq/agent execute-qa-agent` and verify parallel processing
- [ ] Check that all requirements generate test cases
- [ ] Verify retry logic works correctly on tool failures
- [ ] Confirm no message contamination between parallel executions


frontend/internal-packages/agent/src/qa-agent/types.ts
```diff
@@ -8,7 +8,7 @@ export const testcaseSchema = v.object({
   requirement: v.string(),
   title: v.string(),
   description: v.string(),
-  dmlOperations: v.array(dmlOperationSchema),
+  dmlOperation: dmlOperationSchema,
```
MH4GF, line 11: The decision to split the QAAgent into separate runs led to the creation of a large number of test cases for a single requirement. One of the moves we made to solve this was to generate all requirements, test cases and dmlOperation one by one.

PR #3238, opened 2025-08-29
✅(agent): Add integration tests and remove setupGraph helper

MH4GF:
## Issue

- resolve: N/A (Proactive improvement)

## Why is this change needed?

This PR adds integration test infrastructure for the agent package and removes the setupGraph helper to fix type errors. The changes include:

1. **Integration test infrastructure**: Added test utilities and configuration for running integration tests separately from unit tests
2. **Test coverage**: Added integration tests for critical workflows (designSchemaNode, summarizeWorkflow, analyzeRequirementsNode)  
3. **Type safety improvement**: Removed the setupGraph helper function which was causing type errors and replaced with direct usage of typed methods

These changes improve the reliability and maintainability of the agent package by providing better test coverage and cleaner type-safe code.

### DIFF
--- a/frontend/internal-packages/agent/package.json
+++ b/frontend/internal-packages/agent/package.json
@@ -55,6 +55,7 @@
     "lint:biome": "biome check .",
     "lint:eslint": "eslint .",
     "lint:tsc": "tsc --noEmit",
-    "test": "vitest --watch=false --passWithNoTests"
+    "test": "vitest --watch=false --passWithNoTests",
+    "test:integration": "vitest --watch=false --passWithNoTests --config vitest.config.integration.ts"
   }
 }
--- a/frontend/internal-packages/agent/src/db-agent/nodes/designSchemaNode.integration.test.ts
+++ b/frontend/internal-packages/agent/src/db-agent/nodes/designSchemaNode.integration.test.ts
@@ -1,0 +1,43 @@
+import { HumanMessage } from '@langchain/core/messages'
+import { END } from '@langchain/langgraph'
+import { describe, it } from 'vitest'
+import {
+  getTestConfig,
+  outputStream,
+  setupGraph,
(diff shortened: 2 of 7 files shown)


frontend/internal-packages/agent/package.json
```diff
@@ -55,6 +55,7 @@
     "lint:biome": "biome check .",
     "lint:eslint": "eslint .",
     "lint:tsc": "tsc --noEmit",
-    "test": "vitest --watch=false --passWithNoTests"
+    "test": "vitest --watch=false --passWithNoTests",
+    "test:integration": "vitest --watch=false --passWithNoTests --config vitest.config.integration.ts"
```
hoshinotsuyoshi, line 59: LGTM! 

Q. The interleaved output from parallel test execution seems like a good tradeoff for performance. I assume this was intentional - correct?

hoshinotsuyoshi (approved):
LGTM!

frontend/internal-packages/agent/package.json
MH4GF: @hoshinotsuyoshi You are referring to interleaved output, where multiple tests are output together at once, right?
If so, then Yes. However, I was not really thinking of a use case to run all the tests, but mainly to specify a file to run when I want to run a particular node.

hoshinotsuyoshi: @MH4GF  thx!  So you'd use it like `pnpm --filter @liam-hq/agent test:integration src/db-agent/nodes/designSchemaNode.integration.test.ts`, right? That makes total sense! Much more practical to run specific integration tests on  demand rather than all at once.

MH4GF: That's right 👍🏻 However, some documentation may be necessary, as it is not intuitive for AI agents about assuming stand-alone execution and the lack of assertion. We would like to improve this as we use it.

PR #3243, opened 2025-08-31
✨ feat(command palette): display CommandPreview when command options are selected

tnyo43:
## Issue

~- resolve:~

## Why is this change needed?

This PR will create `CommandPreview ` component which shows previews for the `CommandOptions`.
To show images and videos as previews, it refers to assets from the CDN server (https://assets.liambx.com).

preview with `CommandPreiew` component -> https://liam-ltk5d1wqq-liambx.vercel.app/

https://github.com/user-attachments/assets/ab55853e-361e-41c6-af03-3cae60bd413f

I finally removed the `CommandOptions` and the `CommandPreview` component, so this change will not affect the production.

## Summary by CodeRabbit

- New Features
  - Command Palette now displays visual previews for commands (videos or images) while browsing suggestions.
  - Command options panel is shown when in default or command input modes.
- Style
  - Updated preview panel with fixed dimensions and larger corner radius.
  - Adjusted padding, ensured full-width content, and stabilized scrollbar behavior.
  - Images and videos now scale to fill the preview area.
- Tests
  - Added unit tests verifying correct rendering of video and image previews for supported commands.


frontend/apps/app/next.config.ts
```diff
@@ -102,6 +102,11 @@ const nextConfig: NextConfig = {
       },
     })
 
+    config.module.rules.push({
+      test: /\.mp4$/,
+      type: 'asset/resource',
+    })
```
tnyo43: To import mp4 video file, I added these lines.

It used to be very common to use [`file-loader`](https://www.npmjs.com/package/file-loader/v/1.1.1), but the webpack itself supports the feature since webpack 5.

ref: https://webpack.js.org/guides/asset-modules/#resource-assets

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/CommandPreview.tsx
```diff
@@ -0,0 +1,47 @@
...
+  'copy link': copyLink,
+  'Zoom to Fit': zoomToFit,
+  'Tidy Up': tidyUp,
+}
+
+const COMMAND_IMAGE_SOURCE: Record<string, string | StaticImageData> = {
+  'Show All Fields': showAllFields,
+  'Show Key Only': showKeyOnly,
+  'Show Table Name': showTableName,
+}
```
tnyo43, line 23: We may be able to define command texts type to link the `CommandOptions` and this preview,
like:

```tsx
const COMMAND_VIDEO_SOURCE: Partial<Record<CommandText, string>> = {
  'copy link': copyLink,
  'Zoom to Fit': zoomToFit,
  'Tidy Up': tidyUp,
}
```

I'll try it in another PR.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/CommandPreview.tsx
```diff
@@ -0,0 +1,47 @@
...
+      {COMMAND_VIDEO_SOURCE[commandName] && (
+        <video muted autoPlay className={styles.video} key={commandName}>
+          <source src={COMMAND_VIDEO_SOURCE[commandName]} type="video/mp4" />
+        </video>
+      )}
+      {COMMAND_IMAGE_SOURCE[commandName] && (
+        <img
+          src={getImageSrc(COMMAND_IMAGE_SOURCE[commandName])}
+          className={styles.image}
+          alt=""
```
tnyo43, line 42: I think these images are basically for decorations. It may help understanding, but the option names are good enough to understand the command contents. Therefore I didn't set additional `alt` texts for them.

frontend/packages/erd-core/src/images.d.ts
```diff
@@ -1,4 +1,9 @@
 declare module '*.png' {
+  const value: string | import('next/image').StaticImageData
```
tnyo43: I installed `next` in @liam-hq/erd-core as devDependency to refer to the type.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/CommandPreview.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+const COMMAND_IMAGE_SOURCE: Record<string, string | StaticImageData> = {
+  'Show All Fields': showAllFields,
+  'Show Key Only': showKeyOnly,
+  'Show Table Name': showTableName,
+}
+
+const getImageSrc = (imageSrc: string | StaticImageData) =>
+  typeof imageSrc === 'string' ? imageSrc : imageSrc.src
```
tnyo43: This is to handle image object for both Vite CLi app and Next.js app.
I think we have several options to resolve this problem, but I don't know the best way (even I don't know there is a good way 😅).
Let's discuss it and make it better later.

tnyo43:
I'll open this PR after https://github.com/liam-hq/liam/pull/3199 will be resolved.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/CommandPreview.tsx
MH4GF: I've been struggling with this for a bit, but a more fundamental issue might be whether to bundle the assets or retrieve them remotely.
This is a method of preparing file storage and distributing from a CDN.
What I'm wondering about is the issue of different bundling methods between vite and webpack, and the fact that the asset files are each included in the dist directory when the user runs the CLI.
(Actually, there is already an image for OGP included in the dist directory, but I was willing to accept it because I thought it was ok if only one image was included.)
@hoshinotsuyoshi I would like to hear your opinion as well.

hoshinotsuyoshi: @MH4GF 

thx! There seem to be several topics to discuss:
- The effort required to write code that bridges the gap between vite and webpack
- Taking an approach of distributing from CDN instead of bundling
  - Additional `Pros` in this case: reduced size when installing the CLI
  - `Cons`: would make it impossible to use in completely closed environments like corporate intranets
  - `Cons`: CDN maintenance cost

I don't see this as a blocker for this PR, but it's worth discussing!

For now, I've confirmed that the assets added in this PR are approximately 1.2MB. I tried `cd frontend/packages/cli && pnpm build && pnpm pack` and confirmed that the CLI tarball increases from 2.7MB to 3.9MB. The numbers add up.

----

Also, I'm curious about how `Expo` CLI and `Storybook` CLI handle this. I couldn't finish investigating before responding... :)

MH4GF: Thanks for doing the research! My opinion is as follows.

- Cons: would make it impossible to use in completely closed environments like corporate intranets
    - This may be a problem, but we can consider it after we get feedback from users.
- Cons: CDN maintenance cost
    - I guess we should consider how much this cost would be.

frontend/apps/app/types/images.d.ts
junkisai, line 1: [nits]
This is a small detail, but I thought it would be better to name the file `assets.d.ts` if you're going to add a type definition for `.mp4` video files.👀

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/CommandPreview.tsx
tnyo43: @hoshinotsuyoshi @MH4GF 
Thanks for pointing it out! I didn't consider the way to use CDN to distribute the assets that much 🙇
Increasing the package size is a problem, so I'm happy to hear your opinions.

> I don't see this as a blocker for this PR, but it's worth discussing!

So, is it okay if I go with the current way for now

frontend/apps/app/types/images.d.ts
tnyo43: Thanks for pointing it out!
I move the definition of ".mp4" into assets.d.ts for both @liam-hq/app and @liam-hq/erd-core.
For the @liam-hq/erd-core, I moved the images.d.ts file into the `type` directory to make the same structure with the @liam-hq/app package.

https://github.com/liam-hq/liam/pull/3243/commits/4cf02224688feb4ed6fc32e5f97f51070e6b6651

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/CommandPreview.tsx
MH4GF: @tnyo43 cc @hoshinotsuyoshi 

> So, is it okay if I go with the current way for now

Since the implementation policy seems to be very different, I thought the latter would be less costly than "change to using remote url after merging this PR" or "wait for remote url to be ready and then implement it again". What do you think?
I don't think it will take that long to prepare the CDN. Perhaps we can have it ready today.

tnyo43: @MH4GF 

> I don't think it will take that long to prepare the CDN. Perhaps we can have it ready today.

Really? That’s wonderful to hear 😄 
I also believe releasing with the CDN-based implementation will be less costly, so I’ll wait until it’s ready.
Thank you!

frontend/apps/app/types/images.d.ts
junkisai: thx!!

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/CommandPreview.tsx
tnyo43: @hoshinotsuyoshi @MH4GF 

Thanks for the discussion and creating the CDN server!
I modified the code to refer to the assets from the server https://github.com/liam-hq/liam/pull/3243/commits/2d87ec2f91385d3cc789664c47ec79eaa51d9ad5

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalettePreview/CommandPreview.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+const COMMAND_IMAGE_SOURCE: Record<string, string> = {
+  'Show All Fields':
+    'https://assets.liambx.com/erd-core/2025-09-01/images/show-all-fields.png',
+  'Show Key Only':
+    'https://assets.liambx.com/erd-core/2025-09-01/images/show-key-only.png',
+  'Show Table Name':
+    'https://assets.liambx.com/erd-core/2025-09-01/images/show-table-name.png',
+}
```
tnyo43, line 23: I made maps to refer to the assets which is set in https://github.com/liam-hq/liam/pull/3326
I think the assets are the same for any environments, so that I didn't consider to use environment variable, as like `${process.env.CDN_BASE_URL}/erd-core/...`.

MH4GF: Yes, I think this is enough for now since we are in the same environment.

MH4GF (approved):
LGTM 👍🏻 Thanks for adjusting!!

PR #3296, opened 2025-09-03
🐛 Fix Vercel preview environment baseUrl undefined issue

hoshinotsuyoshi:
## Issue

- resolve: Fix undefined baseUrl in Vercel preview environment for Rive asset loading https://github.com/route06/liam-internal/issues/5207

## Why is this change needed?

The OutputPlaceholder component was attempting to construct full URLs for the Rive animation asset using environment variables, which caused issues in Vercel preview environments where VERCEL_URL and VERCEL_BRANCH_URL are not accessible on the client-side.

The solution was to simplify the approach by using a relative path `/assets/loading_jack.riv` instead of constructing full URLs with environment-specific logic. This change:
- Removes complex environment variable handling that wasn't working reliably
- Uses a simple relative path that works across all environments (local, preview, production)
- Eliminates the undefined URL issue in Vercel preview deployments
- Makes the code more maintainable and less error-prone

## tested

- [x] in my local pc
- [x]  vercel preview env


MH4GF (approved):
nice!!

Issue #1897, opened 2025-06-05
Enable E2E tests for liam-erd-sample deployments


MH4GF:
## Problem
Currently, E2E tests are only executed for `@liam-hq/app` deployments. The CLI-generated ER diagrams (`liam-erd-sample`) are excluded from E2E testing, as evidenced in `.github/workflows/e2e_tests.yml`:

```ruby
elsif environment_val.include?("Preview") && \!target_url.include?("liam-erd-sample") && \!target_url.include?("liam-docs") && \!target_url.include?("liam-storybook")
```

## Proposal
Extend E2E test coverage to include `liam-erd-sample` deployments to ensure the CLI-generated ER diagrams work correctly in the deployed environment.

## Current Behavior
- E2E tests run only for `liam-app` deployments
- `liam-erd-sample` deployments are explicitly excluded from E2E testing

## Expected Behavior
- E2E tests should also run for `liam-erd-sample` deployments
- This ensures that CLI-generated ER diagrams function correctly when deployed

## Implementation Notes
The condition in `.github/workflows/e2e_tests.yml` line 46 needs to be updated to allow E2E tests for `liam-erd-sample` deployments while maintaining the current exclusions for `liam-docs` and `liam-storybook`.

PR #1174, opened 2025-04-07
🔧 Fix conditional logic to ensure expected tests run

FunamaYukina:
## Why is this change needed?
This PR fixes an issue with the conditional logic used to determine whether CI tests should run after a deployment.

### 🐛 Problem
Due to incorrect use of wildcard matching ([*]) in the Bash conditional statement, some test workflows were unintentionally skipped — even when they should have been triggered.

ss 3051

### 🔧 Fix
The Bash condition was rewritten to avoid wildcard matching syntax, using [[ ... == *string* ]] instead of [ ... = *string* ], which doesn't work as expected in POSIX sh or basic bash conditionals.

### ✅ Test

I have confirmed that the e2e test ran on this pull request checks.
https://github.com/liam-hq/liam/actions/runs/14299339854/job/40071025611?pr=1174

ss 3052

### 🤖 Generated by PR Agent at 81c9d6fc8539ec9a04237657ec9091b69392823b

- Fixed incorrect Bash conditional syntax in `e2e_tests.yml`.
- Replaced `[ ... = *string* ]` with `[[ ... == *string* ]]` for better compatibility.
- Ensured CI tests trigger correctly after deployments.
- Improved readability and consistency of conditional logic.

## Detailed Changes

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>e2e_tests.yml</strong><dd><code>Fix and enhance conditional logic in CI workflow</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

.github/workflows/e2e_tests.yml

<li>Updated Bash conditional syntax for better compatibility.<br> <li> Replaced single brackets with double brackets in conditions.<br> <li> Improved logic to correctly handle deployment events and environments.

</details>

  </td>
  <td><a href="https://github.com/liam-hq/liam/pull/1174/files#diff-37aed4ad4a8c3a4220c5e3c2eee33093073b8d64f480cc742b63764c859b6c72">+3/-3</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></td></tr></tr></tbody></table>

## Additional Notes

___

[...]

PR #3330, opened 2025-09-04
🔧 Exclude liam-assets from E2E test triggers

hoshinotsuyoshi:
## Issue

- resolve: N/A

## Why is this change needed?
Add liam-assets to the exclusion list for E2E test deployment triggers to prevent unnecessary test runs for asset-only deployments. This optimizes the CI pipeline by avoiding E2E tests when only asset files are deployed.

## Summary by CodeRabbit

* **Tests**
  * Refined preview deployment criteria for E2E runs to exclude asset-only previews, reducing unnecessary test executions.
  * Improves CI efficiency and speeds up feedback on relevant previews.
  * No impact on production test gating or behavior.
* **Chores**
  * Updated workflow conditions to better target meaningful preview environments for automated testing.

### DIFF
--- a/.github/workflows/e2e_tests.yml
+++ b/.github/workflows/e2e_tests.yml
@@ -46,7 +46,7 @@ jobs:
               if (environment_val.include?("Production") && target_url.include?("liam-app-git-main"))
                 "should_run=true"
               # Preview deployment
-              elsif environment_val.include?("Preview") && !target_url.include?("liam-erd-sample") && !target_url.include?("liam-docs") && !target_url.include?("liam-storybook")
+              elsif environment_val.include?("Preview") && !target_url.include?("liam-erd-sample") && !target_url.include?("liam-docs") && !target_url.include?("liam-storybook") && !target_url.include?("liam-assets")
                 "should_run=true"
               else
                 "should_run=false"


.github/workflows/e2e_tests.yml
```diff
@@ -46,7 +46,7 @@ jobs:
               if (environment_val.include?("Production") && target_url.include?("liam-app-git-main"))
                 "should_run=true"
               # Preview deployment
-              elsif environment_val.include?("Preview") && !target_url.include?("liam-erd-sample") && !target_url.include?("liam-docs") && !target_url.include?("liam-storybook")
+              elsif environment_val.include?("Preview") && !target_url.include?("liam-erd-sample") && !target_url.include?("liam-docs") && !target_url.include?("liam-storybook") && !target_url.include?("liam-assets")
```
hoshinotsuyoshi, line 49: naive denylist 🙏

NoritakaIkeda: I don’t have deep expertise, but I thought one possible approach could be to shift the responsibility of deciding whether to run E2E tests to the PR event side (e.g., by using PR labels), and then let the deployment_status workflow just check for the presence of that label.

However, if we do that, it might become unclear from just looking at .github/workflows/e2e_tests.yml what the actual run conditions are, so I feel the readability could suffer.

My intuition is that the current approach is actually safer to operate with than expressing the logic through PR labels.

Q: Is there a reason why you’re not taking the allowlist approach?

NoritakaIkeda (approved):
I’ve left a comment regarding my concerns.
Since I’d like to hear other perspectives as well, I’d prefer to wait for additional reviews — for example from @MH4GF.

.github/workflows/e2e_tests.yml
MH4GF: I would prefer a simple allow-list format if possible, but it seems to be a challenging problem, as Kaji-san has been stuck on it quite a bit in the past: https://github.com/liam-hq/liam/issues/1897 
It would be good to try to tackle it again!

MH4GF (approved):
LGTM 👍🏻

.github/workflows/e2e_tests.yml
hoshinotsuyoshi: Maybe `elsif environment_val.include?("Preview") && environment_val.include?("liam-app")` would actually be enough, but I’m not entirely sure about the `"liam-app"` part. Do you remember how it was? cc: @FunamaYukina

ref. https://github.com/liam-hq/liam/pull/1174

Might be a good idea to open a test PR and add some print debugging. I might give that a try.

FunamaYukina: Thank you. I think using environment_val is a good idea!👍
（I wrote it in a negative way because the `github.event.deployment_status.target_url` for the liam-app is difficult to validate. It's in a format like `https://liam-7s1mczcth-liambx.vercel.app/`, so it's not easy to create a specific check for it）

PR #3352, opened 2025-09-05
🔧 Add npm update workaround for trusted publishing in release workflow

hoshinotsuyoshi:
## Issue

- resolve:N/A

## Summary by CodeRabbit

* **Chores**
  * Ensured the release pipeline updates npm to a compatible version before publishing, reducing publishing failures and improving release reliability.
  * No user-facing changes; app behavior and APIs remain unchanged.
  * This is a temporary workaround to meet trusted-publishing requirements and does not affect local development or runtime behavior.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,12 +26,6 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
-      # NOTE: Temporary workaround
-      # - "Trusted publishing" requires npm 11.5.1+.
-      #   See: https://docs.npmjs.com/trusted-publishers
-      - name: Update npm
-        run: npm update -g npm
-
       - name: Create GitHub App Token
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -26,6 +26,12 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
+      # NOTE: Temporary workaround
+      # - "Trusted publishing" requires npm 11.5.1+.
+      #   See: https://docs.npmjs.com/trusted-publishers
+      - name: Update npm
+        run: npm update -g npm
+
```
MH4GF: Perhaps it should be after pnpm-setup on line 48. There is a concern that npm will be overwritten there.

hoshinotsuyoshi: ah, please wait

hoshinotsuyoshi: moved 🙏  https://github.com/liam-hq/liam/pull/3352/commits/8115dc6f52bf55bc156bd0a8dc5ed50234f55cba

PR #3366, opened 2025-09-06
🔧 chore: set `save-exact=true` to install packages with exact versions

tnyo43:
## Issue

~- resolve:~

## Why is this change needed?

We always manually remove `^` the package versions in package.json so that dependencies are installed with exact versions.
This PR will [set `save-exact=true` in .npmrc](https://docs.npmjs.com/cli/v9/using-npm/config?v=true#save-exact) to force exact version installs automatically.

When we run `pnpm install jest --workspace-root`, the package.json will be modified as like:

```jsonc
// before
"dependencies": {
  "jest": "^30.1.3"
}

// after
"dependencies": {
  "jest": "30.1.3"
}
```

## Summary by CodeRabbit

- Chores
  - Configured package manager to save dependencies with exact versions, ensuring consistent installations across environments.
  - No user-facing functionality changes.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+save-exact=true


MH4GF (approved):
Thanks 😄

PR #3390, opened 2025-09-08
🐛 Fix SSR error in useStream hook

junkisai:
## Issue

- resolve: SSR error "sessionStorage is not defined" in useStream hook

## Why is this change needed?

The sessionStorage API was being accessed during server-side rendering in the useState initializer, causing Next.js SSR to fail. This fix moves the sessionStorage access to a useEffect hook, ensuring it only runs on the client side while maintaining the same functionality of loading and applying the initial message from session storage.

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/c1226c0f-5e6e-4570-9c3d-5b098f19b925" /> | <video src="https://github.com/user-attachments/assets/cba542df-2b13-46a2-8c37-11a201e64da4" /> | 

## Summary by CodeRabbit

* **Bug Fixes**
  * Restores any saved initial message reliably when opening or switching design sessions, ensuring it appears once and in the correct order.
  * Adds the stored message only if it’s a user message, preventing accidental inclusion of non-user content.
  * Eliminates intermittent issues where the initial message was missing or duplicated.
  * Improves consistency by loading the message after the conversation initializes. No changes to how you use the feature.

### DIFF
--- a/frontend/apps/app/components/SessionDetailPage/hooks/useStream/useStream.ts
+++ b/frontend/apps/app/components/SessionDetailPage/hooks/useStream/useStream.ts
@@ -7,7 +7,7 @@ import {
 } from '@langchain/core/messages'
 import { SSE_EVENTS } from '@liam-hq/agent/client'
 import { err, ok } from 'neverthrow'
-import { useCallback, useRef, useState } from 'react'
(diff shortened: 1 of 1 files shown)


hoshinotsuyoshi (commented):
seems good!

PR #3403, opened 2025-09-09
⬆️ Update LangChain dependencies to support improved tool calling

junkisai:
## Issue

- resolve: Update LangChain packages to address warning about tool calling efficiency

## Why is this change needed?

This update addresses a warning about new LangChain packages being available that more efficiently handle tool calling. The warning specifically recommended upgrading packages like `@langchain/anthropic` and `@langchain/openai` to versions that properly set message tool calls.

### Changes made:
- Updated `@langchain/core` from 0.3.72 to 0.3.73
- Updated `@langchain/openai` from 0.6.9 to 0.6.10

These patch version updates include improvements for tool calling efficiency while maintaining backward compatibility.

### DIFF
--- a/frontend/apps/app/package.json
+++ b/frontend/apps/app/package.json
@@ -10,7 +10,7 @@
     "@codemirror/state": "6.5.2",
     "@codemirror/view": "6.38.1",
     "@electric-sql/pglite": "0.3.7",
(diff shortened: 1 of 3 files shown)


MH4GF (approved):
👍🏻

PR #3413, opened 2025-09-10
refactor: centralize useExhaustiveDependencies rule to root biome config

FunamaYukina:
## Issue

- resolve: https://github.com/liam-hq/liam/issues/2317

## Why is this change needed?

Previously, we had individually configured the `useExhaustiveDependencies` rule in three packages (`apps/app`, `erd-core`, and `ui`) to ensure proper React hook dependency tracking. Now that all three packages have successfully adapted to and are compliant with this rule, we can centralize the configuration.

By moving this rule to the root `biome.jsonc`, we:
- Eliminate duplicate configurations across the three package-level `biome.json`c files
- Establish a single source of truth for this rule that applies project-wide
- Reduce maintenance overhead and ensure consistent enforcement going forward

This centralization is now possible because all affected packages have completed their migration to comply with the `useExhaustiveDependencies` rule.

## Summary by CodeRabbit

- Chores
  - Standardized linting settings across the frontend codebase for consistency.
  - Enforced stricter dependency checks in shared/internal configurations.
  - Removed a noisy dependency-related rule from app and package configs to reduce false positives.
  - No changes to runtime behavior, UI, or user workflows.
  - Improves developer experience and code quality without affecting end-users.

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -3,8 +3,7 @@
   "linter": {
     "rules": {
       "correctness": {
-        "noUndeclaredDependencies": "off",
-        "useExhaustiveDependencies": "error"
+        "noUndeclaredDependencies": "off"
       }
     }
   }
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -3,7 +3,8 @@
   "linter": {
     "rules": {
       "correctness": {
-        "noUndeclaredDependencies": "off"
+        "noUndeclaredDependencies": "off",
+        "useExhaustiveDependencies": "error"
       }
     }
   }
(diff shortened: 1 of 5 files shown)


MH4GF (commented):
If the goal is to standardize the rules, please place them in the configs package.


https://github.com/liam-hq/liam/blob/8305328911c832e4df9a9326424a7080add68243/frontend/internal-packages/configs/biome.jsonc

FunamaYukina:
Oh, sorry! I'll modify the biome.jsonc file in the internal-package directory.🙏

FunamaYukina:
@MH4GF I'm sorry for the mistake. I've corrected it.🙏

PR #3453, opened 2025-09-11
Fix: Foreign key relationship missing visualization when same names

huuyafwww:
## Issue

- resolve:In SQLite, foreign key relationships with the same name are not visualized.

## Why is this change needed?

Thank you for developing this excellent library for visualizing database ERDs.
I am currently using SQLite, and I have set up foreign keys in that database.
In SQLite, it is not necessary to name foreign key constraints.
I am using this library via tbls,and when foreign keys are not named, the constraint names start at 0 and increment by 1 as follows:
```
- (Foreign key ID: 0)
- (Foreign key ID: 1)
- (Foreign key ID: 2)
```
This naming follows the same rule regardless of the table name.

In such cases,Liam treats foreign keys with identical constraint names as the same foreign key, regardless of the table.
This causes problems in this scenario.
As you can see in the modified section, the value of the relationshipKey corresponding to the relationship gets overwritten repeatedly.
Therefore, I modified it so that even if the constraint names are the same, they are treated as separate relationships if they belong to different tables.

Additionally, this change now supports visualization of the schema.json corresponding to tbls's official sqlite, so we added the corresponding json and README.

Please let me know if you have any issues with this PR change.

## Summary by CodeRabbit

- New Features
  - Added a comprehensive SQLite schema sample (schema.json) covering tables, views, virtual tables, constraints, triggers and relationships for parser/validation.

- Refactor
  - Standardized relationship key/name generation to include the related table and an index for consistent, non-colliding identifiers.

- Documentation
  - Added README noting the sample’s origin and MIT license; added changeset entry describing a foreign-key visualization fix.

- Tests
  - Updated and added tests to expect the new naming scheme and ensure unique relationship keys.

### DIFF
--- a/.changeset/foreign-key-relationship-missing-visualization-when-same-names.md
+++ b/.changeset/foreign-key-relationship-missing-visualization-when-same-names.md
@@ -1,0 +1,6 @@
+---
+"@liam-hq/schema": patch
+---
+
+- 🐛 Fixed a bug where foreign key relationships with the same name were not visualized.
+  - Modified to be recognized as a unique foreign key.


frontend/packages/schema/src/utils/constraintsToRelationships.ts
```diff
@@ -57,8 +57,7 @@ export const constraintsToRelationships = (tables: Tables): Relationships => {
         }
 
         // For composite keys, append index to make unique relationship names
-        const relationshipKey =
-          columnCount > 1 ? `${constraint.name}_${i}` : constraint.name
+        const relationshipKey = `${table.name}_${constraint.name}_${i}`
```
huuyafwww, line 60: We have modified the foreign key constraint name so that the key remains unique even if the constraint name is the same across multiple different tables.

frontend/packages/schema/src/parser/tbls/input/sqlite/schema.json
```diff
@@ -0,0 +1,830 @@
...
+        },
+        {
+          "name": "- (Foreign key ID: 1)",
+          "type": "FOREIGN KEY",
+          "def": "FOREIGN KEY (post_id) REFERENCES posts (id) ON UPDATE NO ACTION ON DELETE NO ACTION MATCH NONE",
+          "table": "comments",
+          "referenced_table": "posts",
+          "columns": ["post_id"],
+          "referenced_columns": ["id"]
+        },
```
huuyafwww, line 312: As mentioned in the PR body, when using tbls, the naming convention for foreign key constraints follows this pattern.

frontend/packages/schema/src/parser/tbls/input/sqlite/README.md
MH4GF, line 1: 👍🏻

huuyafwww:
@MH4GF 
The following commit contains the fix!
It adds new test cases that account for cases where the same constraint name exists in separate tables, and modifies existing tests.
https://github.com/liam-hq/liam/pull/3453/commits/bfafe7d3d45b53fb1945b3a9e6cc99c330aab616

MH4GF (changes requested):
Thanks!! Sorry, could you create a changeset for the end?
Since this is a bug fix for the schema package, a patch is fine.

ref: https://github.com/liam-hq/liam/blob/c4db34da7eab1249df65125acd192a3b968874fd/docs/changeset-guide.md

huuyafwww:
@MH4GF 
Thanks for checking it out so quickly.
I've added the doc to the changeset!
https://github.com/liam-hq/liam/pull/3453/commits/2b5dd89bd13e782717bcb6f97b852a9ecd577597

MH4GF (approved):
LGTM 👍🏻 Thanks for your first contribution!!

PR #3493, opened 2025-09-16
🔒(middleware): Block direct access to /erd/* paths on app.liambx.com

hoshinotsuyoshi:
## Issue

- resolve: route06/liam-internal#5578

## Why is this change needed?

Currently, users can access the ERD viewer through two different URLs:
- `https://liambx.com/erd/*` (intended entry point)
- `https://app.liambx.com/erd/*` (should be blocked)

Having multiple access points creates user confusion. This change ensures /erd/* paths can only be accessed through the liambx.com domain.

## Implementation

Added middleware validation to check for a custom header (`x-liam-rewrite-source`) that is only set when requests come through the liambx.com rewrite. Direct access to app.liambx.com/erd/* will now return a 403 error with a message directing users to the correct URL.

Note: The corresponding change to add the custom header in the service-site repository needs to be deployed for this to work properly.


MH4GF (approved):
👍🏻 
I leave the merge work to you!

hoshinotsuyoshi:
@MH4GF Sorry! I couldn’t quickly figure out how to add a custom header on the service site side 😇 . For now, I’d like to switch to using `x-forwarded-host` instead! https://github.com/liam-hq/liam/pull/3493/commits/1a62bb53584681ee11d5e7e1a46612d9cfd1f28c

MH4GF (approved):
👍🏻👍🏻👍🏻

hoshinotsuyoshi:
thx!

PR #3499, opened 2025-09-16
💄 show `BlinkCircle` indicator when column element is selected in `TableDetail`

tnyo43:
## Issue

~- resolve:~

## Why is this change needed?

This PR introduces BlinkCircle and show emphasized boarders to clearly indicate when a column element is selected. I think it’s especially helpful when users open a page via a URL that contains a hash value.

### demo

preview URL: https://liam-erd-sample-ef7r5609z-liambx.vercel.app/?active=tags#tags__columns__name

When you open a page via a URL that contains a hash value.

https://github.com/user-attachments/assets/7ff1a9da-086e-46e9-97e7-4b0b57b5a10d

When you select a column element

https://github.com/user-attachments/assets/2e2090fd-ad3b-416d-ac3a-ba196be75c80

## Summary by CodeRabbit

* **New Features**
  * Adds an animated circular indicator that appears next to a selected column in Table Details.
  * Selection highlighting follows the URL fragment (hash) and updates when the fragment changes.

* **Style**
  * Minor spacing and layout tweaks for column and collapsible header alignment.
  * Adds animation styles for the blink indicator.

* **Tests**
  * Adds tests verifying the indicator renders for the focused column and not for others.

* **Chores**
  * Adds a changeset entry for a patch release of the ERD core package.

### DIFF
--- a/frontend/packages/erd-core/src/features/erd/components/ERDContent/components/TableNode/TableDetail/BlinkCircle/BlinkCircle.module.css
+++ b/frontend/packages/erd-core/src/features/erd/components/ERDContent/components/TableNode/TableDetail/BlinkCircle/BlinkCircle.module.css
@@ -1,0 +1,69 @@
+.container {
+  position: relative;
+}
+
+.container:before {
+  position: absolute;
+  display: block;
+  content: '';
+  border-radius: 50%;
+  transform: translate(-50%, -50%);
+  background-color: var(--primary-accent);
+  animation: sizePulse 2700ms forwards;
+}
+
+@keyframes sizePulse {
+  /* 0/7 */
+  0% {
+    width: 0;
+    height: 0;
+  }
+
+  /* 1/7 */
+  14% {
+    width: 10px;
+    height: 10px;
+    animation-timing-function: ease-out;
+  }
+
+  /* 2/7 */
+  28% {
+    width: 10px;
+    height: 10px;
+  }
+
+  /* 3/7 */
+  43% {
+    width: 3px;
+    height: 3px;
+    animation-timing-function: ease-out;
+  }
+
+  /* 4/7 */
+  57% {
+    width: 3px;
+    height: 3px;
+  }
+
+  /* 5/7 */
+  71% {
+    width: 10px;
+    height: 10px;
+    animation-timing-function: ease-out;
+  }
+
+  /* 6/7 */
+  86% {
+    width: 10px;
+    height: 10px;
+    opacity: 1;
+  }
+
+  /* 7/7 */
+  100% {
+    width: 0px;
+    height: 0px;
+    opacity: 0;
+    animation-timing-function: ease-out;
+  }
+}
--- a/frontend/packages/erd-core/src/features/erd/components/ERDContent/components/TableNode/TableDetail/BlinkCircle/BlinkCircle.module.css
+++ b/frontend/packages/erd-core/src/features/erd/components/ERDContent/components/TableNode/TableDetail/BlinkCircle/BlinkCircle.module.css
@@ -9,7 +9,7 @@
   border-radius: 50%;
   transform: translate(-50%, -50%);
   background-color: var(--primary-accent);
-  animation: sizePulse 2700ms forwards;
+  animation: sizePulse 5400ms forwards;
 }
 
 @keyframes sizePulse {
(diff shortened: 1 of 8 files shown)


frontend/packages/erd-core/src/features/erd/components/ERDContent/components/TableNode/TableDetail/BlinkCircle/BlinkCircle.module.css
```diff
@@ -0,0 +1,69 @@
...
+}
+
+.container:before {
+  position: absolute;
+  display: block;
+  content: '';
+  border-radius: 50%;
+  transform: translate(-50%, -50%);
+  background-color: var(--primary-accent);
+  animation: sizePulse 2700ms forwards;
```
tnyo43: This length of time, 2700ms, feels good to me.
Let me know if you have any suggestions 😉

frontend/packages/erd-core/src/features/erd/components/ERDContent/components/TableNode/TableDetail/Columns/Columns.tsx
```diff
@@ -10,6 +10,22 @@ type Props = {
 }
 
 export const Columns: FC<Props> = ({ table }) => {
+  const [focusedElementId, setFocusedElementId] = useState(
+    // location.hash starts with '#'; decode to match actual DOM id
+    location.hash.slice(1),
+  )
```
tnyo43, line 16: It requires a state of hash to render BlinkCircle component with conditions.

tnyo43: I’m still considering the best place to handle this state. It might belong in the `UserEditingProvider` or the `TableDetail` component, but for now let’s keep it here and make a decision in a future update.

frontend/packages/erd-core/src/features/erd/components/ERDContent/components/TableNode/TableDetail/BlinkCircle/BlinkCircle.module.css
MH4GF: Personally, it didn't stand out much and seemed to fail at effectively drawing attention. I'd like to emphasize it a bit more.
How about increasing the size while also boosting the flash count from 2 to 4 times? Since 2 cycles at 2700ms seems good, what about 4 cycles at 5400ms?

@kumanoayumi I'd appreciate your feedback if possible.

kumanoayumi: @tnyo43 @MH4GF 

I think increasing the cycles is a good idea. Since the FocusRing showing on TableDetail is one of the factors, how about scoping the FocusRing to the relevant section?

tnyo43: Thanks for the feedbacks. Honestly, I was get used to with the indicator and that I became sensitive enough to understand it easily 😅 
I'll try you ideas!

tnyo43: I feel that relying only on `BlinkCircle` isn’t ideal, since it can be hard to notice focuses when people miss the first a few seconds. So I think the focusring approach is good 👍 
I also increase the cycles, I feel that is also fine.

preview: https://liam-erd-sample-nqv0g29by-liambx.vercel.app/?active=tags#tags__columns__name

(The ring looks gray in the screen capture, but it's green actually 😅. I'm sorry but I can fixed it)

Screenshot 0007-09-26 at 8 39 26

https://github.com/user-attachments/assets/fddf9b9f-3dde-49be-828d-b34a6a78ce26

tnyo43: FYI: I also considered changing the background color, but I found it difficult to keep enough contrast with the table’s background. It would be nice if we could find a good color for it.

(the focus color in the image is `--pane-background-active`)
Screenshot 0007-09-26 at 8 36 58

frontend/packages/erd-core/src/features/erd/components/ERDContent/components/TableNode/TableDetail/CollapsibleHeader/CollapsibleHeader.module.css
```diff
@@ -51,6 +51,6 @@
   transition: max-height 0.1s var(--default-timing-function);
 
   /* enable to scroll to the top children element, */
-  padding-top: 40px;
-  margin-top: -40px;
+  padding-top: 41px;
+  margin-top: -41px;
```
tnyo43, line 55: It looks 41px is right height of the header.
This change make the focus ring be displayed well.

https://github.com/liam-hq/liam/blob/64017c3ebec3156556d5e14fd219bf29f4068fb6/frontend/packages/erd-core/src/features/erd/components/ERDContent/components/TableNode/TableDetail/Indexes/Indexes.tsx#L20-L22

without this change:
Image

tnyo43:
I'll fix the ci fail if the current changes are approved 🙇

frontend/packages/erd-core/src/features/erd/components/ERDContent/components/TableNode/TableDetail/BlinkCircle/BlinkCircle.module.css
MH4GF: I think the revised approach is much clearer now! I'm LGTM 👍🏻 @kumanoayumi please take a look too!

I noticed the focus ring on TableDetail is blue. Could this also be fixed separately? @tnyo43

kumanoayumi: @tnyo43 Thank you for your support! It’s much clearer now, and I think it’s become really good!

tnyo43: @MH4GF @kumanoayumi Thank you for the comments!
I’ll go with the approach.

> I noticed the focus ring on TableDetail is blue. Could this also be fixed separately?

Sure 👍

tnyo43: @MH4GF 

I made a PR to improve the TableDetail drawer accessibility. This change will remove the focus ring on the component (actually I don't understand why it happened exactly). Could you please check this as well? We can continue considering how to handle this issue with this change in mind 🙏 
https://github.com/liam-hq/liam/pull/3626

MH4GF (approved):
LGTM 👍🏻 Great work!!

PR #3517, opened 2025-09-18
🐛(schema): Fix jsonb default value escaping in PostgreSQL deparser

hoshinotsuyoshi:
## Issue

- resolve: https://github.com/route06/liam-internal/issues/5682

## Why is this change needed?

The PostgreSQL deparser was incorrectly handling jsonb default values, causing syntax errors when generating DDL. The issue occurred with different jsonb default value formats that LLMs might generate:

1. Cast expressions like `'{}'::jsonb` were being double-escaped
2. Pre-quoted values like `'{}'` were being double-escaped
3. Unquoted values like `{}` needed proper quoting

This fix ensures all three formats are handled correctly without double-escaping, preventing SQL syntax errors during DDL execution.

## Summary by CodeRabbit

* **Bug Fixes**
  * Correctly deparses JSONB default values in generated DDL, respecting cast expressions and already-quoted literals to prevent double-escaping.
  * Ensures function-based defaults (e.g., UUID generators) remain unquoted in output.

* **Tests**
  * Added comprehensive tests for JSONB defaults (objects, arrays, and cast expressions) in table definitions.
  * Verifies zero deparser errors and exact SQL output matching the expected DDL.

### DIFF
--- a/frontend/packages/schema/src/deparser/postgresql/schemaDeparser.test.ts
+++ b/frontend/packages/schema/src/deparser/postgresql/schemaDeparser.test.ts
@@ -1773,6 +1773,64 @@ describe('postgresqlSchemaDeparser', () => {
         expect(result.value).not.toContain("DEFAULT 'gen_random_uuid()'")
         expect(result.value).not.toContain("DEFAULT 'now()'")
       })
+
+      it('should handle JSONB default values with cast expressions correctly', () => {
+        const schema = aSchema({
+          tables: {
+            users: aTable({
+              name: 'users',
+              columns: {
+                id: aColumn({
+                  name: 'id',
+                  type: 'uuid',
+                  notNull: true,
+                  default: 'gen_random_uuid()',
+                }),
+                metadata: aColumn({
+                  name: 'metadata',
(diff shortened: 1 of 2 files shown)


NoritakaIkeda (approved):
> Cast expressions like `'{}'::jsonb` were being double-escaped
> Pre-quoted values like `'{}'` were being double-escaped
> Unquoted values like `{}` needed proper quoting

Looks like there were notation/format inconsistencies, and this change covers them comprehensively! Thanks for adding the tests—excellent fix. LGTM!

NoritakaIkeda:
@hoshinotsuyoshi
I'd appreciate it if you could just take a quick look at the E2E being down. I suspect a retry will fix it.

hoshinotsuyoshi:
thx! I'll check that now

hoshinotsuyoshi:
e2e passes https://github.com/liam-hq/liam/actions/runs/17819702006/job/50660292699 👀  👍  🐱

MH4GF:
@hoshinotsuyoshi This is a great improvement! Thank you!!

PR #3527, opened 2025-09-19
✨(commands): add deparse-dataset-to-schema command

NoritakaIkeda:
## Issue

- resolve: N/A

## Why is this change needed?

This change adds a new Claude Code command that enables users to convert JSON schema datasets back to SQL DDL format using the existing `@liam-hq/schema` deparser functionality.

### Key Features

- **JSON to SQL conversion**: Leverages `@liam-hq/schema`'s `postgresqlSchemaDeparser` for accurate DDL generation
- **Flexible file paths**: Accepts any input JSON file path and output SQL file path
- **Format support**: Currently supports PostgreSQL with extensible design for future formats
- **Comprehensive examples**: Includes realistic usage examples with actual file paths

### Use Cases

- Converting benchmark datasets to SQL for testing and validation
- Generating SQL schemas from JSON representations
- Supporting reverse engineering workflows
- Facilitating schema format conversions

### Command Usage

```bash
/deparse-dataset-to-schema input.json output.sql
/deparse-dataset-to-schema data/case-001.json schema.sql --format postgres
```

This complements the existing schema parsing capabilities by providing the reverse operation, enabling bidirectional conversion between JSON schema representations and SQL DDL.

## Summary by CodeRabbit

- Documentation
  - Added a guide for a CLI command to convert JSON benchmark datasets into SQL schema files.
  - Includes step-by-step process, required arguments (input JSON path, output path), and an optional format flag.
  - Provides example invocations and notes on validating the generated schema.
  - Currently supports PostgreSQL format (default).

### DIFF
--- a/.claude/commands/deparse-dataset-to-schema.md
+++ b/.claude/commands/deparse-dataset-to-schema.md
@@ -1,0 +1,31 @@
+---
+description: Convert JSON benchmark datasets back to schema files
+---
+
+# Deparse Dataset to Schema
+
+## Task
+
+Convert a JSON schema dataset to a schema file in the specified format using @liam-hq/schema deparser functionality.
+
+### Process
+
+1. Read the JSON dataset from the specified input path
+2. Use @liam-hq/schema's `postgresqlSchemaDeparser` function to convert JSON to SQL DDL
+3. Save the generated schema file to the specified output path
+4. Verify the generated schema is valid and properly formatted
+
+### Arguments
+
+$ARGUMENTS
+
+Expected arguments:
+
+- Input JSON file path - path to the JSON schema dataset to convert
+- Output file path - where to save the generated schema file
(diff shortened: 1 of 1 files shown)


MH4GF (approved):
@NoritakaIkeda
This task doesn't seem suited to leveraging the non-deterministic nature of LLMs; it would be better to build it as a deterministic tool using a program. This is because it's a task where the same input should always produce the same output.

Specifically, I envision creating a script like the following. I think this could be done with Vibe Coding.

```
$ pnpm dlx tsx deparse-dataset-to-schema data/case-002.json output/schema.sql
```

However, I think introducing this command for now is a good idea.



Translated with [DeepL.com](https://www.deepl.com/?utm_campaign=product&utm_source=web_translator&utm_medium=web&utm_content=copy_free_translation) (free version)

PR #3530, opened 2025-09-19
🐛(agent): Fix PM Agent infinite loop on tool success

MH4GF:
## Issue

- resolve: Prevents infinite loop in PM Agent when artifact save tool succeeds

## Why is this change needed?

The PM Agent was experiencing an infinite loop where after successfully saving an artifact, it would continue to loop back to the analyze requirements node instead of properly terminating. This change introduces a flow control mechanism to track when the artifact save is successful and properly end the workflow.

### Changes Made:
- Added  flag to PM Agent state to track save success
- Modified  to set the flag to true on successful completion
- Updated routing logic in  to check the flag and route to END when save is successful
- Updated all related tests to include the new state flag

### Technical Details:
- The conditional edge from  now checks 
- If true, routes to END to terminate the workflow
- If false, routes back to  for retry logic
- This prevents the previous behavior where it would always route back to analyze requirements

## Summary by CodeRabbit

- New Features
  - Adds a success flag to the agent’s state to track whether saving requirements succeeded.
  - Workflow now conditionally ends after a successful save or returns to requirements analysis on failure.
- Documentation
  - Updated workflow diagram to reflect the new conditional routing paths.
- Refactor
  - Streamlined analysis prompts by removing verbosity guidance and stop-condition text.
- Tests
  - Updated tests to include the new success flag in initial state and to validate the revised workflow routing.

### DIFF
--- a/frontend/internal-packages/agent/src/pm-agent/prompts/pmAnalysisPrompts.ts
+++ b/frontend/internal-packages/agent/src/pm-agent/prompts/pmAnalysisPrompts.ts
@@ -72,16 +72,8 @@ The current schema structure will be provided:
 ### Non-Functional Requirements
 - Capture HOW WELL the system performs (quality, security, performance, etc.)
 
-# Workflow Completion
-- After requirements are saved, validate the save was successful in 1–2 lines and report success (e.g., "Requirements analysis complete and saved to artifact."). Do not suggest further steps or invite additional requirements, as the next step will be for DB Agent to perform database design in accordance with the requirements.
-- Do not call the save tool again unless the user requests new analysis.
-- End tool use and respond in text only.
-
 # Verbosity
 - Use concise summaries. For requirements and code, provide clear, structured outputs.
-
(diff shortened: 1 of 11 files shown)


frontend/internal-packages/agent/src/pm-agent/prompts/pmAnalysisPrompts.ts
```diff
@@ -72,16 +72,8 @@ The current schema structure will be provided:
 ### Non-Functional Requirements
 - Capture HOW WELL the system performs (quality, security, performance, etc.)
 
-# Workflow Completion
-- After requirements are saved, validate the save was successful in 1–2 lines and report success (e.g., "Requirements analysis complete and saved to artifact."). Do not suggest further steps or invite additional requirements, as the next step will be for DB Agent to perform database design in accordance with the requirements.
```
FunamaYukina, line 76: Thank you!

PR #3560, opened 2025-09-21
🐛 fix(command palette): fix Show Mode commands descriptions, ⌘ -> ⇧

tnyo43:
## Issue

~- resolve:~


frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteOptions/CommandOptions.tsx
```diff
@@ -58,7 +58,8 @@ export const CommandPaletteCommandOptions: FC = () => {
         }}
         className={styles.item}
       >
-        <TidyUpIcon className={styles.itemIcon} />
+        {/* The option has explicit text description, so that the icon does not have to have an aria-label */}
+        <TidyUpIcon className={styles.itemIcon} aria-label={undefined} />
```
tnyo43, line 62: This is unrelated to the bug, but I fixed it to make the test case.
[The `TidyUpIcon` itself has an aria-label attribute value "tidy up"](https://github.com/liam-hq/liam/blob/eb42eceb8c7791bcf3bdc36dedbd24af59f9b6ed/frontend/packages/ui/src/icons/TidyUpIcon.tsx#L9), however it makes the option name as "tidy up Tidy Up ⇧ T", which may not make sense. So I added `aria-label={undefined}` to remove that property.
I can fix the `TidyUpIcon` if needed, but as another issue.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteOptions/CommandOptions.tsx
```diff
@@ -73,7 +74,7 @@ export const CommandPaletteCommandOptions: FC = () => {
       >
         <PanelTop className={styles.itemIcon} />
         <span className={styles.itemText}>Show All Fields</span>
-        <span className={styles.keyIcon}>⌘</span>
+        <span className={styles.keyIcon}>⇧</span>
```
tnyo43, line 77: The right commands are ⇧2, 3, 4 😓 
ref: https://github.com/liam-hq/liam/blob/a683c9375e19464e9d614be7d2b5c4a5bb50b70f/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/hooks/useSubscribeCommands/useSubscribeSwitchShowMode.ts#L7-L19

PR #3612, opened 2025-09-26
♻️(agent): Remove unused latest version plumbing

MH4GF:
## Issue

- resolve:

## Why is this change needed?
Retires the unused latest version plumbing from the chat-to-agent workflow so the API, LangGraph state, and tests rely on repository reads instead of passing redundant values.

## Summary by CodeRabbit

- New Features
  - Clearer error message when a design session ID is missing.

- Refactor
  - Removed version-tracking from the schema design workflow to streamline processing and reduce overhead.
  - Simplified agent configuration and results returned by schema tools.

- Documentation
  - Updated usage examples and public interfaces to reflect the removal of version-tracking parameters.

- Tests
  - Updated integration and unit tests to match the new workflow shape; added coverage for missing design session ID validation.


junkisai (approved):
I see, this was an unused value.

PR #3626, opened 2025-09-28
♿️ fix(erd-core): set autofocus on TableDetail to focus whenever the component opens

tnyo43:
## Issue

~- resolve:~

## Why is this change needed?

When a user selects a TableDetail via ERD nodes or CommandPalette table options, the TableDetails drawer will opens but won't be focused. Therefore the users with only keyboards can't browse the contents.
This PR will add `autofocus` property on to the DrawerRoot and make it accessible.

### demo

In the following demos, I pressed Tab key several times after the TableDetail drawer opens.

**before**

https://github.com/user-attachments/assets/5d18891a-c447-4457-a4b3-70db3c78e808

**after**

https://github.com/user-attachments/assets/55a964b9-e25b-48ff-bb2b-d8f1b880f8da

### note

It used to focus on the TableDetail drawer when the page is accessed via URL contains active table name. This change will not affect that behaviour.

---

The focusing on the TableDetail drawser is not pretty good. Let's improve it as another issue 👍 

https://github.com/user-attachments/assets/a43b9fc5-8890-465d-ac81-050af8c07dcb

## Summary by CodeRabbit

- New Features
  - Table Details drawer now auto-focuses when opened, enabling immediate keyboard navigation and input without extra clicks. This improves accessibility, speeds up frequent edits, and reduces interaction friction whenever the panel opens.

- Chores
  - Applied a patch update to internal dependencies to maintain stability and compatibility.

### DIFF
--- a/.changeset/old-islands-pump.md
+++ b/.changeset/old-islands-pump.md
@@ -1,0 +1,5 @@
+---
+"@liam-hq/erd-core": patch
+---
+
+- ♿️ set autofocus on TableDetail to focus whenever the component opens


junkisai (approved):
nice work!!

PR #3627, opened 2025-09-28
✨ feat(command palette): support "table" SearchInput mode

tnyo43:
## Issue

~- resolve:~

## Why is this change needed?

To support column search in the CommandPalette, we would like to introduce a new input mode which filters options with specific tables.
This PR will introduce "table" input mode with a table name. It is not activatable by default, so that users will not see for a while.

### demo

I created a preview with table mode activatable -> https://liam-erd-sample-7mknrlgc9-liambx.vercel.app
You can press "Tab" key when selecting a table option to go into the table mode with it. We haven't implemented the table mode behaviours, so it will be empty, but you can see the input element changing.

https://github.com/user-attachments/assets/ca89d009-6d23-48fa-b32d-035f1d8eaf12

## Summary by CodeRabbit

* **New Features**
  * Command palette shows contextual suggestions (table or command) and can enter a table/column mode that displays "tableName /" when accepted.
  * Tab (when enabled) can accept a table suggestion and switch to table mode, clearing the input; '>' still triggers command mode.

* **Tests**
  * Added tests for suggestion-driven behavior, Tab acceptance, Backspace transitions, and related mode edge cases.

### DIFF
--- a/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteContent/CommandPaletteContent.test.tsx
+++ b/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteContent/CommandPaletteContent.test.tsx
@@ -157,6 +157,23 @@ describe('input mode and filters', () => {
       expect(screen.queryByRole('option')).not.toBeInTheDocument()
       expect(screen.getByText('No results found.')).toBeInTheDocument()
     })
+
+    // TODO: implement table input mode and remove this test
+    it('does not switch to "table" mode when pressing Tab key', async () => {
+      const user = userEvent.setup()
+      render(<CommandPaletteContent />, { wrapper })
+
+      // switch to "command" input mode
+      await user.keyboard('{Tab}')
+
+      // command options still be displayed because table mode is not activated
+      expect(
+        screen.getByRole('option', { name: 'Copy Link ⌘ C' }),
+      ).toBeInTheDocument()
+      expect(
+        screen.getByRole('option', { name: 'Zoom to Fit ⇧ 1' }),
+      ).toBeInTheDocument()
+    })
   })
 })
 
--- a/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteSearchInput/CommandPaletteSearchInput.tsx
+++ b/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteSearchInput/CommandPaletteSearchInput.tsx
@@ -18,12 +18,16 @@ type Props = ComponentProps<typeof Command.Input> & {
   mode: CommandPaletteInputMode
   suggestion: CommandPaletteSuggestion | null
   setMode: (mode: CommandPaletteInputMode) => void
+
+  // TODO: remove this prop and always activate table mode when releasing the feature
+  isTableModeActivatable?: boolean
 }
 
 export const CommandPaletteSearchInput: FC<Props> = ({
   mode,
   suggestion,
   setMode,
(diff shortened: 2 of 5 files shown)


frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteContent/CommandPaletteContent.test.tsx
```diff
@@ -157,6 +157,23 @@ describe('input mode and filters', () => {
...
+      await user.keyboard('{Tab}')
+
+      // command options still be displayed because table mode is not activated
+      expect(
+        screen.getByRole('option', { name: 'Copy Link ⌘ C' }),
+      ).toBeInTheDocument()
+      expect(
+        screen.getByRole('option', { name: 'Zoom to Fit ⇧ 1' }),
+      ).toBeInTheDocument()
+    })
```
tnyo43, line 176: I checked this test will fail when I add `isTableModeActivatable={true}` as a prop of [`CommandPaletteSearchInput` in `CommandPaletteContent`](https://github.com/liam-hq/liam/blob/05362c87a81f3c708865f99318c4568f29574d5e/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteContent/CommandPaletteContent.tsx#L43-L48) which mean the flag works as expected.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteSearchInput/CommandPaletteSearchInput.tsx
```diff
@@ -28,38 +38,52 @@ export const CommandPaletteSearchInput: FC<Props> = ({
...
+          // TODO: remove this condition and always activate table mode when releasing the feature
+          if (isTableModeActivatable) {
+            if (event.key === 'Tab' && suggestion?.type === 'table') {
+              // switch to "table" mode if a table is suggested and Tab key is pressed
+              event.preventDefault()
+              setMode({ type: 'table', tableName: suggestion.name })
+              setValue('')
+              return
+            }
+          }
```
tnyo43, line 67: It doesn't turn to table mode yet since `isTableModeActivatable` is `false` in the release.
I added a test case in `CommandPaletteContent` to make sure the `Tab` doesn't work.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteSearchInput/CommandPaletteSearchInput.tsx
tnyo43, line 1: I did some refactoring at the same time 🙇 
Please check the code hiding whitespaces.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteSearchInput/CommandPaletteSearchInput.tsx
```diff
@@ -72,6 +96,8 @@ export const CommandPaletteSearchInput: FC<Props> = ({
           onValueChange={setValue}
           className={styles.input}
           placeholder="Search"
+          autoFocus
+          onKeyDown={handleKeydown}
```
tnyo43, line 100: I thought adding event listeners with `useEffect` is more complicated and more dengerous, so I use `onKeyDown` instead.
`autoFocus` is not necessary in the application since cmdk originally support the `autoFocus`. However it's required to replicate the same behavior in tests.

junkisai (approved):
LGTM!

PR #3652, opened 2025-10-01
✨(agent): Implement atomic checkpoint insertion with Supabase RPC

MH4GF:
## Issue

- resolve: 

## Why is this change needed?

This change improves the reliability of checkpoint persistence by ensuring atomic insertion of checkpoints and their associated blobs. Previously, checkpoint and blob insertions were separate operations, which could lead to inconsistent state if one succeeded while the other failed.

The implementation introduces a PostgreSQL RPC function `put_checkpoint` that handles both insertions within a single database transaction, guaranteeing consistency.

**Key changes:**
- Created `put_checkpoint` RPC function in PostgreSQL to handle atomic insertions
- Updated `SupabaseCheckpointSaver` to use the RPC instead of separate insert operations
- Added comprehensive tests for transactional behavior and error handling
- Simplified blob deserialization by removing unnecessary double-encoding

### DIFF
--- a/frontend/internal-packages/agent/src/checkpoint/SupabaseCheckpointSaver.test.ts
+++ b/frontend/internal-packages/agent/src/checkpoint/SupabaseCheckpointSaver.test.ts
@@ -87,23 +87,29 @@ describe('SupabaseCheckpointSaver', () => {
       return HttpResponse.json(data)
     })
 
-  const mockSaveCheckpointResponse = () =>
-    http.post(`${SUPABASE_URL}/rest/v1/checkpoints`, () => {
-      return HttpResponse.json({ success: true })
-    })
-
-  const mockSaveBlobsResponse = () =>
-    http.post(`${SUPABASE_URL}/rest/v1/checkpoint_blobs`, () => {
-      return HttpResponse.json({ success: true })
-    })
-
   const mockSaveWritesResponse = () =>
     http.post(`${SUPABASE_URL}/rest/v1/checkpoint_writes`, () => {
       return HttpResponse.json({ success: true })
     })
 
+  const mockPutCheckpointRpc = () =>
+    http.post(`${SUPABASE_URL}/rest/v1/rpc/put_checkpoint`, () => {
+      return new HttpResponse(null, { status: 204 })
+    })
+
+  const mockPutCheckpointRpcError = (errorMessage: string) =>
+    http.post(`${SUPABASE_URL}/rest/v1/rpc/put_checkpoint`, () => {
+      return HttpResponse.json(
+        {
+          error: errorMessage,
+          message: errorMessage,
+        },
+        { status: 500 },
+      )
+    })
+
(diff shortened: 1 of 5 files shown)


frontend/internal-packages/agent/src/checkpoint/SupabaseCheckpointSaver.test.ts
```diff
@@ -636,12 +639,12 @@ describe('SupabaseCheckpointSaver', () => {
...
-        typeof capturedSaveData === 'object' &&
-        'organization_id' in capturedSaveData
-      ) {
-        expect(capturedSaveData.organization_id).toBe(TEST_ORG_ID)
+      if (capturedSaveData && typeof capturedSaveData === 'object') {
+        // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+        const data = capturedSaveData as {
+          p_checkpoint: { organization_id: string }
+        }
+        expect(data.p_checkpoint.organization_id).toBe(TEST_ORG_ID)
```
MH4GF, line 667: Since we are verifying this in a test, we have decided to allow type casting here.

frontend/internal-packages/agent/src/checkpoint/SupabaseCheckpointSaver.ts
```diff
@@ -611,13 +590,10 @@ export class SupabaseCheckpointSaver extends BaseCheckpointSaver<number> {
...
-      // Data flow: Hex (from Supabase) -> Base64 string -> Actual data
-      // All data is double-encoded through Supabase REST API
-      const hexDecoded = hexToUint8Array(blob.blob)
-      const base64String = uint8ArrayToString(hexDecoded)
-      const actualData = base64ToUint8Array(base64String)
-      const value = await this.serde.loadsTyped(blob.type, actualData)
+      // Convert BYTEA hex string from Supabase JS client to binary data
+      // Supabase returns BYTEA as hex-encoded string (e.g., "\\x48656c6c6f" becomes "48656c6c6f")
+      const binaryData = hexToUint8Array(blob.blob)
+      const value = await this.serde.loadsTyped(blob.type, binaryData)
```
MH4GF, line 596: With Supabase JS, unnecessary double-encoding occurred, but by handling it via RPC, that became unnecessary.

frontend/internal-packages/db/supabase/migrations/20251001190247_put_checkpoint.sql
```diff
@@ -0,0 +1,67 @@
+-- Function to handle checkpoint and blobs insertion atomically
+create or replace function put_checkpoint(
+  p_checkpoint jsonb,
+  p_blobs jsonb
+) returns void as $$
+begin
```
NoritakaIkeda, line 6: My understanding is that adding this DB function ensures the relationship between checkpoints and blobs is preserved. 👍

PR #3656, opened 2025-10-02
🐛(agent): Fix Supabase checkpoint task persistence and version handling

MH4GF:
## Issue

- resolve: 

## Why is this change needed?

This PR fixes critical issues in the Supabase checkpoint saver related to task persistence and version handling:

1. **Tasks channel version handling**: The checkpoint saver was not properly managing versions for the `tasks` channel, causing issues when retrieving checkpoint history
2. **Task persistence**: Tasks were not being correctly persisted in checkpoint blobs, leading to incomplete state snapshots

These fixes ensure that:
- Task information is correctly saved and retrieved in checkpoints
- Version handling for the tasks channel is consistent with other channels
- Checkpoint history accurately reflects pending tasks and next nodes

The changes include comprehensive integration tests that validate the correct behavior of checkpoint next nodes and pending tasks across multiple execution steps.

### DIFF
--- a/frontend/internal-packages/agent/src/checkpoint/SupabaseCheckpointSaver.ts
+++ b/frontend/internal-packages/agent/src/checkpoint/SupabaseCheckpointSaver.ts
@@ -815,10 +815,6 @@ export class SupabaseCheckpointSaver extends BaseCheckpointSaver<number> {
       versionEntries.set(TASKS, resolvedVersion)
     }
 
-    if (versionEntries.size === 0) {
-      return []
-    }
-
     const blobs: Database['public']['Tables']['checkpoint_blobs']['Insert'][] =
       []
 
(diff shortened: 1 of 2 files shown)


frontend/internal-packages/agent/src/checkpoint/SupabaseCheckpointSaver.ts
```diff
@@ -777,18 +798,30 @@ export class SupabaseCheckpointSaver extends BaseCheckpointSaver<number> {
...
+    }
+
     const blobs: Database['public']['Tables']['checkpoint_blobs']['Insert'][] =
       []
 
-    for (const [channel, version] of Object.entries(versions)) {
+    for (const [channel, version] of versionEntries) {
       const value = values[channel]
 
-      if (value === undefined || value === null) {
```
MH4GF, line 791: One major cause was that null was excluded here. In the `branch:to:*` channel used by LangGraph's edge, null can sometimes be the value. The problem was that it wasn't being saved.

PR #3702, opened 2025-10-07
♻️ refactor(erd-core): add Hash schema and integrate URL handling functions

tnyo43:
## Issue

~- resolve:~

## Why is this change needed?

We have implement some similar URL handling function in different components. This PR will integrate them into util directory. Also, it introduce `HashSchema` to give better types to the functions.

## Summary by CodeRabbit

- Refactor
  - Centralized URL/query handling and element ID generation for ERD internals; no visible UI changes.
- Tests
  - Added unit tests for query-string behavior when activating a table and for column-hash validation.
- Chores
  - Expanded public exports and added types/schemas to make ERD utilities reusable across the codebase.


frontend/packages/erd-core/src/schemas/hash/types.ts
```diff
@@ -0,0 +1,4 @@
+import type { InferOutput } from 'valibot'
+import type { hashSchema } from './schemas'
+
+export type Hash = InferOutput<typeof hashSchema>
```
tnyo43, line 4: This `Hash` type is same with `${string}__columns__${string}`. It will extend its definition when we allow other hash values by updating `hashSchema`.

junkisai (approved):
thx!

PR #3710, opened 2025-10-07
Replace hardcoded userName with actual user data in chat messages

sasamuku:
## Issue

No specific issue, improvement task

## Why is this change needed?

Currently, when users send messages in a design session, the `userName` in the optimistic HumanMessage is hardcoded as 'Dummy'. This change replaces the hardcoded value with the actual user's name fetched from Supabase.

## Changes

- **SessionDetailPage.tsx**: Fetch userName from Supabase using `auth.getUser()` and query the `users` table
- **SessionDetailPageClient.tsx**: Pass userName as a prop to useStream hook
- **useStream.ts**: Use the actual userName prop instead of hardcoded 'Dummy' in optimistic message creation
- **PublicSessionDetailPage.tsx**: Add userName prop with 'Guest' as the default value for public sessions

## Implementation Details

The userName is:
- Fetched once in the Server Component (SessionDetailPage) for efficiency
- Passed through props to maintain clear data flow: Server Component → Client Component → Hook
- Defaults to 'User' for authenticated users without a name, and 'Guest' for public sessions

This approach aligns with the existing pattern in `sessionCreationHelpers.ts` where new session creation also fetches userName from Supabase.

### DIFF
--- a/frontend/apps/app/components/SessionDetailPage/SessionDetailPageClient.tsx
+++ b/frontend/apps/app/components/SessionDetailPage/SessionDetailPageClient.tsx
@@ -30,6 +30,7 @@ type Props = {
   initialIsPublic: boolean
   initialWorkflowError?: string | null
   initialArtifact: Artifact | null
+  userName: string
 }
 
 // Determine the initial active tab based on available data
@@ -68,6 +69,7 @@ export const SessionDetailPageClient: FC<Props> = ({
   initialIsPublic,
   initialWorkflowError,
   initialArtifact,
+  userName,
 }) => {
   const [activeTab, setActiveTab] = useState<OutputTabValue | undefined>(
     determineInitialTab(initialArtifact, initialVersions),
@@ -117,6 +119,7 @@ export const SessionDetailPageClient: FC<Props> = ({
   const { isStreaming, messages, start, error } = useStream({
     initialMessages: chatMessages,
     designSessionId,
+    userName,
   })
 
   // Combine streaming error with workflow errors
--- a/frontend/apps/app/components/SessionDetailPage/hooks/useStream/useStream.ts
+++ b/frontend/apps/app/components/SessionDetailPage/hooks/useStream/useStream.ts
@@ -59,8 +59,13 @@ const extractStreamErrorMessage = (rawData: unknown): string => {
 type Props = {
   designSessionId: string
   initialMessages: BaseMessage[]
+  userName: string
 }
-export const useStream = ({ designSessionId, initialMessages }: Props) => {
+export const useStream = ({
+  designSessionId,
+  initialMessages,
+  userName,
+}: Props) => {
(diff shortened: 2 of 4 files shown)


sasamuku:
@claude 
Please revise the PR description so that it aligns with the title “Introduce optimistic UI updates for second time HumanMessage.”

frontend/apps/app/components/SessionDetailPage/hooks/useStream/useStream.ts
```diff
@@ -260,9 +266,20 @@ export const useStream = ({ designSessionId, initialMessages }: Props) => {
...
 
+      const tempId = `optimistic-${crypto.randomUUID()}`
+      const optimisticMessage = new HumanMessage({
+        content: params.userInput,
+        id: tempId,
+        additional_kwargs: {
+          userName: userName,
+        },
+      })
+      setMessages((prev) => [...prev, optimisticMessage])
```
junkisai: Since the current implementation duplicates the first sent message, it would be better to either add an argument like `isFirst` to the `start` function and branch inside that function, or add a function like `addOptimisticHumanMessage` inside `useStream` and adjust it so that this function runs only on `onSendMessage`.

https://github.com/user-attachments/assets/96c301db-7369-4a11-a404-c28ffd744ea4

frontend/apps/app/components/SessionDetailPage/SessionDetailPageClient.tsx
```diff
@@ -30,6 +30,7 @@ type Props = {
   initialIsPublic: boolean
   initialWorkflowError?: string | null
   initialArtifact: Artifact | null
+  userName: string
```
junkisai: [IMO]
To avoid confusion when looking only at the component props, how about renaming `userName` with a role-specific identifier? Based on its usage, `senderName` seems appropriate.

frontend/apps/app/components/SessionDetailPage/hooks/useStream/useStream.ts
sasamuku: Sorry, I overlooked that. I'll fix it.

frontend/apps/app/components/SessionDetailPage/SessionDetailPageClient.tsx
sasamuku: I think that's a good idea. I will revise it! 

@claude Please revise it.

MH4GF (approved):
Saito-san has commented on the concerns. Everything else seems fine. Thank you!

frontend/apps/app/components/SessionDetailPage/hooks/useStream/useStream.ts
sasamuku: @junkisai 
Fix this in 3c9d714f8d76e1388b4cfabf4b9d5fffa5fc4fc2

PR #3730, opened 2025-10-09
♻️ Remove buildingSchemaId from workflow state management

MH4GF:
## Issue

- resolve: https://github.com/route06/liam-internal/issues/5815

## Why is this change needed?

Remove redundant `buildingSchemaId` from workflow state management. Instead of passing `buildingSchemaId` through the entire workflow state, we now retrieve it from `repositories.schema.getSchema(designSessionId).id` when needed. This ensures we always use the most up-to-date `buildingSchemaId` and reduces state redundancy.

## Changes

- Remove `buildingSchemaId` from `workflowAnnotation`, `dbAgentAnnotation`, `qaAgentAnnotation`
- Remove from `AgentWorkflowParams` type and all workflow setup functions
- Update `schemaDesignTool` to use `schemaResult.value.id` instead of passed `buildingSchemaId`
- Remove from tool configuration validation (`getToolConfigurable`)
- Remove from all test files
- Update schema-bench executor to use `designSessionId`
- Remove unnecessary `buildingSchema` database query from chat stream API

## Technical Rationale

The `buildingSchemaId` can always be obtained from the repository using `designSessionId`:
```typescript
const schemaResult = await repositories.schema.getSchema(designSessionId)
// schemaResult.value.id is the buildingSchemaId
```

This approach:
- Ensures using the most up-to-date `buildingSchemaId`
- Reduces redundant state management
- Simplifies the workflow state structure
- Makes the data flow more explicit


frontend/internal-packages/agent/src/db-agent/tools/schemaDesignTool.ts
```diff
@@ -113,7 +112,7 @@ export const schemaDesignTool: StructuredTool = tool(
     const totalStatements = results.length
 
     const versionResult = await repositories.schema.createVersion({
-      buildingSchemaId,
+      buildingSchemaId: schemaResult.value.id,
```
MH4GF, line 115: Originally, it was only used here, and since the latest buildingSchemaId was obtained within the tool, it was unnecessary.

PR #3732, opened 2025-10-09
✨ feat(command palette): support Column options

tnyo43:
## Issue

- resolve:

## Why is this change needed?

**This change will not published yet**

I added `TableColumnOptions` to display column options of each table. Users can select table by pressing "Tab" key and see the Column options in the CommandPalette. Selecting column options will open `TableDetail` component, so they can access to more information easily.

preview: https://liam-oybrx95ix-liambx.vercel.app/erd/p/github.com/mastodon/mastodon/blob/1bc28709ccde4106ab7d654ad5888a14c6bb1724/db/schema.rb

demo

Users  can go to the "table" input mode with a specific table by pressing "Tab" key. In the input mode, users can see only the selected table option and its column options.

https://github.com/user-attachments/assets/9b3a503a-b514-4cbb-97dd-f14278380338

## Summary by CodeRabbit

- New Features
  - Command Palette can activate a table-focused mode with refined filtering.
  - Browse tables with their columns, including icons for key/nullable types.
  - Preview tables and columns directly from suggestions.
  - Keyboard navigation enhanced; open table/column targets or new tabs.
  - Deep links to specific columns within a table.

- Style
  - Indented list styling with a guide line for grouped table/column options.

- Tests
  - Expanded coverage for table/column navigation, previews, suggestion parsing, and URL generation.

### DIFF
--- a/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalette.tsx
+++ b/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPalette.tsx
@@ -42,7 +42,7 @@ export const CommandPalette: FC = () => {
             A search-based interface that allows quick access to various
             commands and features within the application.
           </DialogDescription>
-          <CommandPaletteContent isTableModeActivatable />
+          <CommandPaletteContent />
         </DialogContent>
       </DialogPortal>
     </Dialog>


frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteOptions/utils/getColumnTypeMap.ts
tnyo43, line 1: There is [a very similar logic here](https://github.com/liam-hq/liam/blob/a683c9375e19464e9d614be7d2b5c4a5bb50b70f/frontend/packages/erd-core/src/features/erd/components/ERDContent/components/TableNode/TableColumnList/TableColumn/TableColumn.tsx#L37-L90). However, the way of checking "foreign key" is different approach, so I think it's not possible to reuse it here.

Maybe we can refactor them and integrate their logic into a same one, but let's make it as another issue.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteOptions/TableColumnOptions.tsx
```diff
@@ -0,0 +1,180 @@
...
+          window.open(getTableLinkHref(suggestedTableName))
+        } else {
+          goToERD(suggestedTableName)
+        }
+      }
+    }
+
+    document.addEventListener('keydown', down)
+    return () => document.removeEventListener('keydown', down)
+  }, [suggestion, goToERD])
```
tnyo43, line 88: This is the same logic with [the TableOption component](https://github.com/liam-hq/liam/blob/b22a67806988e7957c9898b5fe558160b6c81d36/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteOptions/TableOptions.tsx#L30-L51). I've just hard copied to reduce diffs of this PR. I'll work on refactoring after this PR is merged and remove duplicated code.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteContent/CommandPaletteContent.tsx
```diff
@@ -1,31 +1,68 @@
...
+  return 0
 }
 
-export const CommandPaletteContent: FC = () => {
+type Props = {
+  isTableModeActivatable?: boolean
+}
+
+export const CommandPaletteContent: FC<Props> = ({
+  isTableModeActivatable = false,
```
tnyo43, line 52: `isTableModeActivatable` will be `false` by default, so it won't be displayed for the users yet.

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteOptions/TableColumnOptions.tsx
```diff
@@ -0,0 +1,180 @@
...
+    case 'PRIMARY_KEY':
+      return <KeyRound {...props} />
+    case 'FOREIGN_KEY':
+      return <Link {...props} />
+    case 'NOT_NULL':
+      return <DiamondFillIcon aria-label={undefined} {...props} />
+    case 'NULLABLE':
+      return <DiamondIcon aria-label={undefined} {...props} />
+  }
+}
```
tnyo43, line 49: Icons of each column type is like this 😉 

Image

MH4GF: THIS IS GREAT!!!!!

PR #3736, opened 2025-10-10
✨ Add scroll-to-bottom button to improve chat UX

junkisai:
## Issue

- resolve: N/A (UX improvement)

## Why is this change needed?

In long chat conversations, users often need to scroll back to the latest messages after reviewing previous content. This change adds a scroll-to-bottom button that appears when users scroll up, providing a clear and convenient way to return to the most recent messages without manual scrolling.

## Changes

- Added ScrollToBottomButton component with smooth fade-in/out animations
- Implemented visibility threshold (48px) to prevent button flicker during small scroll movements
- Refactored chat layout structure for proper button positioning (relative wrapper + absolute button)
- Simplified useScrollToBottom hook by removing automatic scroll logic and exposing scrollToBottom function
- Added ArrowDown icon export to @liam-hq/ui package
- Added scroll event listeners to dynamically toggle button visibility based on scroll position

## Technical Details

The button visibility is calculated by measuring the distance between scroll position and bottom of the container. A 48px threshold (button height + spacing) ensures the button doesn't flicker when users are near the bottom. The component also recomputes visibility on message updates to handle streaming scenarios.

## Demo

| Streaming | Reload |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/faf967d0-6162-45ad-bded-fb2122339381" /> | <video src="https://github.com/user-attachments/assets/ec680f30-5cc1-40bb-9404-b0ac2f6d6cce" /> | 

## Summary by CodeRabbit

* **New Features**
  * Added a “Scroll to bottom” button in Chat to quickly jump to the latest messages.
* **Style**
  * Updated chat layout with a dedicated message list wrapper for smoother scrolling and better placement of status and error indicators.
  * Introduced polished styling and animations for the new scroll button.
* **Chores**
  * Added a new down-arrow icon to the shared icon set.
  * Improved default smooth auto-scroll behavior when new messages appear.

### DIFF
--- a/frontend/apps/app/components/SessionDetailPage/components/Chat/Chat.module.css
+++ b/frontend/apps/app/components/SessionDetailPage/components/Chat/Chat.module.css
@@ -7,14 +7,19 @@
   min-height: 0; /* Allow flex items to shrink */
 }
 
-.messagesContainer {
+.messageListWrapper {
+  position: relative;
   flex: 1;
+  min-height: 0;
+}
+
+.messageList {
   overflow-y: scroll;
   overflow-x: hidden; /* Prevent horizontal scrolling */
   padding: var(--spacing-4);
   display: flex;
   flex-direction: column;
   gap: var(--spacing-5);
-  min-height: 0; /* Allow the container to shrink */
+  height: 100%;
   scroll-behavior: smooth; /* Smooth scrolling */
 }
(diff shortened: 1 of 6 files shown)


frontend/apps/app/components/SessionDetailPage/components/Chat/useScrollToBottom.ts
```diff
@@ -1,44 +1,23 @@
-import { useCallback, useEffect, useRef, useState } from 'react'
```
junkisai, line 1: In the previous implementation of the custom hook, the view automatically scrolled to the bottom whenever the `messages` value changed. 

This caused unwanted scrolling while users were reading streaming messages, degrading the UX. I have adjusted it so that auto-scroll to the bottom occurs only on the initial load.

PR #3773, opened 2025-10-14
♻️ Extract GitHub API utilities and integrate token refresh

NoritakaIkeda:
## Issue

- resolve: https://github.com/route06/liam-internal/issues/5764

Functional Test
https://github.com/route06/liam-internal/issues/5764#issuecomment-3410090541

I'd appreciate it if you could read my inline comments during the review.

**my inline comments**
- https://github.com/liam-hq/liam/pull/3773#discussion_r2435309607
- https://github.com/liam-hq/liam/pull/3773#discussion_r2435502430
- https://github.com/liam-hq/liam/pull/3773#discussion_r2435521640
- https://github.com/liam-hq/liam/pull/3773#discussion_r2435529842
- https://github.com/liam-hq/liam/pull/3773#discussion_r2435544574
- https://github.com/liam-hq/liam/pull/3773#discussion_r2435553743
- https://github.com/liam-hq/liam/pull/3773#discussion_r2435588186
- https://github.com/liam-hq/liam/pull/3773#discussion_r2435767566
- https://github.com/liam-hq/liam/pull/3773/files#r2438134767

## Why is this change needed?

This change extracts GitHub API calls into dedicated utility modules and integrates automatic token refresh functionality. This improves code organization and ensures that expired GitHub OAuth tokens are automatically refreshed before making API calls, preventing authentication failures.

## Summary by CodeRabbit

* **New Features**
  * Encrypted HttpOnly storage for GitHub OAuth tokens.
  * Server-side GitHub token refresh endpoint and client-side trigger to invoke it.
  * Token-driven GitHub installations flow with expiry detection and refresh signaling.

* **Refactor**
  * Logout now performed via a server API that clears tokens and session state.
  * Project creation UI gates and disables installation actions when a refresh is required.

* **Chores**
  * Added internal security utilities and env guidance for the encryption keyring.

### DIFF
--- a/frontend/internal-packages/security/package.json
+++ b/frontend/internal-packages/security/package.json
@@ -18,11 +18,11 @@
     "neverthrow": "8.2.0"
   },
   "devDependencies": {
-    "@biomejs/biome": "2.2.4",
+    "@biomejs/biome": "2.2.5",
     "@liam-hq/configs": "workspace:*",
-    "@types/node": "22.18.6",
-    "eslint": "9.36.0",
-    "typescript": "5.9.2"
+    "@types/node": "22.18.8",
+    "eslint": "9.37.0",
+    "typescript": "5.9.3"
   },
   "scripts": {
     "fmt": "concurrently \"pnpm:fmt:*\"",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -536,7 +536,7 @@ importers:
         version: 8.2.0
       valibot:
         specifier: 1.1.0
-        version: 1.1.0(typescript@5.9.2)
+        version: 1.1.0(typescript@5.9.3)
     devDependencies:
       '@biomejs/biome':
         specifier: 2.2.5
@@ -700,20 +700,20 @@ importers:
         version: 8.2.0
     devDependencies:
       '@biomejs/biome':
-        specifier: 2.2.4
-        version: 2.2.4
+        specifier: 2.2.5
+        version: 2.2.5
       '@liam-hq/configs':
         specifier: workspace:*
         version: link:../configs
       '@types/node':
-        specifier: 22.18.6
-        version: 22.18.6
+        specifier: 22.18.8
+        version: 22.18.8
       eslint:
-        specifier: 9.36.0
-        version: 9.36.0(jiti@2.6.1)
+        specifier: 9.37.0
+        version: 9.37.0(jiti@2.6.1)
       typescript:
-        specifier: 5.9.2
-        version: 5.9.2
+        specifier: 5.9.3
+        version: 5.9.3
 
   frontend/internal-packages/storybook:
     dependencies:


frontend/apps/app/app/auth/callback/[provider]/route.ts
```diff
@@ -1,19 +1,59 @@
...
+        () => undefined,
+        (e) => {
+          const err = e instanceof Error ? e : new Error(String(e))
+          Sentry.captureException(err, {
+            tags: {
+              area: 'auth_cookie_write',
+            },
+          })
+          return undefined
+        },
```
Comment (NoritakaIkeda) [inline frontend/apps/app/app/auth/callback/[provider]/route.ts:56]: 📝 Access tokens and refresh tokens must be obtained during login. Since this is a critical path, we want to monitor it with Sentry without stopping the app due to errors.

NoritakaIkeda:
I closed and reopened Supabase to get it running.

frontend/internal-packages/security/src/cryptoBox.ts
```diff
@@ -0,0 +1,91 @@
...
+  keyId: string,
+  ciphertext: Buffer,
+  initializationVector: Buffer,
+  authenticationTag: Buffer,
+): Result<string, Error> {
+  const entry = RING.find((k) => k.id === keyId)
+  if (!entry) return err(new Error('Unknown key id'))
+  return fromThrowable(() => {
+    const decipher = crypto.createDecipheriv(
+      'aes-256-gcm',
```
NoritakaIkeda, line 84: 📝
I chose AES-256-GCM for the following reasons:

* It can be easily implemented in TypeScript.
* AES-256 is a strong symmetric encryption algorithm recommended by NIST and Japan’s IPA.
* GCM mode is standardized in NIST SP 800-38D as an authenticated encryption mode, offering both security and efficiency.
* It has wide adoption and proven reliability as a government and industry standard.

NoritakaIkeda:
@codex review

frontend/internal-packages/github/src/api.browser.ts
NoritakaIkeda, line 1: 📝 nits: 
Since .browser.ts appeared to be client-specific, I revised it to a more precise naming convention.

frontend/apps/app/app/projects/new/page.tsx
```diff
@@ -31,12 +33,40 @@ export default async function NewProjectPage() {
...
+    shouldTriggerRefresh = true
+  }
 
   return (
-    <ProjectNewPage
-      installations={installations}
-      organizationId={organizationId}
-    />
+    <>
+      <TokenRefreshKick trigger={shouldTriggerRefresh} />
```
NoritakaIkeda: Cookie writes are disallowed inside RSCs, so automatic refresh logic cannot live in libs/github/token.ts.
Instead, the refresh is handled via a Route Handler and explicitly triggered from the client.
After cookies are updated, a client-side re-render ensures consistency between the RSC state and the browser state.

With server actions, the state synchronization after cookie updates can be unclear, which may cause inconsistencies between the UI and the server state.
Therefore, we currently handle this explicitly via a Route Handler + client trigger.
That said, if there are strong reasons or use cases for moving this logic to a server action, we’d love to hear them.

.env.template
```diff
@@ -24,3 +24,8 @@ SENTRY_ORG=""
 SENTRY_PROJECT=""
 SUPABASE_SERVICE_ROLE_KEY=""
 NEXT_PUBLIC_ALLOWED_DOMAINS=""
+
+# KEYRING
+# Usage: Secret for app-side encryption/signing.
+# Generate (Node.js): node -e "console.log(require('crypto').randomBytes(32).toString('base64'))"
+KEYRING=""
```
NoritakaIkeda: 📝
The key will be rotated to ensure a more secure implementation.
In the current implementation, developers need to perform the rotation manually,
but by using a KMS (Key Management Service), it can be automated.
The procedure document will be added in a separate PR.

frontend/apps/app/app/api/github/token/refresh/route.ts
```diff
@@ -0,0 +1,162 @@
...
+      { ok: false, error: 'refresh_failed' },
+      { status: 401 },
+    )
+  }
+
+  // 4) Persist into cookies (access + refresh)
+  await writeAccessToken(refreshed.access_token, refreshed.newExpiresAt)
+  const refreshExpiresAt = new Date(Date.now() + THIRTY_DAYS_MS).toISOString()
+  await writeRefreshToken(refreshed.refresh_token, refreshExpiresAt)
+  // refreshed successfully; cookies updated
```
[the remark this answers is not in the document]
NoritakaIkeda: The file already sets httpOnly: true, secure: isProd(), and sameSite: 'lax'.
The suggestion about “missing sameSite” is a false positive.

Setting SameSite: 'strict' would prevent cookies from being sent during top-level navigations from external sites (e.g., GitHub OAuth redirects or deep links), which could cause login detection failures or state inconsistencies on the first request.

frontend/apps/app/libs/github/cookie.ts
NoritakaIkeda, line 1: ✨
## Adoption Policy: Why Access Tokens Are Stored in HttpOnly Cookies

We chose to store GitHub access and refresh tokens in HttpOnly Cookies rather than a database, for the following reasons:

###  Reasons

- **XSS-Resistant**: Cookies marked as `HttpOnly` cannot be accessed via JavaScript, reducing the risk of token theft via XSS.
- **Reduced Leak Impact**: Since tokens are not stored server-side, a database leak won't expose all user tokens at once.
- **Fast and Scalable**: No need to query the database on every request—tokens are sent automatically by the browser, reducing server load.
- **Industry Best Practice**: Many modern web applications use HttpOnly cookies for session/token management.
- Also, the token is used **only during active user sessions**, not for background or batch processing. So server-side persistence is unnecessary.

### Concerns and Mitigations

- **CSRF Protection Required**: Cookies are sent automatically with requests, so CSRF prevention is necessary.  
  → We apply `SameSite=Lax`, which mitigates most CSRF risks.

### Supabase Constraint

- While `getSession()` allows reading the provider token, there is no built-in way to set a new access/refresh token after it is rotated.
- Therefore, we store and rotate tokens manually via HttpOnly cookies.

frontend/apps/app/app/api/github/token/refresh/route.ts
```diff
@@ -0,0 +1,143 @@
...
+
+const GITHUB_TOKEN_URL = 'https://github.com/login/oauth/access_token'
+const THIRTY_DAYS_MS = 30 * 24 * 60 * 60 * 1000
+
+const GitHubTokenSuccessSchema = v.object({
+  access_token: v.string(),
+  refresh_token: v.string(),
+  token_type: v.string(),
+  expires_in: v.number(), // seconds
+})
```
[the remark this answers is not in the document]
NoritakaIkeda: GitHub OAuth (expiring user tokens) typically rotates the refresh_token upon refresh, invalidating the previous one. If your implementation assumes this behavior, making the refresh_token optional risks silently breaking downstream processing by swallowing inconsistencies. Treating a missing refresh_token as an error is the safer approach.

[✨ Use GitHub's refresh_token_expires_in for cookie expiry](https://github.com/liam-hq/liam/pull/3773/commits/697fa4f46643e7d20bb45d64461e538cd6f3e2a5)

frontend/apps/app/app/api/github/token/refresh/route.ts
```diff
@@ -0,0 +1,143 @@
...
+      { ok: false, error: 'refresh_failed' },
+      { status: 401 },
+    )
+  }
+
+  // 4) Persist into cookies (access + refresh)
+  await writeAccessToken(refreshed.access_token, refreshed.newExpiresAt)
+  const refreshExpiresAt = new Date(Date.now() + THIRTY_DAYS_MS).toISOString()
+  await writeRefreshToken(refreshed.refresh_token, refreshExpiresAt)
+  // refreshed successfully; cookies updated
```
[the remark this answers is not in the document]
NoritakaIkeda: "Fallback to existing refresh_token" is not recommended. "Match TTL to refresh_token_expires_in" is acceptable.
- Reason: As mentioned above, assuming GitHub rotates tokens, falling back to an old refresh_token will immediately invalidate it.
Instead, "no new refresh_token" should be detected, logged, and treated as an error.
On the other hand, setting the cookie expiration to match the server-provided TTL is correct.

[✨ Use GitHub's refresh_token_expires_in for cookie expiry](https://github.com/liam-hq/liam/pull/3773/commits/697fa4f46643e7d20bb45d64461e538cd6f3e2a5)

frontend/apps/app/libs/github/token.ts
```diff
@@ -0,0 +1,25 @@
...
+import { readAccessToken } from './cookie'
+
+const SAFETY_MS = 3 * 60 * 1000 // 3 minutes safety window
+// No schema/refresh handling here; handled in Route Handler
+
+function isExpiringSoon(expiresAtIso: string | null | undefined): boolean {
+  if (!expiresAtIso) return true
+  const expiresAt = new Date(expiresAtIso).getTime()
+  return Date.now() + SAFETY_MS >= expiresAt
+}
```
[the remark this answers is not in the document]
NoritakaIkeda: [🐛 Add NaN check to isExpiringSoon function](https://github.com/liam-hq/liam/pull/3773/commits/8ebcb27fa57d88b114cb65438cf21ae9ef809516)
I fixed it!

frontend/internal-packages/security/src/cryptoBox.ts
NoritakaIkeda, line 1: * **Input:** Uses the key (ID and 32-byte key from `KEYRING`), a **nonce** (12-byte random number, different each time), and the plaintext to encrypt.
* **Method:** Encrypts using **AES-256-GCM**, producing a **ciphertext** and an **authentication tag** (16-byte marker to verify the data hasn’t been tampered with).
* **Storage:** Combines `{ key_id, ciphertext, initialization_vector, authentication_tag }`, encodes them in Base64, and stores the result in a secure **HttpOnly cookie** with an expiration time.

frontend/apps/app/middleware.ts
```diff
@@ -135,11 +135,12 @@ export const config = {
...
     /*
      * Match all request paths except for the ones starting with:
+     * - api (api routes - we'll run authentication manually in each route)
      * - _next/static (static files)
      * - _next/image (image optimization files)
      * - favicon.ico (favicon file)
      * Feel free to modify this pattern to include more paths.
      */
-    '/((?!_next/static|_next/image|favicon.ico|.*\\.(?:svg|png|jpg|jpeg|gif|webp)$).*)',
+    '/((?!api|_next/static|_next/image|favicon.ico|.*\\.(?:svg|png|jpg|jpeg|gif|webp)$).*)',
```
NoritakaIkeda: 📝
To delete the token stored in cookies upon logout, I added frontend/apps/app/app/api/auth/logout/route.ts.
However, since frontend/apps/app/middleware.ts was causing a redirect to /api/auth/logout after login, I applied the following fix.

.env.template
```diff
@@ -24,3 +24,15 @@ SENTRY_ORG=""
...
+# Usage: Secret for app-side encryption/signing.
+# Format: comma-separated entries "<kid>:<base64key>" — first entry is the current key.
+# - <kid>: recommended format is a readable version like "kYYYY-MM" (e.g., k2025-01)
+# - <base64key>: 32-byte random key, base64-encoded
+# Generate 32-byte key (Node.js):
+#   node -e "console.log(require('crypto').randomBytes(32).toString('base64'))"
+# Example (current key first, with a rotated fallback):
+#   KEYRING="k2025-01:BASE64_KEY_FOR_2025_01,k2024-12:BASE64_KEY_FOR_2024_12"
+# Leave empty in local dev if not needed.
+KEYRING=""
```
MH4GF: Isn't this name too generic? Maybe we should narrow down the namespace.

frontend/apps/app/app/api/auth/logout/route.ts
```diff
@@ -0,0 +1,11 @@
...
+import { NextResponse } from 'next/server'
+import { clearTokens } from '../../../../libs/github/cookie'
+
+export async function POST() {
+  await fromAsyncThrowable(clearTokens)().match(
+    () => undefined,
+    () => undefined,
+  )
+  return NextResponse.json({ ok: true })
+}
```
MH4GF: Confirmation: I was wondering if it's acceptable to return an OK response regardless of success or failure. As a compromise, I think we should return an error while having the frontend ignore it. Returning an error should trigger automatic notifications via Sentry.

frontend/apps/app/app/api/github/token/refresh/route.ts
```diff
@@ -0,0 +1,152 @@
...
+      })
+      const res = await fetch(GITHUB_TOKEN_URL, {
+        method: 'POST',
+        headers: {
+          Accept: 'application/json',
+          'Content-Type': 'application/x-www-form-urlencoded',
+        },
+        body,
+        cache: 'no-store',
+      })
```
MH4GF, line 78: Did oktokit not work for you?

frontend/apps/app/app/api/github/token/refresh/route.ts
```diff
@@ -0,0 +1,152 @@
...
+    return NextResponse.json(
+      { ok: false, error: 'refresh_failed' },
+      { status: 401 },
+    )
+  }
+
+  // 4) Persist into cookies (access + refresh)
+  await writeAccessToken(refreshed.access_token, refreshed.newExpiresAt)
+  const rtei = refreshed.refresh_token_expires_in ?? 0
+  const refreshTtlMs = rtei > 0 ? rtei * 1000 : THIRTY_DAYS_MS
```
MH4GF: Is 30 days really okay?
Since the situation where it doesn't return is unintended behavior, it might be better to treat it as an error.

frontend/apps/app/components/CommonLayout/AppBar/UserDropdown.tsx
```diff
@@ -63,6 +64,19 @@ export const UserDropdown: FC<Props> = ({ avatarUrl, userName, userEmail }) => {
...
+        ok: boolean
+        headers: { get(name: string): string | null }
+        json: () => Promise<unknown>
+      }
+      await fromPromise<HttpResponse>(
+        fetch('/api/auth/logout', { method: 'POST', cache: 'no-store' }),
+      ).match(
+        () => {},
+        () => {},
+      )
```
MH4GF: The difference between Supabase's signout and this logout API is unclear.
Right now, the intent is more about clearing cookies than logging out.
Since the organizationId cookie is also cleared above, if that's the case, it seems acceptable to move the following three operations to /api/logout:

- Delete the organizationId cookie
- Delete the token cookie
- Supabase signout

This could also be another PR.

frontend/apps/app/components/ProjectNewPage/components/InstallationSelector/InstallationSelector.tsx
```diff
@@ -42,11 +43,9 @@ export const InstallationSelector: FC<Props> = ({
...
   const handleSelectInstallation = (installation: Installation) => {
     setSelectedInstallation(installation)
-    startTransition(() => {
-      const formData = new FormData()
-      formData.append('installationId', installation.id.toString())
-      repositoriesAction(formData)
-    })
+    const formData = new FormData()
+    formData.append('installationId', installation.id.toString())
+    repositoriesAction(formData)
```
MH4GF: It seems better to return startTransition.

frontend/packages/security/README.md
MH4GF, line 1: Please navigate to frontend/internal-packages.

frontend/packages/security/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "type": "git",
+    "url": "https://github.com/liam-hq/liam.git"
+  },
+  "license": "Apache-2.0",
+  "private": false,
+  "version": "0.1.0",
+  "type": "module",
+  "publishConfig": {
+    "access": "public"
+  },
```
MH4GF: Please set it to private.

frontend/packages/security/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "devDependencies": {
+    "@biomejs/biome": "2.2.4",
+    "@liam-hq/configs": "workspace:*",
+    "@types/node": "22.18.6",
+    "eslint": "9.36.0",
+    "typescript": "5.9.2",
+    "vitest": "3.2.4"
+  },
+  "scripts": {
+    "build": "tsc",
```
MH4GF: Please do not build.

MH4GF (changes requested):

.env.template
NoritakaIkeda: I fixed it!
[♻️(security): Migrate to LIAM_GITHUB_OAUTH_KEYRING environment variable](https://github.com/liam-hq/liam/pull/3773/commits/8f4ded6c2d9fa6e40edb4951cb895bab7ea81f4c)

frontend/apps/app/app/api/auth/logout/route.ts
NoritakaIkeda: I fixed it!
[✨(auth): Add comprehensive logout API endpoint](https://github.com/liam-hq/liam/pull/3773/commits/aece9bd07b5214c0aefe70ca6df6b27281053ee7)

frontend/apps/app/components/CommonLayout/AppBar/UserDropdown.tsx
NoritakaIkeda: I fixed it!
[♻️(auth): Move logout endpoint from /api/auth/logout to /api/logout](https://github.com/liam-hq/liam/pull/3773/commits/d80f93e969c370a181d146966e902a7a83d70477)
[✨(auth): Add comprehensive logout API endpoint](https://github.com/liam-hq/liam/pull/3773/commits/aece9bd07b5214c0aefe70ca6df6b27281053ee7)
[✨(organizations): Add clearOrganizationIdCookie server action](https://github.com/liam-hq/liam/pull/3773/commits/61e31467c39b585e0996c7411655ac273eefb6f3)
[(ui): Refactor UserDropdown to use new logout API](https://github.com/liam-hq/liam/pull/3773/commits/164db736f99bb2de5a391d4b5175cf83695526cf)

frontend/apps/app/app/api/github/token/refresh/route.ts
NoritakaIkeda: I fixed it! No fallback was needed.
[✨(auth): Make refresh_token_expires_in required in token refresh](https://github.com/liam-hq/liam/pull/3773/commits/0aeeaa027b9b44a4da5c94a10fe447c54b295390)

frontend/apps/app/app/api/github/token/refresh/route.ts
NoritakaIkeda: According to [the Octokit documentation](https://github.com/octokit/auth-oauth-user.js/?tab=readme-ov-file#authoptions-or-octokitauthoptions)
, the "refresh" option is available only for GitHub Apps, and only when expiring user tokens are enabled.
Since I couldn’t find a corresponding option in Octokit, I implemented it using fetch instead.

frontend/apps/app/components/ProjectNewPage/components/InstallationSelector/InstallationSelector.tsx
NoritakaIkeda: [🐛(ui): Fix installation selector state update timing](https://github.com/liam-hq/liam/pull/3773/commits/69c832fb046efd0ba5e1cb7dc44eb0cdf01bf61e)
I fix it!

frontend/packages/security/README.md
NoritakaIkeda: [♻️(security): Move security package to internal-packages](https://github.com/liam-hq/liam/pull/3773/commits/183f7d542892d5161d08761e6acd656cc97dd775)
I fixed it!

frontend/packages/security/package.json
NoritakaIkeda: [♻️(security): Move security package to internal-packages](https://github.com/liam-hq/liam/pull/3773/commits/183f7d542892d5161d08761e6acd656cc97dd775)
I fixed it!

frontend/packages/security/package.json
NoritakaIkeda: [♻️(security): Move security package to internal-packages](https://github.com/liam-hq/liam/pull/3773/commits/183f7d542892d5161d08761e6acd656cc97dd775)
I fixed it!

NoritakaIkeda:
@MH4GF I’ve also updated how neverthrow is used, so please take a look!
[♻️(security): Refactor cryptoBox to use Result monad patterns](https://github.com/liam-hq/liam/pull/3773/commits/eb6843866f6c1afa7023d305e1df2fa5166c44db)
[♻️(auth): Refactor token refresh to use neverthrow Result patterns](https://github.com/liam-hq/liam/pull/3773/commits/8541e7f20eb1003410f976b90bf7b433ad517071)

frontend/apps/app/app/api/github/token/refresh/route.ts
MH4GF: OK, I understood!

frontend/apps/app/app/projects/new/page.tsx
```diff
@@ -25,18 +27,36 @@ export default async function NewProjectPage() {
...
+
+  const { installations, needsRefresh } = await tokenResult
+    .asyncAndThen((token) => {
+      if (!token) {
+        return fromPromise(
+          Promise.resolve({
+            installations: [],
+            needsRefresh: true,
+          }),
+        )
```
MH4GF: @claude Since it can be written this way, please revise it.

```suggestion
        return okAsync({
            installations: [],
            needsRefresh: true,
          })
```

frontend/apps/app/middleware.ts
```diff
@@ -135,6 +138,7 @@ export const config = {
   matcher: [
     /*
      * Match all request paths except for the ones starting with:
+     * - api (api routes - we'll run authentication manually in each route)
```
MH4GF: @claude Please delete this comment

frontend/apps/app/tsconfig.json
```diff
@@ -9,6 +9,14 @@
...
     "exactOptionalPropertyTypes": false,
     "noUncheckedIndexedAccess": true,
+    "baseUrl": ".",
+    "paths": {
+      "@liam-hq/security/*": ["../../internal-packages/security/src/*"],
+      "@liam-hq/neverthrow": [
+        "../../internal-packages/neverthrow/src/index.ts"
+      ],
+      "@liam-hq/neverthrow/*": ["../../internal-packages/neverthrow/src/*"]
+    },
```
MH4GF: This should be unnecessary. Implement it using package dependency resolution without using paths.

frontend/apps/app/features/organizations/services/clearOrganizationIdCookie.ts
```diff
@@ -0,0 +1,14 @@
...
+import { ORGANIZATION_ID_KEY } from '../constants'
+
+export async function clearOrganizationIdCookie(): Promise<void> {
+  const store = await cookies()
+  store.set(ORGANIZATION_ID_KEY, '', {
+    path: '/',
+    maxAge: 0,
+    expires: new Date(0),
+  })
+}
```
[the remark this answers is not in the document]
NoritakaIkeda: I fixed it!

frontend/apps/app/app/projects/new/page.tsx
NoritakaIkeda: [🔧(app): Remove unused path aliases from tsconfig.json](https://github.com/liam-hq/liam/pull/3773/commits/4af40bd70a936ac4de1b6c1c8ca54b37fa79be37) 
[📝(middleware): Remove outdated API routes comment](https://github.com/liam-hq/liam/pull/3773/commits/d3e80e7acdad90a8b5cadaa8da3b6c48b2e74c4a) 
[♻️(projects): Simplify okAsync usage in new project page](https://github.com/liam-hq/liam/pull/3773/commits/972b2c9f3341542a0fb14c3a8788dbec4034d7a3)

Since Claude wasn't working properly, I handled it.

frontend/apps/app/middleware.ts
NoritakaIkeda: [🔧(app): Remove unused path aliases from tsconfig.json](https://github.com/liam-hq/liam/pull/3773/commits/4af40bd70a936ac4de1b6c1c8ca54b37fa79be37) 
[📝(middleware): Remove outdated API routes comment](https://github.com/liam-hq/liam/pull/3773/commits/d3e80e7acdad90a8b5cadaa8da3b6c48b2e74c4a) 
[♻️(projects): Simplify okAsync usage in new project page](https://github.com/liam-hq/liam/pull/3773/commits/972b2c9f3341542a0fb14c3a8788dbec4034d7a3)

Since Claude wasn't working properly, I handled it.

frontend/apps/app/tsconfig.json
NoritakaIkeda: [🔧(app): Remove unused path aliases from tsconfig.json](https://github.com/liam-hq/liam/pull/3773/commits/4af40bd70a936ac4de1b6c1c8ca54b37fa79be37) 
[📝(middleware): Remove outdated API routes comment](https://github.com/liam-hq/liam/pull/3773/commits/d3e80e7acdad90a8b5cadaa8da3b6c48b2e74c4a) 
[♻️(projects): Simplify okAsync usage in new project page](https://github.com/liam-hq/liam/pull/3773/commits/972b2c9f3341542a0fb14c3a8788dbec4034d7a3)

Since Claude wasn't working properly, I handled it.

frontend/apps/app/app/auth/callback/[provider]/route.ts
```diff
@@ -11,9 +17,44 @@ export async function GET(request: Request) {
...
       await ensureUserHasOrganization()
+      // Persist provider tokens into encrypted HttpOnly cookies (best-effort)
+      const { session } = data
+      const access = session?.provider_token ?? ''
+      const refresh = session?.provider_refresh_token ?? ''
+      // received provider tokens for cookie write
+      // Access token expiration: 8 hours.
+      // GitHub's OAuth access tokens typically expire after 8 hours. See: https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps#step-3-github-redirects-back-to-your-site
+      const EIGHT_HOURS_MS = 8 * 60 * 60 * 1000
+      const THIRTY_DAYS_MS = 30 * 24 * 60 * 60 * 1000
```
Comment (sasamuku) [inline frontend/apps/app/app/auth/callback/[provider]/route.ts:31]: GitHub refresh tokens are valid for 6 months (180 days), not 30 days. This hardcoded value causes refresh tokens to be considered expired after 30 days, even though they're still valid. Why?
https://docs.github.com/ja/apps/creating-github-apps/authenticating-with-a-github-app/refreshing-user-access-tokens

frontend/apps/app/components/TokenRefreshKick/TokenRefreshKick.tsx
```diff
@@ -0,0 +1,47 @@
...
+        if (ok) {
+          router.refresh()
+        }
+      }
+      await fromPromise<HttpResponse>(
+        fetch('/api/github/token/refresh', {
+          method: 'POST',
+          cache: 'no-store',
+        }),
+      ).match(onOk, () => {})
```
sasamuku, line 41: When the refresh token expires (after 6 months of inactivity), the refresh API returns 401, but the error is silently ignored. Users are left with a non-functional UI with no explanation.

Current user experience:
1. User returns after 6 months
2. UI becomes unresponsive (installations don't load)
3. No error message shown
4. User is confused

Please let me know if my understanding is wrong.

NoritakaIkeda: Thank you. We acknowledge that we do not currently present the behavior after the refresh token expires to users.
We will implement one of the following solutions, or a better alternative, while referencing other systems' UX:
- Prompt the user to sign out and sign in again
- Have the user re-authorize the GitHub OAuth connection

frontend/apps/app/tsconfig.json
MH4GF: Thanks!!

frontend/apps/app/app/auth/callback/[provider]/route.ts
Comment (NoritakaIkeda) [reply frontend/apps/app/app/auth/callback/[provider]/route.ts:31]: I missed this point earlier.
I intended to align with any standard HttpOnly cookie storage duration, but since web standards and common security guidelines don't specifically mention it, I'll match the lifetime of GitHub tokens instead. I'll address this in the next PR.

Comment (sasamuku) [reply frontend/apps/app/app/auth/callback/[provider]/route.ts:31]: >I intended to align with any standard HttpOnly cookie storage duration.

I don't think there's a problem with that itself, but using magic numbers is not good.

PR #3787, opened 2025-10-16
✨ feat(command palette): support Tab completion

tnyo43:
## Issue

~- resolve:~

## Why is this change needed?

This PR enables input completion when pressing the Tab key. In particular, when the completion is triggered for a table, it switches to Table input mode and allows users to select column options.

**preview** -> https://liam-erd-sample-j1zbxbquo-liambx.vercel.app/

**completion suffix**

Screenshot 0007-10-16 at 9 01 24

**when the suggestion value doesn't start from input value**

Screenshot 0007-10-16 at 9 01 36

**completion suffix with column option**

Screenshot 0007-10-16 at 9 08 36

## Summary by CodeRabbit
- New Features
    - Added inline suggestion/completion overlay in the command palette search input.
    - Enabled Tab to autocomplete suggestions, including command and table options when applicable.
    - Improved mode handling: smoother activation of table mode, backspace switches to default when input is empty, and no unintended mode switches on completion.
    - Refined placeholder behavior to reduce visual noise when suggestions are shown.
- Tests
    - Overhauled test coverage for input modes, Tab/Backspace behavior, and suggestion suffix rendering.

## Summary by CodeRabbit

- New Features
  - Inline suggestion overlay in the command palette search input shows a live completion suffix alongside your typing.
  - Press Tab to accept visible suggestions for commands and columns without leaving your current context.
- Bug Fixes
  - Backspace behavior is more predictable when the input is empty vs. not.
  - Suggestions no longer attempt to complete with tables where not applicable.
  - Placeholder visibility is improved to avoid clashing with inline suggestions.
- Style
  - Updated input layout to support overlaid suggestions with improved readability.

### DIFF
--- a/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteSearchInput/CommandPaletteSearchInput.test.tsx
+++ b/frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteSearchInput/CommandPaletteSearchInput.test.tsx
@@ -276,6 +276,7 @@ describe('displays a suggestion to complete the input', () => {
         mode={{ type: 'default' }}
         suggestion={{ type: 'table', name: 'user-settings' }}
         setMode={mockSetMode}
+        isTableModeActivatable
       />,
       { wrapper },
     )
@@ -294,6 +295,7 @@ describe('displays a suggestion to complete the input', () => {
         mode={{ type: 'default' }}
         suggestion={{ type: 'table', name: 'user-settings' }}
         setMode={mockSetMode}
+        isTableModeActivatable
       />,
       { wrapper },
     )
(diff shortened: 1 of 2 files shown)


frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteSearchInput/CommandPaletteSearchInput.module.css
```diff
@@ -6,13 +6,21 @@
...
+  position: relative;
+  width: 100%;
+}
+
 .input {
+  position: relative;
   padding: var(--spacing-2) 0;
   width: 100%;
   border-radius: var(--border-radius-smbase);
+  z-index: 1;
```
tnyo43, line 23: This change is to make the input element clickable when suffix displayed.
Without this change, the suffix (with `.completionSuffix` class) comes in front of the input because it has `position: absolute`.

Image

frontend/packages/erd-core/src/features/erd/components/ERDRenderer/CommandPalette/CommandPaletteSearchInput/CommandPaletteSearchInput.test.tsx
```diff
@@ -103,46 +99,97 @@ describe('in case input mode is "default"', () => {
...
+        expect(input).toHaveValue('Copy Link')
+      })
     })
   })
 })
 
-describe.each<CommandPaletteInputMode>([
-  { type: 'command' },
-  { type: 'table', tableName: 'users' },
-])('in case input mode is $type', (inputMode) => {
```
tnyo43, line 139: I break them into a different describe blocks since the behaviours with "Tab" pressing are different for these input modes.

PR #3806, opened 2025-10-18
♻️ refactor(erd-core): unify items into `DetailItem` component in `TableDetail`

tnyo43:
## Issue

~- resolve:~

## Why is this change needed?

This PR refactors `Indexes` component and `Columns` component in `TableDetail` since they share very similar implementation. I created `DetailItem` component and move the styles into it. The `Indexes` and `Columns` are basically just call `DetailItem` component for styling.

With `DetailItem`, we can create the items of `CollapsibleHeader`s as like:

```tsx
<CollapsibleHeader>
  <DetailItem>
    <DetailItemHeading href="...">name</DetailItemHeading>
  </DetailItem>
  {/* to show the details as a table */}
  <GridTableRoot>
    ...
  </GridTableRoot>
</CollapsibleHeader>
```

## Summary by CodeRabbit

* **Style**
  * Improved scroll alignment when navigating between items in nested collapsible sections
  * Enhanced visual styling for focused detail items with updated color scheme

### DIFF
--- a/frontend/packages/erd-core/src/features/erd/components/ERDContent/components/TableNode/TableDetail/CollapsibleHeader/CollapsibleHeader.tsx
+++ b/frontend/packages/erd-core/src/features/erd/components/ERDContent/components/TableNode/TableDetail/CollapsibleHeader/CollapsibleHeader.tsx
@@ -1,5 +1,7 @@
 import { ChevronDown, ChevronUp, IconButton } from '@liam-hq/ui'
+import clsx from 'clsx'
 import {
+  type ComponentProps,
   type FC,
   type KeyboardEvent,
   type MouseEvent,
@@ -73,3 +75,10 @@ export const CollapsibleHeader: FC<CollapsibleHeaderProps> = ({
     </>
   )
 }
+
+type CollapsibleHeaderItemProps = ComponentProps<'div'>
+
+export const CollapsibleHeaderItem: FC<CollapsibleHeaderItemProps> = ({
+  className,
+  ...props
+}) => <div className={clsx(styles.item, className)} {...props} />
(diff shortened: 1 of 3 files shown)


frontend/packages/erd-core/src/features/erd/components/ERDContent/components/TableNode/TableDetail/CollapsibleHeader/CollapsibleHeader.module.css
```diff
@@ -58,8 +58,16 @@
   margin-top: -41px;
 }
 
+.header + .content .item {
+  scroll-margin-top: 41px;
+}
```
tnyo43, line 63: To avoid calculate `scroll-margin-top` in each detail component, I declare this here.
If we would like to introduce third collapsible header item in the future, we need to add something like:

```css
/* 123px (= 41px * 3) for the second CollapsibleHeader component */
.header + .content + .header + .content {
  padding-top: 123px;
  margin-top: -123px;
}

.header + .content + .header + .content .item {
  scroll-margin-top: 123px;
}
```

frontend/packages/erd-core/src/features/erd/components/ERDContent/components/TableNode/TableDetail/Columns/ColumnsItem/ColumnsItem.tsx
```diff
@@ -59,39 +57,23 @@ export const ColumnsItem: FC<Props> = ({
...
-            />
-          )}
-          <NotNull tableId={tableId} column={column} />
-        </GridTableRoot>
-      </div>
-    </>
+    <DetailItem id={elementId} className={diffStyle} isFocused={isFocused}>
+      <DetailItemHeading href={`#${elementId}`}>
+        {column.name}
+      </DetailItemHeading>
```
tnyo43, line 63: We don't need to define `BlinkCircle` component directory from this component, and no need to set `h3` heading with link separately in the `Columns` and `Indexes` components 👍

frontend/packages/erd-core/src/features/erd/components/ERDContent/components/TableNode/TableDetail/CollapsibleHeader/CollapsibleHeader.module.css
tnyo43: We can calculate this with JS if it doesn't fit with the code bases.
Then the `DetailItem` will have a property something like `index` and set style like `style={{  marginTop: (index * 41), ... }}`

PR #3825, opened 2025-10-20
♻️ refact(erd-core): table detail item hash logic

tnyo43:
## Issue

- resolve:

## Why is this change needed?

This PR will unify the hash logics for indexe and column items in TableDetail.
The hash related logics are composed in the `userEditing` context, because I thought the logic should be adress near the `active` table state, and each TabledDetail item refers them.

## Summary by CodeRabbit

* **Refactor**
  * Reorganized focus state management for column and index elements to use a centralized provider system instead of local component state. Focus tracking from URL hash changes is now managed through a shared context, simplifying component architecture while maintaining existing focus functionality.

### DIFF
--- a/frontend/packages/erd-core/src/stores/userEditing/Provider.tsx
+++ b/frontend/packages/erd-core/src/stores/userEditing/Provider.tsx
@@ -63,8 +63,10 @@ export const UserEditingProvider: FC<Props> = ({
   )
 
   const [focusedElementId, setFocusedElementId] = useState(
-    // location.hash starts with '#'; decode to match actual DOM id
-    location.hash.slice(1),
+    typeof location === 'object'
+      ? // location.hash starts with '#'; decode to match actual DOM id
+        location.hash.slice(1)
+      : '',
   )
 
   // update focusedElementId when hash changes


frontend/packages/erd-core/src/stores/userEditing/context.ts
```diff
@@ -7,6 +7,8 @@ export type UserEditingContextValue = {
   activeTableName: string | null
   setActiveTableName: (tableName: string | null) => void
 
+  focusedElementId: string
```
tnyo43, line 10: I made this to export the hash value, but not a hash updating function.
Currently, the hash will be updated by href of anchor elements, so it's no need to export such an update function.

frontend/packages/erd-core/src/features/erd/components/ERDContent/components/TableNode/TableDetail/Columns/Columns.tsx
```diff
@@ -9,22 +9,6 @@ type Props = {
...
-  // update focusedElementId when hash changes
-  useEffect(() => {
-    const updateState = () => {
-      const elementId = location.hash.slice(1)
-      setFocusedElementId(elementId)
-    }
-
-    window.addEventListener('hashchange', updateState)
-    return () => window.removeEventListener('hashchange', updateState)
-  }, [])
```
tnyo43, line 26: Since the state and update event listener are brought to the `userEditing` context, each element group components (`Columns` and `Indexes`) become very simpler 👍

PR #3860, opened 2025-10-23
✨ Show initial_schema_snapshot as Version 0 in VersionDropdown

junkisai:
## Issue

- resolve: https://github.com/route06/liam-internal/issues/5914

## Why is this change needed?

Currently, the VersionDropdown only displays versions from the `building_schema_versions` table. However, the initial schema stored in `building_schemas.initial_schema_snapshot` (the state before the agent starts generating schemas) is not included in the version history.

By making the initial schema selectable as Version 0, users can track the complete evolution of their schema from the very beginning.

### Core changes
1. **Moved `isEmptySchema` utility to `@liam-hq/schema` package** for reusability across the codebase
2. **Modified `getVersions` service** to include Version 0 when `initial_schema_snapshot` exists and is non-empty
3. **Removed client-side `initial_schema_snapshot` display logic** from UI components
4. **Simplified tab initialization** by removing the `initial_schema_snapshot` parameter
5. **Centralized version display logic** in the `getVersions` service

### Implementation details
- Version 0 is a virtual version (no actual DB record) with:
  - `id: 'initial'`
  - `number: 0`
  - `patch: null`, `reverse_patch: null`
- The service checks if `initial_schema_snapshot` exists and is non-empty using the `isEmptySchema` utility
- Version 0 is appended to the end of the versions array (after sorting by number descending)

## Demo

| After linking the schema file | When switching versions | When the schema file is not associated |
|--------|--------|--------|
| <video src="https://github.com/user-attachments/assets/ca5f15f8-0417-48c7-8374-12fefe2ad2bc" /> | <video src="https://github.com/user-attachments/assets/ed30b2a7-c470-4950-a1f0-119c0aff6ceb" /> | <video src="https://github.com/user-attachments/assets/95f8ffb1-652f-484f-a2ed-426af0398c78" /> |

### DIFF
--- a/frontend/apps/app/components/SessionDetailPage/utils/isEmptySchema.ts
+++ b/frontend/apps/app/components/SessionDetailPage/utils/isEmptySchema.ts
@@ -1,9 +1,0 @@
-import type { Schema } from '@liam-hq/schema'
-
-export function isEmptySchema(schema: Schema): boolean {
-  return (
-    Object.keys(schema.tables).length === 0 &&
-    Object.keys(schema.enums).length === 0 &&
-    Object.keys(schema.extensions).length === 0
-  )
-}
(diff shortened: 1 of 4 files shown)


frontend/apps/app/components/SessionDetailPage/utils/isEmptySchema.ts
NoritakaIkeda, line 1: I hadn't discovered that it existed in the internal package. Thank you for the refactoring.

NoritakaIkeda (approved):
LGTM!
Thanks!

PR #3862, opened 2025-10-23
♻️(app): Fix diff baseline to use initial schema 

junkisai:
## Issue

- resolve: https://github.com/route06/liam-internal/issues/5916

## Why is this change needed?

The current implementation compares selected version N with previous version N-1 in ERD and Migration Viewer diff views. This makes it difficult to see cumulative changes from the initial schema state.

This change fixes the diff comparison baseline to always use Version 0 (initial_schema_snapshot), allowing users to see the total changes from the initial state at any version.

## Demo

https://github.com/user-attachments/assets/3db26b6e-dead-4089-b1cf-b88926c6318c

### Removed complex reverse patch logic
- Deleted `buildPrevSchema` function and related utilities
- Removed recursive patch operations for baseline reconstruction

### Simplified baseline schema retrieval  
- Directly use `initial_schema_snapshot` from `building_schemas` table as baseline
- Pass baseline schema from server components to client components
- Remove client-side baseline schema calculation in `useRealtimeVersionsWithSchema`

### Benefits
- **Performance**: No patch operations needed - direct snapshot access
- **User Experience**: Shows cumulative changes from Version 0 at any version
- **Maintainability**: Reduced code complexity (8 files changed, 37 additions, 140 deletions)

## Impact

- ERD diff view now compares against Version 0 instead of N-1
- Migration Viewer diff view now compares against Version 0 instead of N-1
- Both public and authenticated session views are updated

---


MH4GF (approved):
LGTM 👍🏻

PR #3889, opened 2025-10-26
🧪 trial(command palette) with long name options

tnyo43:
## Issue

- resolve:

## Why is this change needed?

- prevent suggestion overflowing
- make option in multiple lines when too long
- make limit on the table name

|before|after|
| --- | --- |
|Screenshot 0007-10-26 at 12 06 48|Screenshot 0007-10-26 at 12 10 03


tnyo43:
This change will be implemented in https://github.com/liam-hq/liam/pull/3897

PR #3941, opened 2025-10-29
✨(qa-agent): Add automatic test failure analysis and retry workflow

sasamuku:
## Issue

- resolve: https://github.com/route06/liam-internal/issues/5964

## Why is this change needed?

This PR implements automatic test failure analysis and retry functionality for the QA agent. When SQL tests fail during execution, the system now:

1. Analyzes which tests failed and identifies the failure type (SQL vs schema issues)
2. Automatically retries failed SQL tests up to 3 times
3. Provides failure context to the LLM for improved SQL regeneration
4. Tracks retry attempts to prevent infinite loops

This improves the QA agent's reliability and success rate by learning from previous failures and automatically correcting SQL generation mistakes.

### Architecture

The retry workflow is implemented as a loop in the QA agent graph:

```
invokeRunTestTool → analyzeTestFailures → routeAfterAnalyzeFailures
                                              ↓ (if retries available)
                    ← continueToRequirements ← prepareSqlRetry
```

### Key Components

1. **Infrastructure** (`constants.ts`, `qaAgentAnnotation.ts`)
   - `QA_AGENT_RECURSION_LIMIT`: 50 steps to support 3 retry cycles
   - `failureAnalysis` state field with `failedSqlTestIds` and `failedSchemaTestIds`

2. **Analysis Node** (`analyzeTestFailuresNode.ts`)
   - Extracts failed test IDs from test results
   - Populates `failureAnalysis` state for routing decisions

3. **Routing Logic** (`routeAfterAnalyzeFailures.ts`)
   - Checks retry eligibility (max 3 attempts per test)
   - Routes to `prepareSqlRetry` or `END` based on retry count

4. **Retry Preparation** (`prepareSqlRetryNode.ts`)
   - Clears SQL field for failed tests
   - Enables regeneration with failure feedback

5. **Failure Feedback** (`formatPreviousFailures.ts`, prompt updates)
   - Formats error messages and timestamps for LLM context
   - Includes previous failure details in SQL generation prompts

### Retry Limit Configuration

- **Per-test limit**: 3 retry attempts
- **Total recursion limit**: 50 steps (initial run + 3 retry cycles)
- Tests exceeding retry limit will not be retried further

## Test Plan

- [x] Unit tests updated to include `failureAnalysis` state initialization
- [x] Integration test updated with recursion limit configuration
- [x] All linting checks passing
- [ ] Manual testing of retry workflow with intentionally failing SQL
- [ ] Verification that retry limit is enforced correctly
- [ ] Confirmation that failure context improves SQL generation

## Summary by CodeRabbit

* **New Features**
  * QA agent adds automated failure analysis and a retry cycle for failed SQL tests, resetting and regenerating SQL up to configured attempts and including previous failure details in test-generation prompts.

* **Documentation**
  * QA agent workflow docs updated to describe the new failure-analysis, retry, and regeneration flow.

### DIFF
--- a/frontend/internal-packages/agent/src/constants.ts
+++ b/frontend/internal-packages/agent/src/constants.ts
@@ -29,10 +29,6 @@ export const DEFAULT_RECURSION_LIMIT = 10
  * - Retry cycles: 3 × 10 = 30 steps
  * - Total: 40 steps + buffer = 50 steps
  *
- * TODO: Remove this constant when QA agent message summarization is implemented.
- * The message summarization will reduce the overall message size, eliminating
- * the need for this elevated limit.
- *
  * @see {@link routeAfterAnalyzeFailures} for retry limit logic (MAX_RETRY_ATTEMPTS = 3)
  */
 export const QA_AGENT_RECURSION_LIMIT = 50


sasamuku:
The non-determinism of LLMs causes the execution results of the agent to vary in size, sometimes small and sometimes large. 

This time, it was successful, so I will open a pull request. 
If the problem of excessively large comment sizes recurs, we will address it by lowering the test iteration limit.

NoritakaIkeda:
@sasamuku
Thanks for the PR!

> If the problem of excessively large comment sizes recurs, we will address it by lowering the test iteration limit.

I have some concerns about this point.
As I understand it, until a more fundamental solution is implemented (such as consolidating QA messages), there’s a possibility that the CI for `agent-deep-modeling` could fail on *any* future PR — is that correct?

**Possible implications:**

1. If we continue to require `agent-deep-modeling` to pass on every PR, it may take several reruns before it succeeds, potentially increasing lead time.
2. If we decide to ignore CI failures temporarily (assuming they are caused by this issue), merges won’t be blocked, but it will make it harder for CI to catch regressions reliably.

I plan to proceed with reviewing the code itself, but I’d like to discuss the merge timing since I have some concerns about it.

sasamuku:
@NoritakaIkeda 
Your concern is truly right.

Since the fundamental fix will take time, I'd like to proceed with this PR.
Therefore, we will try to temporarily adjust to reduce log output.

sasamuku:
@NoritakaIkeda 
disable test generation retries in the CI environment: 60ed5bf83

The goal in the CI environment isn't to increase the success rate of SQL tests, but rather to confirm that the workflow passes, so I think this is appropriate.

frontend/internal-packages/agent/src/qa-agent/nodes/analyzeTestFailuresNode.ts
```diff
@@ -0,0 +1,45 @@
...
+ * @param state - QA agent state containing test results
+ * @returns Updated state with failureAnalysis, or empty object if no failures
+ */
+export const analyzeTestFailuresNode = (state: QaAgentState) => {
+  const { analyzedRequirements } = state
+  const failedTestIds = extractFailedTests(analyzedRequirements)
+
+  return {
+    ...(failedTestIds.length > 0 && {
+      // TODO: Implement schema vs SQL issue classification
```
NoritakaIkeda, line 38: (Q)
In the future, is the idea to classify the failures and, based on that classification, decide whether to regenerate the test case, skip (or remove) it entirely, or fix the schema instead?

frontend/internal-packages/agent/src/qa-agent/nodes/prepareSqlRetryNode.ts
```diff
@@ -0,0 +1,36 @@
+import type { QaAgentState } from '../shared/qaAgentAnnotation'
+
+/**
+ * Prepares SQL tests for retry by clearing their SQL field
+ * This allows the testcaseGeneration subgraph to regenerate SQL with failure feedback
+ * @param state - QA agent state containing failure analysis
+ * @returns Updated state with cleared SQL for failed tests
+ */
+export const prepareSqlRetryNode = (state: QaAgentState) => {
+  const { failureAnalysis, analyzedRequirements } = state
```
NoritakaIkeda: IMO
The name prepareSqlRetryNode doesn’t clearly convey what the node actually does or why it’s used. Something like resetFailedSqlTestsNode would make its purpose and behavior more explicit.

frontend/internal-packages/agent/src/qa-agent/routing/routeAfterAnalyzeFailures.ts
```diff
@@ -0,0 +1,25 @@
...
+export const routeAfterAnalyzeFailures = (
+  state: QaAgentState,
+): 'prepareSqlRetry' | typeof END => {
+  const { failureAnalysis, analyzedRequirements } = state
+
+  if (!failureAnalysis || failureAnalysis.failedSqlTestIds.length === 0) {
+    return END
+  }
+
+  const MAX_RETRY_ATTEMPTS = 3
```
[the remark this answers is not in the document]
NoritakaIkeda: > This implements "3 attempts total" (initial + 2 retries), which matches the PR's "Per-test: 3 attempts" statement. However, the PR summary also says "retries failed SQL tests up to three times," implying 3 retries (4 attempts total).

I think this is just a wording nuance — having a total of three attempts including the initial run is fine.
cc: @sasamuku

frontend/internal-packages/agent/src/qa-agent/shared/qaAgentAnnotation.ts
```diff
@@ -28,6 +28,11 @@ export const qaAgentAnnotation = Annotation.Root({
   designSessionId: Annotation<string>,
   schemaIssues: qaSchemaIssuesAnnotation,
   generatedSqls: generatedSqlsAnnotation,
+  failureAnalysis: Annotation<
+    { failedSqlTestIds: string[]; failedSchemaTestIds: string[] } | undefined
+  >({
```
NoritakaIkeda, line 33: (Q)
Just to confirm, my understanding is that failedSchemaTestIds is a field intended for future use — is that correct?

frontend/internal-packages/agent/src/qa-agent/utils/formatPreviousFailures.ts
```diff
@@ -0,0 +1,34 @@
...
+  const failuresText = failedResults
+    .map((result, index) => {
+      return `
+## Previous Attempt ${index + 1}
+Executed At: ${result.executedAt}
+Error Message:
+\`\`\`
+${result.message}
+\`\`\`
+`
```
NoritakaIkeda, line 24: 📝 I'd like to continuously observe how well the model improves its SQL generation when we pass the error reasons back to it.

frontend/internal-packages/agent/src/constants.ts
```diff
@@ -20,3 +20,26 @@
...
+/**
+ * Maximum number of retry attempts for failed SQL test cases.
+ *
+ * Behavior:
+ * - CI environment: 0 attempts (fast execution, compact logs)
+ * - Development/Production: 3 attempts (allows learning from failures)
+ *
+ * @see {@link routeAfterAnalyzeFailures} for retry routing logic
+ */
+export const MAX_RETRY_ATTEMPTS = process.env['CI'] === 'true' ? 0 : 3
```
NoritakaIkeda, line 45: (nits)

Thanks! I think we should eventually remove this when we run the QA agent message summarization task.
I'll leave a TODO comment for that.
@claude

frontend/internal-packages/agent/src/qa-agent/nodes/analyzeTestFailuresNode.ts
sasamuku: Maybe no skips.
Failed tests are classified as either SQL issues or schema issues.

frontend/internal-packages/agent/src/qa-agent/nodes/prepareSqlRetryNode.ts
sasamuku: Exactly 👍

@claude fix this

frontend/internal-packages/agent/src/qa-agent/shared/qaAgentAnnotation.ts
sasamuku: Yes👍

frontend/internal-packages/agent/src/constants.ts
sasamuku: Hmm, I don't think that's necessary.

If we run the complete workflow in CI, the execution time will be too long and it will become a blocking issue when opening pull requests. 

I think we need to organize the purpose of why we run the agent on CI.

sasamuku:
@NoritakaIkeda 
>I think this is just a wording nuance — having a total of three attempts including the initial run is fine.

The number of total attempts is four including the initial run.
The retry count is 3, as declared in the constant.

frontend/internal-packages/agent/src/constants.ts
NoritakaIkeda: At this point, my position is that it’s better to remove it.
However, considering practical constraints such as timeouts, my stance is that in the medium to long term, it should ideally be removed.

The reasons are as follows:

1. Risk when retries are omitted during development
   　Even if everything works fine in CI during development, there’s a risk that in the user’s environment, errors could occur during retries, or a bug might cause infinite retries to run after delivery.

2. Risk of losing developers’ sense of execution time
   　Developers may lose a realistic sense of the total workflow completion time (including retries), which could reduce their motivation to shorten the overall processing time.


That is my position, but since this topic is somewhat outside the scope of this PR’s responsibility, I’d like to discuss it separately in another issue.🙏
I’ll revert the commit for now!

NoritakaIkeda: @claude 
[docs: Add TODO comment for QA_AGENT_RECURSION_LIMIT removal](https://github.com/liam-hq/liam/pull/3941/commits/365c987fcfac342f20e6c95fd4d833e2021dd6b3)

Revert this commit.

NoritakaIkeda:
@sasamuku
> The number of total attempts is four including the initial run.
The retry count is 3, as declared in the constant.

I realized I had misunderstood this part.
It looks like the issue has already been resolved with the change in [this commit](https://github.com/liam-hq/liam/commit/60ed5bf83644c7a283479e1d7582e09fad82fb4d)
, so I think we’re all good on this point 🙏

Thanks a lot for taking the time to respond to all my comments!
Once the CI passes, I think we can go ahead and merge this. 🚀

PR #3971, opened 2025-11-02
💄 style(ui): fix scrollbar style declarations to apply `-webkit-scrollbar` pseud classes to Safari

tnyo43:
## Issue

- resolve:

## Why is this change needed?

With The current application, the scrollbar style declarations are not applied in Safari.
This PR will fix scrollbar style declarations to apply `-webkit-scrollbar` pseude classes to Safari.

I also checked the style is applied properly in the latest Chrome and Firefox as like the before.

||before|after|
|---|---|---|
|TableDetail|Screenshot 0007-11-02 at 12 11 54|Screenshot 0007-11-02 at 12 12 06|
|CommandPalette|Screenshot 0007-11-02 at 12 11 03|Screenshot 0007-11-02 at 12 11 38|

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed scrollbar styling to display correctly across all browsers, including Safari

### DIFF
--- a/.changeset/sweet-chicken-stay.md
+++ b/.changeset/sweet-chicken-stay.md
@@ -1,0 +1,5 @@
+---
+"@liam-hq/ui": patch
+---
+
+- 💄 fix scrollbar styles so that ::-webkit-scrollbar pseudo classes take effect in Safari


frontend/packages/ui/src/styles/globals.css
```diff
@@ -7,27 +7,34 @@
...
-    var(--scrollbar-track-background);
-  scrollbar-width: var(--scrollbar-width);
-}
 
-/* WebKit scrollbar styling */
-*::-webkit-scrollbar {
-  width: var(--scrollbar-webkit-width);
-  height: var(--scrollbar-webkit-height);
+/* For browsers that support `scrollbar-*` properties */
+@supports (scrollbar-color: auto) {
```
tnyo43, line 12: I recommend to check the diff with [by hiding white spaces](https://github.com/liam-hq/liam/pull/3971/files?w=1)

I basically follow [this instruction](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar#adding_a_fallback_for_scrollbar_styles).
Since [the latest Safari (v26) supports `scrollbar-width` property](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/scrollbar-width#browser_compatibility), it prioritize the standard scrollbar styles when `scrollbar-width` has value other than `auto` and doesn't refer to `-webkit-scrollbar-*` pseuds classes. However, it [doesn't support the `scrollbar-color` property yet](https://developer.mozilla.org/ja/docs/Web/CSS/Reference/Properties/scrollbar-color#%E3%83%96%E3%83%A9%E3%82%A6%E3%82%B6%E3%83%BC%E3%81%AE%E4%BA%92%E6%8F%9B%E6%80%A7), so scrollbars used to be displayed with the default color in Safari.

This PR will make the `scrollbar-width` property declaration be not applied in the Safari, so `webkit-scrollbar-*` pseud classes will be applied properly.

tnyo43:
I actually found the other cases to be checked if scrollbar styles work properly [from the change of this PR](https://github.com/liam-hq/liam/pull/2913).
However, I'm not familiar with these features, so I appreciate if you would help me to check them 🙇

MH4GF (approved):
That was very educational. Thanks!

